PR #3, opened 2020-06-20
Support custom footer, add Next/Previous links

pacocoursey:
Builds on #2

Adds Previous and Next links to the top part of the footer. Can be disabled in config:

```json
prevLinks: false,
nextLinks: false
```

Custom footer is also changed in config (example in code):

```js
footer: <span>Hello!</span>
```

### DIFF
--- a/.nextra/config.js
+++ b/.nextra/config.js
@@ -1,0 +1,10 @@
+import userConfig from '../nextra.config'
+
+const defaultConfig = {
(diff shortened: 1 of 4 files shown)


shuding:
I added some padding to the 2 links, and made them gray by default because I feel the blue is a bit too eye-catching.

Really love this feature 🏆 !

PR #4, opened 2020-06-20
Add basic version of Search

pacocoursey:
Builds on #3 

Very basic implementation of a search. Right now, only searches through the titles of each page.

Todo:
- [ ] Search the hash links of each page (or full text somehow?)

### DIFF
--- a/.nextra/config.js
+++ b/.nextra/config.js
@@ -3,6 +3,7 @@ import userConfig from '../nextra.config'
 const defaultConfig = {
   nextLinks: true,
   prevLinks: true,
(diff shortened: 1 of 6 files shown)


shuding:
For more advanced use we can integrate Algolia DocSearch: https://docsearch.algolia.com.

But that needs to be added case by case, it’s great to have the basic search functionality (and styles!) so that we can easily switch to DocSearch.

shuding:
Wow I really enjoy the styles! Might be better than DocSearch, thanks!

PR #37, opened 2020-12-08
fix(loader): JSON parse throwing if isn't valid

davesnx:
For some reason when I cloned nextra-docs meta.json files were considered "META" from my editor and I kept doing silly mistakes.

-- Not sure why the deploy failed, but I would assume it's because It's not on my vercel account?

### DIFF
--- a/packages/nextra/src/loader.js
+++ b/packages/nextra/src/loader.js
@@ -21,6 +21,17 @@ function getFileName (resourcePath) {
   return removeExtension(path.basename(resourcePath))
 }
 
+const parseJsonFile = (path) => {
+  let parsed
+  try {
+    parsed = JSON.parse(file)
+  } catch (err) {
+    throw Error(`Error parsing ${path}, make sure it's a valid JSON \n` + err)
+  }
+
+  return parsed
+}
+
 async function getPageMap(currentResourcePath) {
   const extension = /\.(mdx?|jsx?)$/
   const mdxExtension = /\.mdx?$/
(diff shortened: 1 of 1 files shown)


packages/nextra/src/loader.js
```diff
@@ -21,6 +21,17 @@ function getFileName (resourcePath) {
...
 
+const parseJsonFile = (path) => {
+  let parsed
+  try {
+    parsed = JSON.parse(file)
+  } catch (err) {
+    throw Error(`Error parsing ${path}, make sure it's a valid JSON \n` + err)
+  }
+
+  return parsed
```
shuding, line 32: Thank you! Will this error thrown make the dev server recoverable? Like do I need to restart the dev server once I fix the JSON file?

davesnx: Couldn't make the examples working with the local version so I can't test it manually. But I would assume so, since if there's any throw in a webpack loader it's a node process and it would crash. I believe there's a way to "accumulate" those errors in the webpack plugin API somewhere. 

It might be simpler to `console.error` thought.

Let me try to run examples locally.

packages/nextra/src/loader.js
```diff
@@ -21,6 +21,17 @@ function getFileName (resourcePath) {
   return removeExtension(path.basename(resourcePath))
 }
 
+const parseJsonFile = (content, path) => {
+  let parsed
```
shuding, line 25:
```suggestion
  let parsed = {}
```

So even if it fails, the site can still be built.

shuding:
Yeah the deployment fails is because the `core` branch isn't something deployable, I'll fix it.

packages/nextra/src/loader.js
davesnx: Wanted to do that, good catch.

shuding:
Thanks!

PR #50, opened 2020-12-12
Correct github edit path

gr-qft:
Problem: Editing a page on Github currently leads to the wrong path (because `pages` is not specified in `path`). Moreover, there's a double `//` in the generated path because `normalizedPath` ends with `/` and `filepathWithName` begins with `/`; although this is still resolved correctly I think filepathWithName doesn't have to begin with `/`.

### DIFF
--- a/examples/docs/theme.config.js
+++ b/examples/docs/theme.config.js
@@ -1,7 +1,7 @@
 export default {
   repository: 'https://github.com/shuding/nextra',
   branch: 'core',
-  path: '/examples/docs',
+  path: '/examples/docs/pages',
   titleSuffix: ' – Nextra',
   logo: (
     <>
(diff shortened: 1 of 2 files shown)


shuding (commented):
Thanks for reporting this bug! I think a better way to fix is adding `pages` here:

```diff
-  return `${repository}/tree/${branch}${normalizedPath}${filepathWithName}`
+  return `${repository}/tree/${branch}${normalizedPath}pages${filepathWithName}`
```

https://github.com/shuding/nextra/blob/4cabd6d0736fa4f486c4f2f6438d0bf9bffc5a49/packages%2Fnextra-theme-docs%2Fsrc%2Ffooter.js#L44

shuding:
I just fixed it in this branch, thanks again for you efforts!

gr-qft:
👍

PR #55, opened 2020-12-12
Fix/css display property

gr-qft:
A small css correction. 

I have a question also: what was the reason of switching the display property instead of writing `open && <div> ... </div>`?


shuding:
There's no specific reason, I think toggling a classname rather than adding DOM nodes will be better in performance.

gr-qft:
> There's no specific reason, I think toggling a classname rather than adding DOM nodes will be better in performance.

👍

PR #62, opened 2020-12-17
Ignore files with empty basename, like .DS_Store etc

fanweixiao:
If there is a file with empty basename in the `pages` directory, will break the building process. Ignore them to fix.

---
This change is [Reviewable](https://reviewable.io/reviews/shuding/nextra/62)


shuding (approved):
Thank you!

PR #72, opened 2020-12-21
Add type definition for TypeScript

rot1024:
Thank you for developing nextra! I love it because i18n is fully supported.

I added type definition for TypeScript.

We can overwrite `Locale`, `Config`, `Meta` and `FrontMatter` type in theme developer side or user side as needed:

`next-env.d.ts`

```ts
declare module "nextra" {
  export type FrontMatter = { title?: string };
}
```


rot1024:
I will change branch so close this PR

PR #79, opened 2021-01-01
fix: og protocol peropty in meta

huozhi:
follow og protocol, use `property` and `content` to describe pairs
reference: https://ogp.me


pacocoursey (approved):

shuding:
Wow thanks!!

PR #88, opened 2021-01-27
Change the logo wrapper from inline to block

thien-do:
This is my proposal to fix the issue that a logo with SVG image won't be centered vertically properly as described in https://github.com/vercel/swr-site/issues/71


shuding:
Thank you for fixing this! No worries, the failed deployment is expected (I need to fix it).

Issue #82, opened 2021-01-05
Error:  pragma and pragmaFrag cannot be set when runtime is automatic.


LinkSake:
I followed the "Get Started" guide but when changing the initial `index.js` to a `index.md` file the following error appears:

```
SyntaxError: .../pages/index.md: pragma and pragmaFrag cannot be set when runtime is automatic.
```

I solved this by changing the react and react-dom version from 17.0.1 to ^16.13.1 on the package.json.
It may be a good idea to specify in the docs that Nextra dosen't support the new react version yet.

mxschmitt:
Duplicate of that and see there too for a workaround: https://github.com/shuding/nextra/issues/27#issuecomment-753445806

Issue #27, opened 2020-11-30
Compiling error on latest 0.3.1 & React 17


benoror:
```
SyntaxError: /Users/benoror/code/baseql-docs/pages/index.md: pragma and pragmaFrag cannot be set when runtime is automatic
```

<img width="1003" alt="Screen Shot 2020-11-29 at 18 58 15" src="https://user-images.githubusercontent.com/119117/100558638-db31c800-3274-11eb-8ab9-6b99abd78947.png">

Dependencies versions:

```json
  {
    "next": "^10.0.3",
    "nextra": "^0.3.1",
    "nextra-theme-docs": "^0.3.1",
    "react": "^17.0.1",
    "react-dom": "^17.0.1"
  }
```

sarthaktexas:
that's weird, works for me

benoror:
This is the commit where I was facing the issue, maybe you can reproduce it: https://github.com/BaseQL/baseql-docs/tree/910b1c86b37e5247251dda2a837e6d0bf88a82c7

sarthaktexas:
Ohhh I see. Looks like your `main` branch has the same version dependencies as mine but the commit you sent doesn't.

benoror:
This is the filing combo afaict: https://github.com/BaseQL/baseql-docs/blob/910b1c86b37e5247251dda2a837e6d0bf88a82c7/package.json

shuding:
@benoror Looks like something with React 17, I'll be taking a look. Thanks for reporting!

Related: https://github.com/vercel/next.js/issues/18461

plongoria:
Any update on this? I also have the same issue

shuding:
@plongoria FYI currently the only work around I have is to use `react@16`.

smakosh:
This might help @shuding https://github.com/system-ui/theme-ui/issues/1160

shuding:
You can add this `.babelrc` configuration to your project root:

```
{
  "presets": [
    [
      "next/babel",
      {
        "preset-react": {
          "runtime": "classic"
        }
      }
    ]
  ]
}
```

Tested and it would solve the problem for now. 👍

PR #96, opened 2021-02-21
Fix error with React 17, i18n routes and SSG/SSR

shuding:
This should fix the error with React 17:

```
pragma and pragmaFrag cannot be set when runtime is automatic
```

Related to #27 #82.


shuding:
I believe i18n is broken...

shuding:
We will remove this loader later.

PR #98, opened 2021-03-03
Fix search focusing

madmanwithabike:
Currently pressing '/' pulls focus from inputs, etc. Looks like this isn't intended, just an wandering bracket.

Cheers!


shuding (approved):
Good catch, thanks!

Issue #77, opened 2020-12-27
<code> should be inline instead of inline-block


shuding:
It doesn’t wrap if the content is too long.

PR #116, opened 2021-03-29
polish: contain code line inside code box

huozhi:
before vs after

~~not sure if it's the same thing wanted in #77~~

### DIFF
--- a/packages/nextra-theme-docs/src/styles.css
+++ b/packages/nextra-theme-docs/src/styles.css
@@ -298,6 +298,9 @@ pre .prism-code {
 .dark pre .prism-code {
   --c-highlight: #3d4b61;
 }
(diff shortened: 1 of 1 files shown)


shuding:
Oh #77 should have been fixed, what I meant was `<code>` should be inlined so it can wrap in this case:

image

leodr:
I think having horizontal scrolling is actually nicer than having line-wrapping. Especially on mobile devices, almost every line would be wrapped and code would become pretty unreadable since the indentation is broken.

Docusaurus also doesn't wrap lines of code (see [this page](https://docusaurus.io/docs/cli) on mobile), I think that's the way to do it.

shuding:
Closing this as ^

PR #120, opened 2021-04-01
build stork index only when stork:true

navin-moorthy:
Even when the stork is not enabled.
Stork runs on the build.


shuding (approved):
Thanks!

PR #121, opened 2021-04-01
add the output once and not while cached

navin-moorthy:
Image 1 - First Vercel build - Runs the `addStorkIndex` & creates the `.toml` file - No issues in this case.

Image 2 - Second Vercel build - Seems the nextra loader is cached & doesn't run the `addStorkIndex` & hence no `.toml` file.

image
image

I think now that the PR changes is not required. Because to fix the above issue, we removed the `.toml` file from gitignore. So when the `.toml` file is not created, `[output]` just gets appended to the same `.toml` file.


shuding (changes requested):
This is an experimental feature that I'm testing, thanks for finding this :)
But could you please explain more about the changes? Why stop building the index if there's no `[output]` in the toml file...?

navin-moorthy:
Updated the PR description. Please check !

B2o5T:
Closing since code was changed a lot

Issue #109, opened 2021-03-22
[question] collapsed by default


chilikasha:
hi,
is it possible to make folders in sidebar collapsed by default?

0-void:
Please let me know if you find any solution to this. I'm also looking for the same.

benoror:
Opened PR here: https://github.com/shuding/nextra/pull/122

PR #122, opened 2021-04-04
Allow menus collapsed by default

benoror:
Via `config.defaultMenuCollapsed`

Fix https://github.com/shuding/nextra/issues/109
Fix https://github.com/vercel/swr-site/issues/81

### DIFF
--- a/packages/nextra-theme-docs/src/misc/default.config.js
+++ b/packages/nextra-theme-docs/src/misc/default.config.js
@@ -10,6 +10,7 @@ export default {
   prevLinks: true,
   search: true,
   darkMode: true,
+  collapsed: false,
   font: true,
   footer: true,
   footerText: 'MIT 2020 © Nextra.',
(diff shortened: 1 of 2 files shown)


packages/nextra-theme-docs/src/misc/default.config.js
```diff
@@ -10,6 +10,7 @@ export default {
   prevLinks: true,
   search: true,
   darkMode: true,
+  collapsed: false,
```
shuding, line 13: Thanks for contributing! I personally prefer a more explicit name, such as `defaultMenuCollapsed: false`, what do you think?

benoror: absolutely, good call!

shuding (approved):
Thank you!

benoror:
Any chance to release this change, or waiting to get more delta?

PR #146, opened 2021-04-27
Fix `test.en.mdx` file typo errors, add search result list circular key handle support

AmagiDDmxh:


packages/nextra-theme-docs/src/search.js
```diff
@@ -38,27 +41,32 @@ const Search = ({ directories }) => {
     return matchSorter(directories, search, { keys: ['title'] })
   }, [search])
 
+  const moveActiveItem = up => {
+    const position = active + (up ? -1 : 1)
+    const { length } = results
+
+    // Modulo instead of remainder,
+    // see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder
+    const next = ((position % length) + length) % length
```
shuding, line 50: Just a nitpick, I think this will be enough since the delta is only 1:

```suggestion
    const next = (position + length) % length
```

AmagiDDmxh: Wow, didn't know that

shuding:
Thanks!

PR #163, opened 2021-05-19
documentation error fix

wotori:
Hello. 
I have just started to use nextra, it`s really inspiring, thanks!
Reading the documentation meet a simple error that break my brains for a while.
It's written to use wrong import:

`import 'nextra-theme-blog/style.css'`  
instead of:
`import 'nextra-theme-docs/style.css'`  

That`s follow with error import when starting project from the scratch instead of copy the repository.

Hope this save some time to somebody in the future.
Thanks.

### DIFF
--- a/pages/blog/get-started.mdx
+++ b/pages/blog/get-started.mdx
@@ -29,7 +29,7 @@ export default {
 4. Create `pages/_app.js` and include the theme stylesheet:
 
 ```jsx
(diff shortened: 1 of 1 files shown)


wotori:
It seems that I made a mistake on my side, i used old fork of code ...
Just followed the instruction again and got succeed.

PR #167, opened 2021-05-28
Fixed fresh loading from URLs containing header tags

nguyenvukhang:
Bug is seen when fresh-loading a page from a URL that contains header tags:
```
http://localhost:3000/foo#bar
```
(as opposed to without header tags: `http://localhost:3000/foo`)

1. Next/Prev Links at the footer will become just one Next Link, which will point to the first file in meta.json
2. Sidebar tree will not expand for the current page

This PR fixes that


shuding:
Thanks!

Issue #115, opened 2021-03-29
Dark Mode for blog theme


LucaNerlich:
Would it be possible, to add the dark mode button to the blog theme?

_Loving the project so far, huge thanks :)_

shuding:
This would be something nice to work on!

PR #169, opened 2021-06-10
Draft - Feature: Add blog darkmode

jacobestell:
Hey!

I'm trying to add dark mode, I've got a few commits in the right direction but got stuck on the style.css.

Most of this stuff is outside my normal experience so feedback is appreciated!


jacobestell:
I guess this would try to close #115

PR #180, opened 2021-06-29
Update loader.js

olmesm:
[As per Nextjs documentation](https://nextjs.org/docs/api-routes/introduction)

> Any file inside the folder pages/api is mapped to /api/* and will be treated as an API endpoint instead of a page. They are server-side only bundles and won't increase your client-side bundle size.

The `/pages/api/` directory shouldn't produce a route/sidemenu item. If API is needed as a title, `meta.json` can produce the mapping.

Really loving this project - thanks @shuding <3

### DIFF
--- a/packages/nextra/src/loader.js
+++ b/packages/nextra/src/loader.js
@@ -32,7 +32,7 @@ async function getPageMap(currentResourcePath) {
           )
 
           if (f.isDirectory()) {
-            if (f.name === "api") return null
+            if (fileRoute === "/api") return null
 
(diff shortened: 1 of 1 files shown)


shuding (commented):
Makes total sense, thanks for opening this PR! Is there any way to check if that `api` directory is under the top level? Pages like `/docs/api` should not be ignore.

olmesm:
Good thinking - have amended and tested locally.

Thanks @shuding!

shuding (approved):
Thank you!

PR #201, opened 2021-08-07
add titleSuffix config to blog theme

thien-do:
This PR brings the "titleSuffix" to the "blog" theme. It's similar to the one in the "docs" theme.

For example, with `titleSuffix: "- Shudin"`, the page https://shud.in/art will have the title "Arts - Shudin" instead of just "Arts". This only applies for the title inside "head" tag.

Implementation notes:
- There is no default value in order to be backward compatible (do you want me to change it to "- Nextra" like in the "docs" theme?)
- There is no "renderComponent" utility in the "blog" project so I just render the titleSuffix naively. Please tell me if this needs change.

Thank you

### DIFF
--- a/packages/nextra-theme-blog/src/index.js
+++ b/packages/nextra-theme-blog/src/index.js
@@ -28,7 +28,10 @@ const Layout = ({
   return (
     <React.Fragment>
       <Head>
-        <title>{title}</title>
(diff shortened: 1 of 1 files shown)


shuding (approved):
Thank you for adding this!

PR #209, opened 2021-08-20
Code example error

pablolizardo:
The last callout (explaining `error`) has type `warning` instead of `error` .


shuding (approved):
Thanks!

PR #212, opened 2021-08-30
Bugfix(docs): Change docs to blog theme

justinalexandershaw:
# Blog Theme
The blog install should use the blog theme, not the docs theme. Not sure if there are other issues that could be fixed related to this, but I found a small gotcha in the docs.


shuding:
Thanks!

Issue #214, opened 2021-08-31
Using nextra on Windows creates incorrect routes


dzearing:
In the loader.js, a `fileRoute` is calculated here:
 https://github.com/shuding/nextra/blob/33987cfb5444e510efca0750a031cb13dc9c90ec/packages/nextra/src/loader.js#L34

On Windows, this route has backslashes in it, while the rest of the code expects forward slashes (e.g. active link is calculated by a match of the relative url.) This causes bugs like `active` state to be miscalculated.

The fix, I think, is to add a `replace` at the end of the computation:

```js
      const fileRoute = path.join(route, removeExtension(f.name).replace(/^index$/, '')).replace(/\\/g, '/');
```

Maybe there's a better way, but this definitely resolves the issue. Want a PR?

dzearing:
I see this PR here fixing another related issue: https://github.com/shuding/nextra/pull/11/files

So I think this line becomes:

```
const fileRoute = slash(path.join(route, removeExtension(f.name).replace(/^index$/, '')));
```

PR #215, opened 2021-09-01
Fixing routes so they work in Windows

dzearing:
Addresses https://github.com/shuding/nextra/issues/214

build nextra site on Windows

Before: 
Click on nav links; no `active` class name added (because routes don't match)

After:
Active nav link has the `active` class


shuding:
Thank you!

PR #224, opened 2021-09-23
Add option to change the Git icon in the top right corner

maltejur:
This adds the new config option `gitIcon` to change the icon in the top right corner and replaces the `github` option with `gitLink` (for uniform naming, but the old option will still work).

Closes #175 

Example config:
```ts
import Gitlab from "@geist-ui/react-icons/gitlab";

const themeConfig = {
  gitLink: "https://gitlab.com/librewolf-community/browser",
  gitIcon: <Gitlab />,
// ...
 ```

Result:
Screenshot_20210923_182318

### DIFF
--- a/packages/nextra-theme-docs/src/misc/default.config.js
+++ b/packages/nextra-theme-docs/src/misc/default.config.js
@@ -1,7 +1,6 @@
 import React from 'react'
 
 export default {
-  gitLink: 'https://github.com/shuding/nextra',
   docsRepositoryBase: 'https://github.com/shuding/nextra',
   titleSuffix: ' – Nextra',
   nextLinks: true,
--- a/packages/nextra-theme-docs/src/misc/default.config.js
+++ b/packages/nextra-theme-docs/src/misc/default.config.js
@@ -1,7 +1,7 @@
 import React from 'react'
 
 export default {
-  github: 'https://github.com/shuding/nextra',
+  gitLink: 'https://github.com/shuding/nextra',
   docsRepositoryBase: 'https://github.com/shuding/nextra',
   titleSuffix: ' – Nextra',
   nextLinks: true,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/nextra-theme-docs/src/misc/default.config.js [hunk 1]
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 export default {
-  github: 'https://github.com/shuding/nextra',
+  gitLink: 'https://github.com/shuding/nextra',
```
shuding, line 4: How about calling it `projectLink` for more general use? For example it can be linked to any website not necessarily a git hosting platform.

shuding:
Thanks for your contribution!

packages/nextra-theme-docs/src/misc/default.config.js
maltejur: Yeah that sounds like a good idea. `projectLink` and `projectLinkIcon`?

packages/nextra-theme-docs/src/misc/default.config.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 export default {
-  github: 'https://github.com/shuding/nextra',
```
maltejur, line 4: Also, should I remove the default option here completely? This way, you can choose if you want a button to your repository or not.

packages/nextra-theme-docs/src/misc/default.config.js
shuding: Yep!

packages/nextra-theme-docs/src/misc/default.config.js
shuding: Sounds great!

shuding:
Thank you!

PR #230, opened 2021-10-17
Add `components` option to allow overriding MDX provider components

shuding:


B2o5T:
was made here https://github.com/shuding/nextra/pull/682

PR #236, opened 2021-11-05
Fix sidebar for `trailingSlash: true`

maltejur:
This will fix the sidebar for the case that `trailingSlash` is enabled in `next.config.js` (That option adds an `/` to the end of the routes and thus makes `next export` more useful).

Before:
Screenshot_20211105_202845

After:
Screenshot_20211105_202751

Of course everything still works as expected when `trailingSlash` is off.


etienne-martin:
Looks good to me

shuding:
Thanks a lot!

PR #246, opened 2021-11-12
Improve nav bar

jaredpalmer:
- `projectChatLink` and `projectChatLinkIcon` added to nav
- Improved a11y of Nav
	- changed translate link to button
	- fixed aria svg 
	- added screen reader only

CleanShot 2021-11-12 at 08 57 51@2x


shuding:
Thanks!

PR #281, opened 2021-12-08
Fix broken "nav" pages

maltejur:
This fixes some issues on pages with the type `nav` in the `meta.json`. Currently, the next links, syntax highlighting and probably other things aren't working on those pages because of what I assume is a typo.


shuding:
Great catch, thank you!

PR #285, opened 2021-12-11
fix(Search): check for active results on `Enter` key

Dev-CasperTheGhost:
This fixes a minor bug when there are no active results, this PR checks if there are results and if there is an `results[active]`


shuding:
Thanks!

PR #286, opened 2021-12-13
Typescript, pnpm and Turbo

promer94:


promer94 (commented):
# Changes

* pnpm
* turbo
* use webpack native ```this.getOptions()``` instead of ```loader-utils```
* get titleText at compile time instead of runtime
* fix blog code line highlight
* add boring types

shuding:
Amazing work, thank you!

PR #312, opened 2022-01-05
Redesign

shuding:
- Shiki
- mdx-pretty-code
- Tailwind 3.0
- SWR-site example
- System theme
- FlexSearch
- Redesign
- Improved i18n

See it live on [swr.vercel.app](https://swr.vercel.app) (https://github.com/vercel/swr-site/pull/207).

---

Released as `nextra@alpha`, `nextra-theme-docs@alpha`. This doesn't include the blog theme yet. It also requires `pnpm install` right now so `yarn` doesn't work, this is a bug.


packages/nextra-theme-docs/src/utils/getHeadingText.ts
```diff
@@ -1,9 +1,5 @@
 import { Heading } from 'nextra'
 
 export default function getHeadingText(heading: Heading) {
-  if (Array.isArray(heading.children) && heading.children.length === 1) {
-    const content = heading.children[0]
-    if (content.type === 'text') return content.value
-  }
-  return ''
+  return heading.value || ''
```
promer94, line 4: This will make docs example like this



I think we should only render headers which are plain-text in TOC ?

packages/nextra/src/loader.ts
```diff
@@ -109,7 +115,10 @@ async function getPageMap(currentResourcePath: string): Promise<PageMapResult> {
       .map(item => {
         if (!item) return
         if (item.route === activeRoute) {
-          activeRouteTitle = dirMeta[item.name] || item.name
+          const metadata = dirMeta[item.name]
+          activeRouteTitle =
+            (typeof metadata === 'string' ? metadata : metadata.title) ||
```
promer94, line 120: The dirMeta could be `{}` in this suitation. So reading ```metadata.title``` is not safe here.

packages/nextra-theme-docs/src/utils/getHeadingText.ts
shuding: Thanks, just fixed it!

packages/nextra/src/locales.ts
```diff
@@ -0,0 +1,27 @@
...
+    !PUBLIC_FILE.test(request.nextUrl.pathname) &&
+    !request.nextUrl.pathname.includes('/api/') &&
+    !request.nextUrl.pathname.includes('/.nextra/')
+
+  if (shouldHandleLocale) {
+    let href = request.nextUrl.href || '/'
+    if (href.endsWith('/' + request.nextUrl.locale)) href += '/'
+    if (href.endsWith('/')) href += 'index'
+
+    const locale = '.' + request.nextUrl.locale
```
promer94, line 16: When preflight request have a cookie `NEXT_LOCALE = en`,
`request.nextUrl.locale` is still same as the accept-language here.  ( nextjs bug ?)

```suggestion
const locale = '.' + (request.cookies['NEXT_LOCALE'] || request.nextUrl.locale)
```

PR #314, opened 2022-01-07
Updated to auto prefix entire output

### LINKED ISSUE #247: Styling breaks when importing nextra stylesheet in an existing tailwind project
Hello,

I'm trying to integrate nextra docs into an existing nextjs tailwind project but importing the nextra stylesheet breaks/override some of the styles throughout the project. Removing the stylesheet seems to fix the issue but that's not really going to work for the docs.

I've tried copying all the styles from [here](https://github.com/shuding/nextra/blob/core/packages/nextra-theme-docs/src/styles.css) into a new stylesheet but importing that doesn't seem to work as well. 

This is what I see when I copy all the styles from the above link and import it 

Screenshot 2021-11-15 at 10 11 20 PM

These styles get overridden when importing `nextra-theme-docs/style.css`

Screenshot 2021-11-15 at 10 15 16 PM

Is there anyway to fix this? Appreciate any help! Thanks!

dashcraft:
Auto prefixed the entire output to be based on the top level class inside of the theme. This prevents styles from creeping over into global styles.

Fixes #247

### DIFF
--- a/packages/nextra-theme-docs/package.json
+++ b/packages/nextra-theme-docs/package.json
@@ -63,6 +63,7 @@
     "esbuild": "^0.14.0",
     "postcss": "^8.2.1",
     "postcss-cli": "^8.3.1",
+    "postcss-scopify": "0.1.9",
     "tailwindcss": "^2.2.4",
     "nextra": "workspace:*"
   }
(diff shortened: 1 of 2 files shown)


dashcraft:
Fixes #247

shuding:
Since we have selectors like `body` and even `.dark body` in styles.css, I'm wondering if postcss-scopify can handle it correctly.

dashcraft:
Great question, i am not currently testing it with the dark mode/theme selectors, i will say that is has added the container to HTML and body as well, so it's working correctly for those styles specifically. 

I'll test the dark/light mode and see what we can do.

B2o5T:
https://github.com/shuding/nextra/issues/492#issuecomment-1205884261

PR #346, opened 2022-01-28
[nextra-theme-blog] Allow forcing a theme

rgoomar:
I wasn't able to override the system theme with some script and figured this would be a good way to force it if you want light or dark. Currently, it defaults to system preference even with `darkMode` set to `false` or not configured.

### DIFF
--- a/packages/nextra-theme-blog/src/index.tsx
+++ b/packages/nextra-theme-blog/src/index.tsx
@@ -242,8 +242,8 @@ const NextraBlog = (opts: PageOpt, _config: NextraBlogTheme) => {
     return (
       <ThemeProvider
         attribute="class"
-        defaultTheme="system"
-        enableSystem={true}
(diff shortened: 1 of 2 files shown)


rgoomar:
Tested this locally with modifying `examples/blog/theme.config.js` with:
```js
forceTheme: 'light'
```
And running this and confirmed that even with system preference of dark, it defaults to light:
```bash
pnpm build:theme-blog
pnpm dev:theme-blog
```

gigamesh:
+1 - would love to be able to set the default to dark (using docs theme in my case)

dimaMachina:
`nextra-theme-blog` from Nextra 4 has `nextThemes` prop, so closing this PR https://github.com/shuding/nextra/blob/1e72319d01cc616e65b4723a02214ed81c224567/packages/nextra-theme-blog/src/components/layout.tsx#L23

PR #355, opened 2022-02-08
fix sort-date

jrvidal:
The comparison function needs to return 0 when elements are equal. Otherwise, when dates are equal (as in a previous version of the blog example), the ordering of the navbar becomes non-deterministic between engines.


shuding:
Thanks for the fix! I think it would be great to compare the title as the secondary key as the fallback — but this PR is good to land 👍

Issue #379, opened 2022-02-24
[nextra-theme-docs] - head to accept callback with frontMatter meta for SEO


Djuki:
At this moment, blog theme can accept callback with `title` and `meta` as arguments,

```
  head: ({ title, meta }) => (
    <>
      {meta.description && <meta name="description" content={meta.description} />}
      {meta.tag && <meta name="keywords" content={meta.tag} />}
      {meta.author && <meta name="author" content={meta.author} />}
    </>
  ),
```

This is not possible in docs theme. We can't use callback with docs theme. Is there any way to get frontMatter meta and set up different title/keywords and description for every page?

shuding:
Which version of Nextra are you on? I think the latest alpha/beta version has this supported as well: https://github.com/shuding/nextra/blob/dd5a08e0b3d817244b6bc90cdd7268af8b2355b5/packages/nextra-theme-docs/src/head.tsx#L27

Djuki:
@shuding 
```
    "nextra": "^2.0.0-beta.5",
    "nextra-theme-docs": "^2.0.0-beta.5",
```

Djuki:
@shuding I recorded a video that may help us to see where the issue is and to find is this a global issue or not.

https://www.loom.com/share/ad54d4c96c64400fb8d87aaa458d00a4

PR #410, opened 2022-03-14
the head config to support function in doc theme

Djuki:
@shuding I was solving my issue described here #379, and I found what is the issue.

This PR will add a support for passing a function in `config.head` configuration for Docs theme.

To support `config.head` as a function we need to have latest code from `core` branch for `renderComponent` metod:

```
const renderComponent = <T,>(
  ComponentOrNode: React.FC<T> | React.ReactNode,
  props: T,
  functionOnly?: boolean
) => {
  if (!ComponentOrNode) return null
  if (typeof ComponentOrNode === 'function') {
    if (functionOnly) return ComponentOrNode(props)
    return <ComponentOrNode {...props} />
  }
  return ComponentOrNode
}
```
Note that this is not a function we have in latest tag (`v2.0.0-beta.5`). The `functionOnly` is added later.

Important part is we have `functionOnly` and in case it is true we return ` ComponentOrNode(props)`. And that is what will make config.head as a function to be rendered into the Head.

Also we need to do what my PR suggest to call `renderComponent` with the `functionOnly` = `true`, because only in that case we will support both cases for `config.head`: function and Element.

Example of the `config.head` as a function we will support after this PR, and releasing new tag.

```
  head: ({title, meta}) => (
    <>
      <meta name="description" content={meta.description ? meta.description: "Default description"} />
    </>
  ),
```

### DIFF
--- a/packages/nextra-theme-docs/src/head.tsx
+++ b/packages/nextra-theme-docs/src/head.tsx
@@ -24,7 +24,7 @@ export default function Head({ title, locale, meta }: HeadProps) {
         {title}
         {renderComponent(config.titleSuffix, { locale, config, title, meta })}
       </title>
-      {renderComponent(config.head, { locale, config, title, meta })}
+      {renderComponent(config.head, { locale, config, title, meta }, true)}
       {config.unstable_faviconGlyph ? (
         <link
           rel="icon"


raproid:
@B2o5T @shuding What is the replacement for meta.description in beta 25? The page won't render showing me that it doesn't recognize meta.description in my theme config.

I'm using this example for testing: 
  head: ({title, meta}) => (
    <>
      <meta name="description" content={meta.description ? meta.description: "Default description"} />
    </>
  ),

B2o5T:
https://github.com/shuding/nextra/blob/09b4102c9141944b4b67b072cc5566993d78a44b/examples/swr-site/theme.config.tsx#L128

PR #417, opened 2022-03-16
usee cross-env to set env vars

aruniverse:
Found this "bug" where you can't build on windows on a fresh clone because it doesn't recognize that you're trying to set an env var. This should work cross-platform


shuding:
Thanks!

PR #439, opened 2022-04-05
Handle MDX parse errors gracefully during webpack compilation

### LINKED ISSUE #438: MDX parse errors cause the webpack compiler to hang
If there is an error during MDX parsing, such as if you use `<!-- markdown comment syntax -->` or an unknown language with your code fences like ```foobar, the whole webpack process hangs (even after you fix the problem) and you have to stop and restart the dev server to get going again.

Screen Shot 2022-04-05 at 7 46 28 PM
 
This looks to be because `loader.ts` exports an async function, which doesn't call `callback(err)` when it rejects.

I've got a fix coming up.

gkaemmer:
Fixes #438 

I tried a few different ways to solve this, and wrapping the whole async function seemed the most clean. Would love to get this in!


shuding:
Looks great, thank you!

shuding (approved):

PR #446, opened 2022-04-18
Add blue info colors to the callout component

harrybawsac:
Just a small styling addition to the callout component.


shuding:
Thanks!

PR #476, opened 2022-06-05
fix(theme-switch): cannot use of ARIA attribute without `role` attribute

numToStr:
Fixes the following lighthouse error around accessibility

> Each ARIA `role` supports a specific subset of `aria-*` attributes. Mismatching these invalidates the `aria-*` attributes.

And also fixes a warning in Chromium

> Elements must only use allowed ARIA attributes: aria-label attribute cannot be used on a span with no valid role attribute.


shuding:
Thank you!

shuding:
The current alpha for the blog theme has the latest code I think. Will confirm and do a release tomorrow if not!

PR #480, opened 2022-06-08
Layout improvements

### LINKED ISSUE #474: Page theme with layout:full, sidebar:false, toc:false results in horizontal page scrollbar
A page with an empty `div` (or anything) using this theme in its `meta.json` results in a page with a horizontal scrollbar.
```
"theme": {
  "layout": "full",
  "sidebar": false,
  "toc": false
}
```

Setting `sidebar: true` eliminates the scrollbar. This appears to be the case since `sidebar:false` sets the `expand` directive in `nextra-theme-docs/src/index.tsx`. My workaround is to set `sidebar:true` just to force the css state I want, but logically it feels wrong.

(Current as of core commit `f11f6d5`)

shuding:
Closes #474.

### DIFF
--- a/examples/swr-site/pages/meta.en-US.json
+++ b/examples/swr-site/pages/meta.en-US.json
@@ -9,8 +9,7 @@
     "title": "Docs",
     "type": "page",
     "theme": {
-      "pagination": true,
-      "sidebar": true
+      "toc": true
     }
   },
   "examples": {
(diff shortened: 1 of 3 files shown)


robbear:
Fantastic! Thank you, @shuding. Love the normalized layouts and thank you for the quick attention!

PR #481, opened 2022-06-09
fix(theme-blog): multiple elements have the same id

numToStr:
Removes the following warning in Elements tab in chromium

> id attribute value must be unique: Document has multiple static elements with the same id attribute

---

~Q) Should I remove the `const slug` and just use `props.id` as slug value in a follow-up PR?~ I removed the `const slug` because of https://github.com/shuding/nextra/pull/481#discussion_r912340686


shuding:
The anchor and id here is necessary so you can have this scroll to anchor effect: https://shud.in/posts/metaballs-1#shaders

packages/nextra-theme-blog/src/mdx-theme.tsx
```diff
@@ -45,7 +45,6 @@ const HeaderLink = ({
   const slug = slugger.slug(ReactDOMServer.renderToStaticMarkup(children))
   return (
     <Tag {...props}>
-      <span className="subheading-anchor" id={slug} />
       <a href={'#' + slug} className="subheading">
```
numToStr, line 48: One thing I noticed that `slug` variable changes every time page reload after change but `props.id` stays the same.

numToStr:
I have tested this change locally and scroll anchor seems to be working fine.

shuding:
The idea of that anchor element was to position the focus a little bit above the title, so it will not be too close to the edge of the viewport (doesn't look good).

numToStr:
> The idea of that anchor element was to position the focus a little bit above the title

If this was the idea then it's not working. Currently it anchors the heading at the top, instead of anchoring the `span` element (which I removed).

B2o5T:
closing since this issues was fixed in this PR https://github.com/shuding/nextra/pull/536

PR #487, opened 2022-06-16
chore(nextra): refactor `utils.ts`

B2o5T:


packages/nextra/src/utils.ts
```diff
@@ -1,41 +1,38 @@
...
 }
 
-export const existsSync = (f: string): boolean => {
+export const existsSync = (filePath: string): boolean => {
   try {
-    fs.accessSync(f, fs.constants.F_OK)
+    fs.accessSync(filePath, fs.constants.F_OK)
     return true
-  } catch (_) {
+  } catch {
```
B2o5T, line 35: can be just omitted

PR #507, opened 2022-07-06
fix(theme-blog): remove extra padding from post-item list

numToStr:
Using `<ul>` adds unnecessary left padding to the post item list and changing it to `<div>` removes the padding.

### Before

image

### After

image


shuding:
Good catch!

PR #512, opened 2022-07-07
chore(nextra-theme-docs): refactor `theme-context.ts`

B2o5T:
I renamed the extension to `ts` since there is no `jsx` content in this function


shuding (approved):
Changeset needed 👍

PR #517, opened 2022-07-08
feat(nextra): export `compile.tsx`

B2o5T:
`compileMdx` is a pretty useful function for me, but unfortunately, it is not exported by Nextra, and I use this function with [`next-mdx-remote`](https://github.com/hashicorp/next-mdx-remote)

P.S. also I found a way to remove dummy `context.js`, `data.js`, `index.js`, `locales.js` and `ssg.js` files by configuring the `exports` field in `package.json`

### DIFF
--- a/packages/nextra/context.js
+++ b/packages/nextra/context.js
@@ -1,1 +1,0 @@
-module.exports = require('./dist/context')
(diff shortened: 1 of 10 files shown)


shuding:
The exports field is much cleaner, thank you! I've been thinking about the ability to compile remote MDX data in Nextra too, and exporting `compile` is definitely a required step. However I'm blocked at the page map thing, which is a bit tricky. Anyway I'll merge this first!

PR #525, opened 2022-07-09
chore: extract `svg` icons in `/icons` folder, reusing same icons `<MoonIcon />` / `<SunIcon />` in blog from docs

B2o5T:
- extract all `svg` icons in `/icons` folder as react components
- bump `autoprefixer`, `postcss`, and `tailwindcss` that supports newly [arbitrary variants](https://tailwindcss.com/docs/hover-focus-and-other-states#custom-modifiers)
- reusing `<MoonIcon />` / `<SunIcon />` in blog from docs, but in docs use different style (like currently)
- export icons with named export to avoid typos

https://user-images.githubusercontent.com/7361780/178115825-d49b2582-0f16-426a-aba9-a103d47515ec.mov

https://user-images.githubusercontent.com/7361780/178115826-7c74567a-059b-4996-9765-4e8421b64c45.mov

PR #509, opened 2022-07-07
feat(nextra-theme-docs): bump `rehype-pretty-code` version, support `showLineNumbers`

B2o5T:
related https://github.com/shuding/nextra/pull/501

PR #527, opened 2022-07-10
chore(docs/blog) reuse `useMounted` hook to avoid code duplication

B2o5T:


packages/nextra-theme-blog/src/theme-switch.tsx
```diff
@@ -1,11 +1,10 @@
-import React, { useState, useEffect } from 'react'
+import React from 'react'
 import { useTheme } from 'next-themes'
+import { useMounted } from '../../nextra-theme-docs/src/utils/use-mounted'
```
B2o5T, line 3: does there is a better way to share some code between blog and docs?
- In this PR I need `useMounted` from docs
- In this PR https://github.com/shuding/nextra/pull/525 I need `<MoonIcon />` and `<SunIcon />` from docs
- And in this PR https://github.com/shuding/nextra/pull/509 we can share code-blocks styles to avoid code duplications

shuding: Since users won't install both blog and docs in one app, I think we can put these outside of these 2 packages, but in a separate folder:

```
packages
  nextra
  nextra-theme-docs
  nextra-theme-blog
  utils
    icons
    hooks
```

Here `utils` is not a package but just a folder. Will that work?

B2o5T: currently, this will work, let's wait for a response from @promer94 since he worked on https://github.com/shuding/nextra/pull/569

B2o5T:
close it for now

PR #534, opened 2022-07-13
chore(blog/docs): use `postcss-import` to import css variables styles

B2o5T:
also, we can share `code-blocks` styles to avoid code duplications but in another PR

### DIFF
--- a/packages/nextra-theme-docs/src/styles.css
+++ b/packages/nextra-theme-docs/src/styles.css
@@ -1,4 +1,15 @@
-@tailwind base;
+@import 'tailwindcss/base';
+@import 'tailwindcss/components';
+@import 'tailwindcss/utilities';
+@import './variables.css';
+
+:root {
+  --nextra-primary-hue: 212deg;
+}
+
+.dark {
+  --nextra-primary-hue: 204deg;
+}
 
 html {
   @apply antialiased;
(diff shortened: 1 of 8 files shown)


packages/nextra-theme-docs/src/styles.css
```diff
@@ -1,4 +1,15 @@
...
+@import 'tailwindcss/utilities';
+@import './variables.css';
+
+:root {
+  --nextra-primary-hue: 212deg;
+}
+
+.dark {
+  --nextra-primary-hue: 204deg;
+}
```
B2o5T, line 12: only these variables are included in next-theme-docs

packages/nextra-theme-docs/package.json
```diff
@@ -61,7 +61,6 @@
   "devDependencies": {
     "@types/flexsearch": "^0.7.2",
     "@types/react": "^17.0.39",
-    "autoprefixer": "^10.2.6",
```
B2o5T, line 64: don't need in docs since `autoprefixer` is already installed in the root `package.json`

B2o5T:
@shuding can I merge this one? and I'll extract code-styles in follow PR

shuding:
Yeah sounds good to ship it 👍❤️

PR #535, opened 2022-07-13
fix(blog): nav styles when there are many links

B2o5T:
also I moved nav/meta styles from `style.css` to component `className` directly

before:
image
after:
image

### DIFF
--- a/packages/nextra-theme-blog/src/meta.tsx
+++ b/packages/nextra-theme-blog/src/meta.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { ReactElement } from 'react'
 import Link from 'next/link'
 import ThemeSwitch from './theme-switch'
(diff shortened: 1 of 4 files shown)


packages/nextra-theme-blog/src/meta.tsx
```diff
@@ -11,31 +11,56 @@ interface MeatProps {
...
-              <a className="tag">{t}</a>
-            </Link>
-          )
-        })}
+    <div className="flex items-center mb-8 gap-3">
+      <div className="flex-1 text-gray-400 gap-1 flex items-center flex-wrap">
+        {author}
+        {author && date && ','}
+        {date && (
+          <time dateTime={new Date(date).toISOString()}>
```
B2o5T, line 30: i guess it's better also provide date in `datetime` attribute

PR #541, opened 2022-07-15
chore(nextra-theme-docs): provide type for `DocsThemeConfig.nextThemes` instead of `object`

B2o5T:
image

### DIFF
--- a/packages/nextra-theme-docs/src/index.tsx
+++ b/packages/nextra-theme-docs/src/index.tsx
@@ -254,7 +254,7 @@ interface DocsLayoutProps extends PageOpt {
 }
 const createLayout = (opts: DocsLayoutProps, _config: DocsThemeConfig) => {
   const extendedConfig = Object.assign({}, defaultConfig, _config, opts)
-
+  const nextThemes = extendedConfig.nextThemes || {}
   const Page = ({ children }: { children: React.ReactChildren }) => children
 
   Page.getLayout = (page: any) => (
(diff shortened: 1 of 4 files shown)


packages/nextra-theme-docs/src/index.tsx
```diff
@@ -177,28 +177,27 @@ const Content: React.FC<LayoutProps> = ({
...
-    <React.Fragment>
-      <Head title={title} locale={locale} meta={meta} />
       <MenuContext.Provider
         value={{
           menu,
           setMenu,
           defaultMenuCollapsed: !!config.defaultMenuCollapsed
         }}
       >
+        <Head title={title} locale={locale} meta={meta} />
```
B2o5T, line 187: since Head uses NextHead we can put it anywhere

packages/nextra-theme-docs/src/index.tsx
```diff
@@ -247,17 +246,17 @@ const Content: React.FC<LayoutProps> = ({
...
+          )}
         </div>
       </MenuContext.Provider>
-    </React.Fragment>
   )
 }
 
 const createLayout = (opts: PageOpt, _config: DocsThemeConfig) => {
-  const extendedConfig = Object.assign({}, defaultConfig, _config)
+  const extendedConfig = { ...defaultConfig, _config }
```
B2o5T, line 258:
```suggestion
  const extendedConfig = { ...defaultConfig, ..._config }
```

packages/nextra-theme-docs/src/index.tsx
```diff
@@ -173,28 +173,27 @@ const Content: React.FC<LayoutProps> = ({
...
         >
-          {themeContext.navbar ? (
+          {themeContext.navbar && (
             <Navbar
               isRTL={isRTL}
               flatDirectories={flatDirectories}
               items={topLevelPageItems}
             />
-          ) : null}
+          )}
```
shuding, line 196: I prefer `?` over `&&` because if the condition `themeContext.navbar` is number `0` then React will render "0" here.

B2o5T: @shuding fixed for all modified cases, also I fixed types for PageTheme 
image
since like regression by @promer94 rebase

Issue #542, opened 2022-07-15
🐛 bug(nextra): `next build` fails if `pages/api` folder contains `md`/`mdx` files


B2o5T:
```sh
info  - Creating an optimized production build  
Failed to compile.

./src/pages/api/bug.mdx
TypeError: Cannot read properties of undefined (reading 'name')
    at getPageMap (file:///Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/packages/nextra/dist/loader.mjs:703:116)
    at file:///Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/packages/nextra/dist/loader.mjs:824:37
    at Generator.next (<anonymous>)
    at file:///Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/packages/nextra/dist/loader.mjs:37:61
    at new Promise (<anonymous>)
    at __async (file:///Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/packages/nextra/dist/loader.mjs:21:10)
    at loader (file:///Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/packages/nextra/dist/loader.mjs:805:10)
    at Object.syncLoader (file:///Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/packages/nextra/dist/loader.mjs:920:3)
    at /Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/packages/nextra/loader.js:14:20
    at runMicrotasks (<anonymous>)


> Build failed because of webpack errors
/Users/dimitri/Desktop/GUILD/graphql-code-generator/nextra/examples/docs:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  docs@1.0.0 build: `next build`
Exit status 1
```

I know there is no reason to have markdown files here, but nextra should not fail if any exists :)

B2o5T:
Closed as completed https://github.com/shuding/nextra/pull/543

PR #543, opened 2022-07-15
fix(nextra): ignore in loader files from `pages/api` folder

B2o5T:
fixes https://github.com/shuding/nextra/issues/542
ignore but print warning
image

### Update: Or maybe we should throw but with a friendly error? and not with `TypeError`


shuding:
I think it’s good to print the message 👍

We might need a `quiet` option in the future

PR #544, opened 2022-07-15
Support new middleware cookies api

huozhi:
x-ref: https://github.com/vercel/swr-site/pull/307#issuecomment-1185842254

* handle with/without trailing slash in url cases to extract correct locale
* handle differen cookies APIs
* add locale tests

rewrite the lcoale extracting from url logic, previously the pathname can be `/`, then the url with host removed can be `zh-CN` or `zh-CN/` in with next version (12.2 vs 12.1). We'd better to find the first slash and get extract the locale from it

Thanks @promer94 for helping with setting up locale tests


promer94 (changes requested):
Currently the preview is broken 
https://nextra-theme-docs-dev-git-fork-huozhi-fix-next-middleware-api.vercel.sh/zh-CN

I think the reason is that
https://github.com/shuding/nextra/blob/6d5c85e7835a1251307a9fcc2818d07ba72b604a/packages/nextra/src/locales.ts#L16
`fullUrl` is different between `next 12.1` and `next 12.2`

When user access `https://swr.vercel.app/zh-CN`

* 12.1
`fullUrl` would be `https://swr.vercel.app/zh-CN/`

* 12.2
`fullUrl` would be `https://swr.vercel.app/zh-CN` 
It doesn't have trailing slash

packages/nextra/src/locales.ts
```diff
@@ -27,9 +43,9 @@ export function locales(request: NextRequest) {
     // If there is a locale cookie, we try to use it. If it doesn't exist or
     // it's invalid, `nextUrl.locale` will be automatically figured out by Next
     // via the `accept-languages` header.
-    if (request.cookies['NEXT_LOCALE']) {
+    if (getCookie(request.cookies, 'NEXT_LOCALE')) {
       try {
-        nextUrl.locale = request.cookies['NEXT_LOCALE']
+        nextUrl.locale = getCookie(request.cookies, 'NEXT_LOCALE')!
```
B2o5T, line 48: No need to call 2nd time, you can assign to a variable

promer94 (approved):

PR #552, opened 2022-07-20
Update discord.tsx

quintenvandamme:
Update discord logo to the new one.

> old
image

> new
image


packages/nextra-theme-docs/src/icons/github.tsx
```diff
@@ -1,14 +1,16 @@
...
-        fill="currentColor"
-      />
+    <svg
+      width="24"
+      height="24"
+      fill="currentColor"
+      viewBox="3 3 18 18"
```
B2o5T, line 9: I recalculated viewBox based on `<path/>` element as previous values were incorrect, not 24 but 21.6, example from turborepo 
image
and the new one:
image
found on StackOverflow https://stackoverflow.com/a/10259465

B2o5T:
Thanks!

packages/nextra-theme-docs/src/icons/github.tsx
shuding: Thanks!

PR #567, opened 2022-07-22
Apply ResizeObserver to main element in nextra-theme-docs

### LINKED ISSUE #551: ToC sync/scroll semantics appear to cause section link targets to scroll out of view
Example: Click on the link: https://docs.relational.ai/rel/ref/data-types#entity and notice the ensuing scroll activity, resulting in the target becoming out of view.

Direct clicks on the ToC properly bring the user to the associated anchored element. Note that there are custom elements on the page that get populated with data, acting similar to an `` element in expanding the height of a page, which may have an impact here.

If I stub out the ToC and build, I get right to the section link as expected.

robbear:
Fixes #551 

Custom elements contained as children of `<main>`  may asynchronously resize, affecting the scroll position for a hash section link. `https://swr.vercel.app/docs/getting-started#example` is an example of page with a hash section link.

When those custom elements hydrate, the resulting scroll position of the page may end up nowhere near the actual link in the content. By setting up a `ResizeObserver`, the callback can force the scroll position back to the appropriate spot if the current page URL contains a hash and the size of the `<main>` element changes due to, for example, hydration of its children.


packages/nextra-theme-docs/src/index.tsx
```diff
@@ -22,6 +22,20 @@ import './polyfill'
...
+let resizeObserver: ResizeObserver
+if (typeof window !== 'undefined') {
+  resizeObserver =
+    resizeObserver! ||
+    new ResizeObserver(entries => {
+      if (window.location.hash) {
+        entries[0]
+        .target
+        .ownerDocument
+        .querySelector(window.location.hash)?.scrollIntoView()
```
promer94, line 34: I think you could use `scroll-into-view-if-needed` here since it's already installed in docs theme

promer94:
Would you mind adding a page that could reproduce the problem to our examples as a kind of test case.  
It would be really helpful 😊

robbear:
@promer94 Apologies if it's obvious, but could you point out the path to where you'd like me to leave the test case?

promer94:
`examples/swr-site/pages/docs/scroll-with-dynamic-height.en-US.mdx`

robbear:
@promer94 I've added the test page, including language variations, but I'm apparently not following the process correctly as the CI test fails. 

I've also made the modification you requested, using `scroll-into-view-if-needed` rather than the native call.

promer94 (approved):

Issue #565, opened 2022-07-22
Getting "Text content does not match server-rendered HTML" on beta version


chilikasha:
Hi,
Updated to beta from v1 and getting 2 errors when opening site:
"Text content does not match server-rendered HTML"
and 
"There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering"

```
Warning: Text content did not match. Server: "Last updated on May 1, 2022" Client: "Last updated on 1 May 2022"
div
main
article
Body@webpack-internal:///./node_modules/nextra-theme-docs/dist/index.js:2227:12
```

```
"nextra": "^2.0.0-beta.6",
"nextra-theme-docs": "^2.0.0-beta.6",
```

B2o5T:
Will be fixed in next release

chilikasha:
```
Error: There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering.
```
@B2o5T hey, should this one also be fixed? still getting this on beta 11

B2o5T:
@chilikasha needs more details where error occurs or reproduction, examples work fine

chilikasha:
@B2o5T unfortunately I haven't pushed these changes, still working on them. but can provide the following details:
_app.js file
```js
import 'nextra-theme-docs/style.css'
import '../public/style.css'
import { usePanelbear } from '@panelbear/panelbear-nextjs'

export default function Nextra({ Component, pageProps }) {
  usePanelbear('JHnEG7vc0FG', {
    debug: false
  })
  const getLayout = Component.getLayout || ((page) => page)
  return getLayout(<Component {...pageProps} />)
}
```

package.json
```
"next": "^12.1.6",
    "nextra": "^2.0.0-beta.11",
    "nextra-theme-docs": "^2.0.0-beta.11",
    "react": "^18.0.0",
    "react-dom": "^18.0.0"
```

I do `yarn next` and open localhost:3000. usually it opens without any errors. but then I just refresh the page and result below on this video: https://jmp.sh/vS6p0Ez

please let me know if I can provide smth else

B2o5T:
@chilikasha have you tried to comment usePanelbear hook?

chilikasha:
@B2o5T just tried, no difference.
what I've mentioned is: I have some span elements (visually look like tags) in mdx files. code is like:
```html
<div><span className="tag">macOS</span><span className="tag">free</span></div>
```
[...]

PR #568, opened 2022-07-22
fix client console error on React 17 / crash on React 18 - Text content does not match server-rendered HTML

B2o5T:
fixes https://github.com/shuding/nextra/issues/565

The error appears since in this place `locale` is `undefined` if `nextConfig.i18n` was not configured
https://github.com/shuding/nextra/blob/e35bbf7e363c83061c1267c4ebb9369a0b070878/packages/nextra-theme-docs/src/index.tsx#L108-L112

now console is clean
image

Also, I typescripified `index.js` to make it type-safety


packages/nextra/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+            themeConfig: args[1] as string
+          }
+        : args[0]
+
+    const nextraPlugin = new NextraPlugin(nextraConfig)
+    const { i18n, pageExtensions = DEFAULT_EXTENSIONS } = nextConfig
+
+    if (i18n?.locales) {
+      console.log(
+        '[nextra] You have Next.js i18n enabled, read here https://nextjs.org/docs/advanced-features/i18n-routing for the docs.'
```
B2o5T, line 23: replaced todo link

shuding: Sounds good! When Nextra has a good i18n docs page we should link there

shuding (approved):

PR #521, opened 2022-07-09
chore: use single `build` scripts with `--dev` flag

B2o5T:
to avoid code duplication, as `dev.js` is mostly the same as `build.js`

PR #569, opened 2022-07-23
feat: change build system to tsup

promer94:


packages/nextra/src/icon-components/arrow-right.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ComponentProps } from 'react'
```
B2o5T, line 1: I think it’s not good place for shared icons, since it’s not related to nextra, but for themes only

promer94: My point is that 
 if user want to build their own themes. They could import icons from nextra.

package.json
```diff
@@ -33,15 +34,19 @@
...
     "semi": false,
     "singleQuote": true,
     "trailingComma": "none",
     "arrowParens": "avoid"
   },
-  "packageManager": "pnpm@7.3.0"
-}
+  "packageManager": "pnpm@7.3.0",
+  "dependencies": {
+    "swr": "^1.3.0"
```
shuding: We’re not using SWR here right?

promer94: It should be removed now.

packages/nextra/.gitignore
```diff
@@ -0,0 +1 @@
+icons/**
```
B2o5T, line 1: Why it’s needed?

packages/nextra/tsup.config.ts
```diff
@@ -0,0 +1,30 @@
...
+  name: 'nextra-loader',
+  dts: true
+},
+{
+  entry: ['src/compile.ts'],
+  format: 'esm',
+  name: 'compile',
+  dts: true
+}, {
+  entry: ['src/icons.ts'],
```
B2o5T, line 26:
```suggestion
  entry: ['src/icons/index.ts'],
```
and you can remove `src/icons.ts` I guess?

packages/nextra/.gitignore
promer94: should be remove

packages/nextra/tsup.config.ts
promer94: we have use icons.ts here since we want it output `icon.mjs` in dist

packages/nextra-theme-blog/src/meta.tsx
```diff
@@ -8,7 +8,7 @@ interface MeatProps {
   author: string
```
B2o5T, line 8: above you have a typo - Meat 🥩

packages/nextra-theme-blog/src/meta.tsx
```diff
@@ -4,11 +4,11 @@ import ThemeSwitch from './theme-switch'
 import type { NextraBlogTheme } from './types'
 import { split } from './utils/get-tags'
 
-interface MeatProps {
```
shuding, line 7: 🤣

shuding (approved):
Looks good to me! Seems like a good idea to put shared things in `nextra/`.

shuding:
This also affects #521 I think.

packages/nextra-theme-docs/src/types.ts
```diff
@@ -8,7 +8,7 @@ export interface DocsThemeConfig {
   titleSuffix?:
     | React.ReactNode
     | React.FC<{
-        locale: string
+        locale?: string
```
B2o5T, line 11: locale now always provided so you need to replace all `locale?: string` by `locale: string`

packages/nextra/src/env.d.ts
```diff
@@ -4,6 +4,8 @@ declare module 'github-slugger' {
   }
 }
 
+declare module 'rehype-mdx-title'
```
B2o5T, line 7: why? this package provide declarations

packages/nextra/src/index.js
```diff
@@ -1,17 +1,16 @@
 import { NextraPlugin, pageMapCache } from './plugin'
 import { MARKDOWN_EXTENSION_REGEX } from './constants'
-import { LoaderOptions, WithNextra } from './types'
 
-const DEFAULT_EXTENSIONS = ['js', 'jsx', 'ts', 'tsx'] as const
-const MARKDOWN_EXTENSIONS = ['md', 'mdx'] as const
```
B2o5T, line 6: hm, previously this file worked well with typescript, just need to use `module.exports` and not `export default`

packages/nextra/src/types.ts
```diff
@@ -60,3 +80,7 @@ export type NextraConfig = {
 export type Nextra = (
   ...args: [NextraConfig] | [theme: Theme, themeConfig: string]
 ) => (nextConfig: NextConfig) => NextConfig
+
+const nextra: Nextra = (...args) => (nextConfig) => ({})
+
+export default nextra
```
B2o5T, line 86: why it's needed?

packages/nextra/src/index.js
promer94: Before:
Screen Shot 2022-07-24 at 02 16 01

After:
Screen Shot 2022-07-24 at 02 19 04

packages/nextra-theme-docs/src/index.tsx
```diff
@@ -29,7 +29,7 @@ function useDirectoryInfo(pageMap: PageMapItem[]) {
     const fsPath = getFSRoute(asPath, locale)
     return normalizePages({
       list: pageMap,
-      locale,
+      locale: locale ? locale : 'en-US',
```
B2o5T, line 32: `en-US` is always will be by default see https://github.com/shuding/nextra/pull/568

packages/nextra-theme-docs/src/index.tsx
```diff
@@ -140,7 +140,7 @@ const Content: React.FC<LayoutProps> = ({
   timestamp,
   children
 }) => {
-  const { route, locale } = useRouter()
+  const { route, locale = 'en-US' } = useRouter()
```
B2o5T, line 143: here also

packages/nextra-theme-docs/src/index.tsx
promer94: Nextjs locale type is string | undefined here

packages/nextra/src/types.ts
promer94: The key problem here is that we want expose all types which are located in `types.ts` but you can't do that with commonjs.

So i compile `types.ts` to a d.ts file and use packjson's `types` field to let typescript resolve the correct type here. 

```ts
// I have to provider a fake function here otherwise typescript will think i only import a type here
import nextra from 'nextra'
```

PR #520, opened 2022-07-09
WIP: New website for v2

B2o5T:

PR #574, opened 2022-07-23
fix(nextra/docs): fallback search to `en-US` instead `default`

B2o5T:
related https://github.com/shuding/nextra/pull/520#issuecomment-1193168203

### DIFF
--- a/packages/nextra-theme-docs/src/flexsearch.js
+++ b/packages/nextra-theme-docs/src/flexsearch.js
@@ -6,7 +6,7 @@ import React, {
   useEffect,
   Fragment
 } from 'react'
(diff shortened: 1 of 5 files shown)


packages/nextra-theme-docs/src/flexsearch.js
```diff
@@ -34,7 +34,7 @@ const Item = ({
           {page}
         </div>
       ) : null}
-      <Link href={Router.basePath + href}>
+      <Link href={href}>
```
B2o5T, line 37: put it here 
image
since no need to use `Router` and `router` from `useRouter`

shuding (approved):

PR #581, opened 2022-07-24
chore: fix build error and format code

promer94:


packages/nextra-theme-docs/src/search.tsx
```diff
@@ -46,7 +46,7 @@ const Search = ({ directories = [] }: SearchProps) => {
   const [search, setSearch] = useState('')
   const [active, setActive] = useState<number | null>(null)
   const input = useRef<HTMLInputElement | null>(null)
-
+  const { locale = 'en-US'} = useRouter()
```
B2o5T, line 49: It’s redundant to have 2 useRouter calls, already exist on line 43

packages/nextra/package.json
```diff
@@ -88,10 +89,16 @@
   "peerDependencies": {
     "react": ">=16.13.1"
   },
+  "prettier": {
```
B2o5T, line 92: It’s better to have 1 top level prettier config, rather than add prettier config for each monorepo package

.prettierignore
```diff
@@ -2,4 +2,5 @@ dist
 .next
 pnpm-lock.yaml
 pnpm-workspace.yaml
-.changeset
\ No newline at end of file
+.changeset
+.husky
```
B2o5T, line 6:
```suggestion
.husky
*.mdx
```
Since prettier don’t support mdx2

packages/nextra/package.json
promer94: It is used for [prettier-plugin-tailwindcss](https://github.com/tailwindlabs/prettier-plugin-tailwindcss)

Issue #577, opened 2022-07-24
[theme-blog] Allow `footer` to be customized


numToStr:
### Problem

Currently, `footer` in `theme.config.js` is a jsx element. Which prevents customization on per-page basis. For example, I would like to have a `Edit on Github` or `Spot a Typo?` button on every posts in my blog to let the reader submit patches.

```js
export default {
    footer: (
        /* ... */
    )
}
```

### Solution

Make `footer` a jsx component, In that way we can customize the footer on per-page basis by providing necessary props to it, like the `head` config.

```js
export default {
    footer(props) (
        /* ... */
    )
}
```

> **NOTE**: To cover my use case we also need to provide the file path in the footer. Though I am not sure how much work this requires.

Issue #538, opened 2022-07-14
[theme-blog] Make `react-cusdis` optional


numToStr:
### Problem

For people who want a simple blog without any comment system, installing `react-cusdis` is something that can be avoided. Also `pnpm` emits the following warning for `react-cusdis`

```
|- nextra-theme-blog
  |- react-cusdis
    |- unmet peer react@^17.0.0: found 18.2.0
    |- unmet peer react-dom@^17.0.0: found 18.2.0
```

### Solution

Make `react-cusdis` and comment system optional.

PR #602, opened 2022-07-27
refactor: blog theme

promer94:


packages/nextra/src/mdx-plugins/remark.ts
```diff
@@ -45,11 +46,20 @@ export function remarkHeadings(this: Processor) {
...
+            node.children.some(
+              (child: any) => (child.type as string) === 'mdxJsxTextElement'
+            )
           const heading = {
             ...(node as Heading),
             value: getFlattenedValue(node)
           }
           data.headingMeta.headings.push(heading)
+          if (jsxH1) {
+            data.headingMeta.jsxH1 = true
```
B2o5T, line 61: could you also add a comment on why this prop is needed?

promer94:
```mdx
import { useRouter } from 'next/router'

export const TagName = () => {
  const { tag } = useRouter().query
  return tag || null
}

# Posts Tagged with “<TagName/>”
```

For headings contains jsx

numToStr:
Is it possible to add support for #577 and #538?

FYI, I already created a PR #595

promer94:
> FYI, I already created a PR #595 to update the footer.

After this pr. 

You should be able to customize footer and other part with blog context

```jsx
import { useBlogContext } from 'nextra-themes-blog'
const Footer = () => {
  const { opt, config } = useBlogConfig()
}
export default {
  footer: <Footer/>
}
```

numToStr:
> > FYI, I already created a PR #595 to update the footer.
> 
> After this pr. 
> 
> You should be able to customize footer and other part with blog context
> 

I was thinking of more like this, making footer a jsx component instead of jsx element. It would be same as the `head` config, and I think docs theme support this syntax.

```jsx
import { useBlogContext } from 'nextra-themes-blog'
const Footer = () => {
  const { opt, config } = useBlogConfig()
}
export default {
  footer: Footer
}
```

promer94:
Footer does not need to be component here. 
It could access all the meta data from context and it is more flexible when it compares to passing data through props

numToStr:
@promer94 Accessing meta data via context is nice but I wonder will this support calling `useRouter()` inside the footer. As I need `pathname` to support my actual use case of having a `Edit on Github` button on every posts in my blog to let the reader submit patches.

promer94:
> @promer94 Accessing meta data via context is nice but I wonder will this support calling `useRouter()` inside the footer. As I need `pathname` to support my actual use case of having a `Edit on Github` button on every posts in my blog to let the reader submit patches.

It is supported. `useRouter` and `useBlogContext` are both react hooks, in fact you could use any react hooks here.

numToStr:
Thanks for the clarification. I'll try doing that.

B2o5T:
@promer94 will test with a review tomorrow (today) 🌚

packages/nextra-theme-blog/src/basic-layout.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const { jsxH1 } = opts
+  const portalHeading = jsxH1
+  return (
+    <article className="container prose prose-sm dark:prose-dark md:prose">
+      <Head>
+        <title>{title}</title>
+        {config.head?.({ title, meta: opts.meta })}
+      </Head>
+      <HeadingContext.Provider value={ref}>
+        {portalHeading ? <h1 ref={ref}></h1> : null}
```
B2o5T, line 20:
```suggestion
        {portalHeading ? <h1 ref={ref} /> : null}
```

B2o5T: You can also use ternary to render either h1 with/without ref

packages/nextra-theme-blog/src/posts-layout.tsx
```diff
@@ -0,0 +1,63 @@
...
+import MDXTheme from './mdx-theme'
+import Nav from './nav'
+import { collectPostsAndNavs } from './utils/collect'
+import getTags from './utils/get-tags'
+
+export const PostsLayout = ({ children }: { children: ReactNode }) => {
+  const { config, opts } = useBlogContext()
+  const { posts } = collectPostsAndNavs({ config, opts })
+  const router = useRouter()
+  const type = opts.meta.type
```
B2o5T, line 15:
```suggestion
  const { type } = opts.meta
```

packages/nextra-theme-blog/src/utils/collect.ts
```diff
@@ -0,0 +1,34 @@
...
+    !page.frontMatter ||
+    !page.frontMatter.type ||
+    page.frontMatter.type === 'post'
+  )
+}
+
+export const collectPostsAndNavs = ({ opts }: LayoutProps) => {
+  const posts: PageMapItem[] = []
+  const navPages: PageMapItem[] = []
+  const route = opts.route
```
B2o5T, line 22:
```suggestion
  const { route } = opts
```

packages/nextra/src/mdx-plugins/remark.ts
```diff
@@ -45,11 +46,20 @@ export function remarkHeadings(this: Processor) {
       },
       node => {
         if (node.type === 'heading') {
+          const jsxH1 =
```
B2o5T, line 49: Maybe more clear name hasJsxInH1?

packages/nextra-theme-blog/src/index.tsx
```diff
@@ -1,180 +1,30 @@
...
-      attrs={{
-        host: config.cusdis.host || 'https://cusdis.com',
-        appId: config.cusdis.appId,
-        pageId: router.pathname,
-        pageTitle,
-        theme: theme === 'dark' || resolvedTheme === 'dark' ? 'dark' : 'light'
-      }}
-    />
-  )
+  const Layout = layoutMap[type]
```
B2o5T, line 23: Need add check `if (!Layout) {...` with throwing friendly error

packages/nextra-theme-blog/src/blog-context.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const BlogContext = createContext<LayoutProps | null>(null)
+
+export const BlogProvider: React.FC<PropsWithChildren<LayoutProps>> = ({
+  config,
+  children,
+  opts
+}) => {
+  return (
+    <BlogContext.Provider value={{ config, opts }}>
```
B2o5T, line 13: Value should be memoized otherwise every rerender - new object

packages/nextra-theme-blog/src/cusdis.tsx
```diff
@@ -0,0 +1,35 @@
...
+  { ssr: false }
+) as typeof ReactCusdis
+
+export const Comments = () => {
+  const { config, opts } = useBlogContext()
+  const router = useRouter()
+  const { pageTitle } = getTitle({ opts, config })
+  const { theme, resolvedTheme } = useTheme()
+  if (config.cusdis && !config.cusdis.appId) {
+    console.warn('[cusdis]', '`appId` is required')
```
B2o5T, line 20:
```suggestion
    console.warn('[nextra/cusdis]', '`appId` is required')
```

packages/nextra-theme-blog/src/blog-context.tsx
promer94: It's a top level component. It probably wont rerender often.

packages/nextra-theme-blog/src/basic-layout.tsx
promer94: I can't use ternary here since the content of h1 is from React.createportal. It needs to be removed with h1 node.

packages/nextra-theme-blog/src/basic-layout.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const portalHeading = hasJsxInH1
+  return (
+    <article className="container prose prose-sm dark:prose-dark md:prose">
+      <Head>
+        <title>{title}</title>
+        {config.head?.({ title, meta: opts.meta })}
+      </Head>
+      <HeadingContext.Provider value={ref}>
+        {portalHeading ? <h1 ref={ref} /> : null}
+        {!portalHeading ? <h1>{pageTitle}</h1> : null}
```
B2o5T, line 21: I mentioned ternary for this part
```suggestion
        {portalHeading ? <h1 ref={ref} /> : <h1>{pageTitle}</h1>}
```

B2o5T: also, you don't need to apply suggestions manually just click to Commit Suggestion button
image

packages/nextra-theme-blog/src/blog-context.tsx
B2o5T: it's will be removed https://github.com/shuding/nextra/pull/602#discussion_r932204813

packages/nextra-theme-blog/src/basic-layout.tsx
B2o5T: it's will be removed https://github.com/shuding/nextra/pull/602#discussion_r932204813

packages/nextra-theme-blog/src/basic-layout.tsx
promer94: Screen Shot 2022-07-28 at 23 57 51

if you do that, it will cause runtime error here. 

Like i said before. You have to make sure everytime portalHeading changes, a new h1 ReactNode will be created instead of reusing the current one

PR #606, opened 2022-07-28
fix `Hydration failed because the initial UI...` while modifying `meta.json` on dev env and react 18

B2o5T:


packages/nextra/src/loader.ts
```diff
@@ -86,23 +86,14 @@ async function loader(
     defaultLocale
   )
 
-  if (IS_PRODUCTION) {
-    // We only add meta files as dependencies for production build,
-    // so we can do incremental builds.
-    Object.entries(fileMap).forEach(([filePath, { name, meta, locale }]) => {
-      if (
-        name === 'meta.json' &&
-        meta &&
```
B2o5T, line 95: I removed `meta` check since it's redundant it will be always trurhy if the error throws during parse

https://github.com/shuding/nextra/blob/417918abb9c87f6862a6f24115ee092e03d49f5b/packages/nextra/src/utils.ts#L24-L31

PR #611, opened 2022-07-29
[nextra-theme-docs]: fix `Warning: A title element received an array` and possible `[object Object]` in title

B2o5T:
fixes console warning
image
and possible issue when declaring `titleSuffix` as function
image

### DIFF
--- a/packages/nextra-theme-docs/src/head.tsx
+++ b/packages/nextra-theme-docs/src/head.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import NextHead from 'next/head'
 import { useTheme } from 'next-themes'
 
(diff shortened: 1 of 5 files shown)


packages/nextra-theme-docs/src/head.tsx
```diff
@@ -16,23 +16,30 @@ export default function Head({ title, locale, meta }: HeadProps) {
...
 
   return (
     <NextHead>
-      <title>
-        {title}
-        {renderComponent(config.titleSuffix, { locale, config, title, meta })}
-      </title>
-      {renderComponent(config.head, { locale, config, title, meta }, true)}
+      <title>{title + (suffix || '')}</title>
+      {renderComponent(config.head, { locale, config, title, meta })}
```
B2o5T, line 30: I removed `functionOnly` argument since no need to have it in this place

packages/nextra-theme-docs/src/utils/render-component.tsx
```diff
@@ -1,13 +1,11 @@
...
-  props: PropsWithChildren<T>,
-  functionOnly?: boolean
-) => {
+function renderComponent<T>(
+  ComponentOrNode: FC<PropsWithChildren<T>> | ReactNode,
+  props: PropsWithChildren<T>
+) {
   if (!ComponentOrNode) return null
   if (typeof ComponentOrNode === 'function') {
-    if (functionOnly) return ComponentOrNode(props)
```
B2o5T, line 10: this parameter is used only in `<Head />` but for which reason?

shuding (approved):
Thank you for fixing this error, it's annoying

Issue #597, opened 2022-07-27
[theme-blog] Unable to build with `v2.0.0-beta.10`


numToStr:
### Problem

Today I updated my blog to `beta.10` and got the following error. With `beta.9`, It is working fine.

```
error - Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/mario/Code/vikasraj.dev/node_modules/.pnpm/nextra-theme-blog@2.0.0-beta.10_53oqlnqnxhbetdy4ljlkkvmoby/node_modules/next/head' imported from /home/mario/Code/vikasraj.dev/node_modules/.pnpm/nextra-theme-blog@2.0.0-beta.10_53oqlnqnxhbetdy4ljlkkvmoby/node_modules/nextra-theme-blog/dist/index.mjs
Did you mean to import next@12.2.3_biqbaboplfbrettd7655fr4n2y/node_modules/next/head.js?
    at new NodeError (node:internal/errors:372:5)
    at finalizeResolution (node:internal/modules/esm/resolve:437:11)
    at moduleResolve (node:internal/modules/esm/resolve:1009:10)
    at defaultResolve (node:internal/modules/esm/resolve:1218:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:580:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:294:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_MODULE_NOT_FOUND',
  page: '/'
}
```

### Packages

```json
"next": "^12.2.3",
"nextra": "2.0.0-beta.10",
"nextra-theme-blog": "2.0.0-beta.10",
"react": "^18.2.0",
"react-dom": "^18.2.0"
```

PabloLION:
I tested with `v2.0.0-beta.10` and `nextra-theme-docs` in my test repo [PabloLION/zzz-test-nextra-356-src-test](https://github.com/PabloLION/zzz-test-nextra-356-src-test) and it worked fine.

PabloLION:
With this commit. https://github.com/PabloLION/zzz-test-nextra-356-src-test/commit/0972fe3baa419ba528237f4c845a56fb89e32511

PabloLION:
[...]

PR #615, opened 2022-07-30
chore: use lower build target

promer94:
* chore: move code highlighting theme to nextra
* chore: changeset
* fix: make cusdis a component #597


shuding (approved):
Thanks!

PR #524, opened 2022-07-09
docs: update configuration part

promer94:

PR #624, opened 2022-08-01
Fix flexsearch option being overridden

### LINKED ISSUE #594: Search broken when deploying to Vercel
I don’t know if it’s Nextra, Vercel, or my issue. But somehow, search is built and works locally but not when deployed. I’m using [nextra@2.0.0-beta.10](https://github.com/shuding/nextra/releases/tag/nextra%402.0.0-beta.10) and newest nextra-theme-docs, but with modified fonts/colors.

Any hints on what could be wrong?

580BEFEB-FE10-41F2-9F27-E5E809BFAFAF
51E0032A-B359-44FA-A069-EBB650006F95

shuding:
Closes #594. I noticed that the flexsearch option being overridden here:

CleanShot 2022-08-01 at 14 37 33@2x

Which is not needed, we should just remove it.

### DIFF
--- a/packages/nextra-theme-docs/src/index.tsx
+++ b/packages/nextra-theme-docs/src/index.tsx
@@ -282,8 +282,7 @@ const Content = ({
 const createLayout = (opts: PageOpts, config: DocsThemeConfig) => {
   const extendedConfig = {
     ...defaultConfig,
-    ...config,
-    unstable_flexsearch: opts.unstable_flexsearch
+    ...config
   }
   const nextThemes = extendedConfig.nextThemes || {}
   const Page = ({ children }: { children: ReactNode }): ReactNode => children
--- a/packages/nextra-theme-docs/src/index.tsx
+++ b/packages/nextra-theme-docs/src/index.tsx
@@ -282,7 +282,8 @@ const Content = ({
 const createLayout = (opts: PageOpts, config: DocsThemeConfig) => {
   const extendedConfig = {
     ...defaultConfig,
-    ...config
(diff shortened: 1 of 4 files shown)


packages/nextra-theme-docs/src/index.tsx
```diff
@@ -282,8 +282,7 @@ const Content = ({
 const createLayout = (opts: PageOpts, config: DocsThemeConfig) => {
   const extendedConfig = {
     ...defaultConfig,
-    ...config,
-    unstable_flexsearch: opts.unstable_flexsearch
+    ...config
```
B2o5T, line 285: we no longer provide `unstable_flexsearch` in theme config, but only in `withNextra`

B2o5T: more info about this change https://github.com/shuding/nextra/pull/524#discussion_r927798692

B2o5T (changes requested):
it's not overridden:
you have `unstable_flexsearch: false` in `nextraConfig`
you have `unstable_flexsearch: true` in theme, but we no longer take care of this value since we take value only from `nextraConfig`

shuding:
Makes sense, I think it should be enabled automatically based on `nextraConfig`

shuding:
OK I know what's the root cause now, under the condition `IS_PRODUCTION && indexContentEmitted.has(filename)` we are changing `unstable_flexsearch` to false to bypass re-indexing, but it affects the layout too.

B2o5T:
Nice catch!

PR #628, opened 2022-08-02
fix: attach mdxtheme to raw layout

### LINKED ISSUE #620: Convert markdown link to next/link
Links like `[](/about)` should be converted to `next/link` if the href is a relative path.

promer94:
close #620

### DIFF
--- a/.changeset/spicy-geckos-lay.md
+++ b/.changeset/spicy-geckos-lay.md
@@ -1,0 +1,5 @@
+---
+"nextra-theme-docs": patch
+---
+
+fix: attach mdxtheme to raw layout
--- a/packages/nextra-theme-docs/src/index.tsx
+++ b/packages/nextra-theme-docs/src/index.tsx
@@ -120,7 +120,7 @@ const Body = ({
         </article>
       ) : themeContext.layout === 'raw' ? (
         <div className="nextra-body full expand relative overflow-x-hidden">
-          {children}
+          <MDXTheme>{children}</MDXTheme>
         </div>
       ) : (
         <article


B2o5T (changes requested):
The concept of `raw` layout is to have unstyled elements, for this reason in raw layout you should use only `a` elements that replace internal links by NextLink

B2o5T:
@promer94 I guess you can’t hide navbar even with layout=raw, but sidebar can be hidden with `meta.json#theme.sidebar`

shuding:
Yeah the intention of "raw" was to not introduce any special styles to that page (so people can build a full-width landing page, for example). But for links and images (`[]()`, ``) we can still convert them to `<NextLink>` and `<NextImage>` because that doesn't add any special style, and it improves the experience. (Not sure if users can opt-out it by using `<a>` or ``, maybe we should try)

promer94:
> But links and images we can still convert them to <NextLink> and <NextImage> because that doesn't add any special style

This pr is intended to enable this feature.


> Not sure if users can opt-out it by using `<a>` or ``, maybe we should try

It will work.

Only markdown syntax will be converted to mdx component and could be replaced the mdx theme components.

plain html will be converted to normal jsx element.


Screen Shot 2022-08-02 at 20 57 38

promer94:
> Yeah the intention of "raw" was to not introduce any special styles to that page (so people can build a full-width landing page, for example).

I would like to clarify my understanding about `raw` layout.

I think docs theme actually provides two kinds of feature.

* Layout - nav, sidebar, toc, search ...
* Built-in components - Nextlink NextImage H1 H2 Tabs 

if user want to opt-out both of them. They could just use normal `.jsx` or `.tsx` page.

`raw` sounds more like a fast way to opt-out the layout feature but keep the built-in components to me.

shuding:
Just thought again about it, yeah it makes sense as long as only markdown syntax like `##` will be converted to the built-in `<H2>` and it doesn't affect `<h2>`, I'm good with this 👍

shuding:
Because usually, people can just do this https://github.com/vercel/turborepo/blob/main/docs/pages/index.mdx to avoid using any built-in component but having the navbar and footer at the same time.

B2o5T:
> Just thought again about it, yeah it makes sense as long as only markdown syntax like `##` will be converted to the built-in `<H2>` and it doesn't affect `<h2>`, I'm good with this 👍

In this case unneeded observers will be observe h2
https://github.com/shuding/nextra/blob/96ed5c2056da5651651bdc93e2848fafd7203ff4/packages/nextra-theme-docs/src/misc/theme.tsx#L104

shuding:
I think it will only observe the `H2` (which is created via `##` and `createHeaderLink`), not the user's JSX `h2` ?

B2o5T:
h2 via `##` - will observe
h2 via `<h2 />`  in mdx - will observe
Saw some helpful comment: https://github.com/shuding/nextra/blob/6644bd5dc476db4877b9d87e50ee5562ffa5fa76/examples/swr-site/pages/blog.en-US.mdx?plain=1#L11

Solution:
https://github.com/shuding/nextra/blob/96ed5c2056da5651651bdc93e2848fafd7203ff4/packages/nextra-theme-docs/src/misc/theme.tsx#L240
should be:

```tsx
<MDXProvider components={layout === 'raw' ? { a: A } : getComponents() as any}>
   {children}
</MDXProvider>
```

promer94:
> h2 via `##` - will observe 
h2 via `<h2 />` in mdx - will observe

It's actually more complicated 

```md
// Normal jsx element
export const Hello = () => <a href="./hello">hello</a>

// Normal jsx element
<a href="./hello">hello</a>

// Normal jsx element
<Hello />

// Mdx component
[hello](./hello)

// Mdx component
{<a href="./hello">hello</a>}
```

```jsx
/*@jsxRuntime automatic @jsxImportSource react*/
const {Fragment: _Fragment, jsx: _jsx, jsxs: _jsxs} = arguments[0];
const Hello = () => _jsx("a", {
  href: "./hello",
  children: "hello"
});
function _createMdxContent(props) {
  const _components = Object.assign({
    p: "p",
    a: "a"
  }, props.components);
  return _jsxs(_Fragment, {
    children: [_jsx("a", {
      href: "./hello",
      children: "hello"
    }), "\n", _jsx(Hello, {}), "\n", _jsx(_components.p, {
      children: _jsx(_components.a, {
        href: "./hello",
        children: "hello"
      })
    }), "\n", _jsx(_components.a, {
      href: "./hello",
      children: "hello"
    })]
  });
}
function MDXContent(props = {}) {
  const {wrapper: MDXLayout} = props.components || ({});
  return MDXLayout ? _jsx(MDXLayout, Object.assign({}, props, {
    children: _jsx(_createMdxContent, props)
  })) : _createMdxContent(props);
}
return {
  Hello,
  default: MDXContent
};
```

B2o5T:
was fixed here https://github.com/shuding/nextra/pull/674

PR #631, opened 2022-08-02
New layout implementation

### LINKED ISSUE #623: Inject the CSS and layout to _app automatically
Currently, one required step to get started with Nextra is to add a style import in _app, and add the custom layout handling.

It would be great if there’s a way to directly inject both via the webpack plugin & loader.

That said, this might be very tricky so opening this issue for discussion.

shuding:
Closes #623. 

The outcome of this PR:
- No need to import the CSS file in _app
- No need to implement the getLayout thing in _app
- No need to have a custom _app file
- Non-MDX pages (pages/hello.js) will not be affected by the global theme styles


packages/nextra/src/loader.ts
```diff
@@ -157,34 +166,73 @@ async function loader(
...
+  if (typeof cachedGlobalThemeStyle === 'undefined') {
+    const themeIncludeStyles = OFFICIAL_THEMES.includes(theme)
+    if (themeIncludeStyles) {
+      try {
+        const globalStylePath = await promisify(context.resolve)(
+          context.rootContext,
+          theme + '/style.css'
+        )
+        if (globalStylePath) {
+          cachedGlobalThemeStyle = readFileSync(globalStylePath).toString()
```
B2o5T, line 178:
```suggestion
          cachedGlobalThemeStyle = readFileSync(globalStylePath, 'utf8')
```

packages/nextra/src/loader.ts
```diff
@@ -157,34 +166,73 @@ async function loader(
     unstable_flexsearch
   }
 
+  if (typeof cachedGlobalThemeStyle === 'undefined') {
```
B2o5T, line 169:
```suggestion
  if (cachedGlobalThemeStyle === undefined) {
```

packages/nextra/src/loader.ts
```diff
@@ -3,6 +3,8 @@ import type { LoaderOptions, PageOpts } from './types'
 import path from 'path'
 import grayMatter from 'gray-matter'
 import slash from 'slash'
+import { promisify } from 'util'
+import { readFileSync } from 'fs'
```
B2o5T, line 7: Do you need to use graceful-fs instead?

packages/nextra/src/loader.ts
```diff
@@ -157,34 +166,73 @@ async function loader(
...
+        )
+        if (globalStylePath) {
+          cachedGlobalThemeStyle = readFileSync(globalStylePath).toString()
+          cachedGlobalThemeStyle = `<style jsx global>{${JSON.stringify(
+            cachedGlobalThemeStyle
+          )}}</style>`
+        } else {
+          cachedGlobalThemeStyle = null
+        }
+      } catch (e) {
```
B2o5T, line 185:
```suggestion
      } catch {
```

packages/nextra/src/loader.ts
```diff
@@ -157,34 +166,73 @@ async function loader(
...
-  titleText: typeof titleText === 'string' ? titleText : undefined,
-  ...__nextra_pageOpts__
-}, ${layoutConfig ? '__nextra_layoutConfig__' : 'null'}))
-
-function NextraPage(props) {
-  return (
-    <NextraLayout {...props}>
+function Content(props) {
+  return <>
+    ${cachedGlobalThemeStyle || ''}
```
B2o5T, line 220: Since `cachedGlobalThemeStyle` Is global style you can put it inside `__nextra_SSGContext__.Provider` and remove unneeded fragment wrapper

B2o5T:
How about size comparison? Imagine that our bundled `style.css` has 50kb and we have 100 mdx pages. Does it means that each page will have 50kb more?

packages/nextra/src/loader.ts
```diff
@@ -157,34 +166,73 @@ async function loader(
...
+  if (cachedGlobalThemeStyle === undefined) {
+    const themeIncludeStyles = OFFICIAL_THEMES.includes(theme)
+    if (themeIncludeStyles) {
+      try {
+        const globalStylePath = await promisify(context.resolve)(
+          context.rootContext,
+          theme + '/style.css'
+        )
+        if (globalStylePath) {
+          cachedGlobalThemeStyle = readFileSync(globalStylePath, 'utf8')
```
B2o5T, line 178: You can use async `readFile` here

shuding:
> How about size comparison? Imagine that our bundled style.css has 50kb and we have 100 mdx pages. Does it means that each page will have 50kb more? And total bundle will be increased by 50kb*100=5mb? 🧐

Yeah this is my biggest concern of using styled-jsx too, another choice is to use a webpack loader to magically inject that global CSS import but then it will have this (existing today) issue:

> Non-MDX pages (pages/hello.js) will not be affected by the global theme styles

shuding:
It would be so good if Next.js supports page-level global CSS...

shuding:
> It would be so good if Next.js supports page-level global CSS import...

OK I managed to make it work 😄

packages/nextra/src/index.js
```diff
@@ -40,23 +40,44 @@ const nextra = (...config) =>
...
+            // Match pages (imports without an issuer).
+            test: MARKDOWN_EXTENSION_REGEX,
+            issuer: request => !request,
+            use: [
+              options.defaultLoaders.babel,
+              {
+                loader: 'nextra/loader',
+                options: {
+                  ...nextraLoaderOptions,
+                  __nextra_page_import__: true
```
B2o5T, line 75: There is no need use underscores in this variable.
Can be renamed to:
```suggestion
                  pageImport: true
```

packages/nextra/src/compile.ts
```diff
@@ -72,9 +72,12 @@ export async function compileMdx(
     ].filter(Boolean)
   })
   try {
-    const result = await compiler.process(source)
+    const result = String(await compiler.process(source))
+      .replace('export const titleText =', 'const titleText =')
+      .replace('export default MDXContent', '')
```
B2o5T, line 77: Missing trailing semi
```suggestion
      .replace('export default MDXContent;', '')
```

packages/nextra/src/loader.ts
```diff
@@ -159,34 +176,50 @@ async function loader(
...
-const NextraLayout = __nextra_withSSG__(__nextra_withLayout__({
-  titleText: typeof titleText === 'string' ? titleText : undefined,
-  ...__nextra_pageOpts__
-}, ${layoutConfig ? '__nextra_layoutConfig__' : 'null'}))
-
-function NextraPage(props) {
-  return (
-    <NextraLayout {...props}>
+function Content(props) {
+  return <>
```
B2o5T, line 207: Unneeded fragment wrapper

packages/nextra/src/loader.ts
```diff
@@ -110,6 +117,15 @@ async function loader(
     },
     resourcePath
   )
+
+  // Imported as a normal component, no need to add the layout.
+  if (!__nextra_page_import__) {
+    return `
+${themeIncludeStyles && `import '${theme}/style.css'`}
```
B2o5T, line 124: Because it’s a boolean and can be evaluated to `false`
```suggestion
${themeIncludeStyles ? `import '${theme}/style.css'` :  ''}
```

packages/nextra/src/loader.ts
```diff
@@ -159,34 +176,50 @@ async function loader(
...
+      .replace(/\/index$/, '')
+      // Remove the only `index` route
+      .replace(/^index$/, '')
+
   return `
-import { withSSG as __nextra_withSSG__ } from 'nextra/ssg'
+import { SSGContext as __nextra_SSGContext__ } from 'nextra/ssg'
+
 import __nextra_withLayout__ from '${layout}'
+${themeIncludeStyles && `import '${layout}/style.css'`}
```
B2o5T, line 194: Same here

shuding: Also updated this to keep it the same format:

```
${layoutConfig ? `import __nextra_layoutConfig__ from '${layoutConfig}'` : ''}
```

B2o5T (approved):

shuding:
Thank you for the review @B2o5T!

shuding:
I also have to deprecate the

```md
<Nextra.Tabs>
```

use case, and built-in components won't be exposed by default. It now requires you to import `import { Tabs, Tab } from 'nextra-docs-theme/Tabs'` first.

This is because we are not always providing these components. So it causes issues since they might not be available (e.g. importing MDX directly).

B2o5T:
@shuding you are welcome! Amazing work

PR #635, opened 2022-08-04
Small copy change

molebox:
Tiny copy change to the getting started


B2o5T:
It’s already clear that Next.js config is `next.config.js` file, see line 23

shuding:
I think this makes sense because we also use this:

> Create a `theme.config.js` file for the docs theme:

PR #638, opened 2022-08-05
fix: apply components to mdx

### LINKED ISSUE #637: Bug: mdxcontent can't access component context
Context has the correct value.
Screen Shot 2022-08-05 at 11 32 07

But mdxcontent does not
Screen Shot 2022-08-05 at 11 31 51

promer94:
close #637

### DIFF
--- a/packages/nextra/src/loader.ts
+++ b/packages/nextra/src/loader.ts
@@ -176,20 +176,8 @@ export default MDXContent`.trimStart()
     newNextLinkBehavior
   }
 
-  const pageNextRoute =
-    '/' +
-    path
-      .relative(pagesDir, resourcePath)
-      // Remove the `mdx?` extension
-      .replace(MARKDOWN_EXTENSION_REGEX, '')
-      // Remove the `*/index` suffix
-      .replace(/\/index$/, '')
-      // Remove the only `index` route
-      .replace(/^index$/, '')
-
   return `
 import { SSGContext as __nextra_SSGContext__ } from 'nextra/ssg'
-
 import __nextra_withLayout__ from '${layout}'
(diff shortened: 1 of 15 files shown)


packages/nextra/src/loader.ts
```diff
@@ -203,23 +191,22 @@ globalThis.__nextra_internal__ = {
...
 
-export default __nextra_withLayout__(
-  ${JSON.stringify(pageNextRoute)},
-  Content,
-  {
-    titleText: typeof titleText === 'string' ? titleText : undefined,
-    ...__nextra_pageOpts__
-  },
-  ${layoutConfig ? '__nextra_layoutConfig__' : 'null'},
-)`.trimStart()
```
shuding, line 222: Let me explain a bit on my previous change. _app.js renders the exported component from each page like `render(<Page/>)`. And because `function Page(props) {` is a different function on every different page, when a client navigation happens, React re-renders the `render(<Page2/>)`. Because `<Page/>` and `<Page2/>` have different types (`Page !== Page2`), React unmounts the `<Page/>` component and mounts the `<Page2/>`.

Because the layout stuff (navbar, sidebar, etc.) are all inside `<Page/>`, the states there will be reset too. That's why we had the `getLayout()` stuff.

And to avoid that, I refactored the `withLayout` HoC to always return the same component `Layout`, but use `route` to get the correct content to render inside. This way the layout will preserve the state and always be stable.

shuding: > the states there will be reset too

There are two states, one is the React state (useState) and another one is the DOM state (scroll positions):

CleanShot 2022-08-05 at 11 40 39

PR #649, opened 2022-08-07
[docs]: extract `<Banner />` from `<Navbar />` to `components/banner.tsx`

B2o5T:


packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -48,7 +48,8 @@ export const DEFAULT_THEME: DocsThemeConfig = {
     <span className="block select-none p-8 text-center text-sm text-gray-400">
       No results found.
     </span>
-  )
+  ),
+  bannerKey: 'nextra-banner',
```
B2o5T, line 52: since it's theme config value, it's better to declare it here rather than
https://github.com/shuding/nextra/blob/a007c646cd6feae62fd77285a446706935c8fc72/packages/nextra-theme-docs/src/navbar.tsx#L83

packages/nextra-theme-docs/src/index.tsx
```diff
@@ -212,6 +213,7 @@ const InnerLayout = ({
           'menu-active': menu
         })}
       >
+        <Banner />
         {themeContext.navbar ? (
```
B2o5T, line 217: I guess rendering `<Banner />` should be not depending on `themeContext.navbar` value

Issue #630, opened 2022-08-02
Make blog dark mode context more accessible to user | make dark mode sync to custom components in mdx


RavenXev:
I'm using [Geist UI](https://geist-ui.dev/en-us) library to build custom components to import into my Nextra blog. In the config, I have enabled the dark mode button.

However, it's very difficult to access and sync my component to the current theme of the nextra website because the theme context appears to be inaccessible. I looked into the nextra-theme-blog index.js to try to reverse engineer it, and still could not find a good way to sync my custom component the nextra's theme.

Things I have tried:
1. [D3Portillo's local storage](https://github.com/shuding/nextra/issues/479) event listeners and tips. Very useful.
This was the closest I got, linking up a ton of listeners to the toggle button with document.querySelector('[aria-label="Toggle Dark Mode"]') and doing my best to sync everything to the click event. 
However, this solution wasn't perfect as I could get the custom component to change colors, but not the background. Something bugged.

2. Installing and implementing another layer of next-themes locally (did not appear to work, maybe because the next-themes built into nextra overrides it). Could not access theme from useThemes, always showed up as undefined.
3. Using useState/useEffect on _app.js to implement my own theme state.  Does not work because nothing seems to re-render when pressing the dark mode toggle button.

```
 const [theme, setTheme] = useState('light')
 useEffect(() => {
    // let darkActive = !!document.querySelector('.dark')
    // let localStorageTheme = localStorage.getItem('theme')
    // if (!darkActive || localStorageTheme == 'light') {
    //   setTheme('light')
    // }
    // if (darkActive || localStorageTheme == 'dark') {
    //   setTheme('dark')
    // }
    // console.log('theme', theme)
    // console.log('darkactive? ', darkActive)
    // console.log('local storage', localStorageTheme)
  }, [theme])

return(     
 <GeistProvider themeType={theme}>
      <Component {...pageProps} />
</GeistProvider>
)
```
**Current Behavior**: When I refresh the page, I am able to sync the component to the current theme in local storage. So every time I refresh, everything works. But when I press the toggle button, everything on the page changes except my custom component.
![image](https://user-images.githubusercontent.com/64918172/182418055-b74e1bfe-530f-4a75-ba31-dac3f4f3ff10.png)
![image](https://user-images.githubusercontent.com/64918172/182418100-e9b6f37d-f208-4b86-aeef-5f505541c279.png)

(the links change color because of nextra's css styling)

[...]

PR #654, opened 2022-08-07
export `useConfig` and `useTheme`

B2o5T:
fixes https://github.com/shuding/nextra/issues/630

### DIFF
--- a/packages/nextra-theme-docs/src/index.tsx
+++ b/packages/nextra-theme-docs/src/index.tsx
@@ -334,5 +334,7 @@ export default function withLayout(
   return Layout
 }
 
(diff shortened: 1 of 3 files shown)


packages/nextra-theme-docs/src/index.tsx
```diff
@@ -334,5 +334,7 @@ export default function withLayout(
   return Layout
 }
 
+export { useConfig }
+export { useTheme } from 'next-themes'
```
promer94, line 338: should also export `useTheme` in blog theme

promer94 (changes requested):

Issue #657, opened 2022-08-07
Unhandled Runtime Error: No content found for <route>


EpicEmeraldPlayz:
I'm facing problem that pages in the subfolder having error with `Unhandled Runtime Error: No content found for <route>`. This problem is not affect pages that in the same level as `_app.js`.

Environment:
---
```
OS: Windows 11
Node: v16.16.0
Package Manager: pnpm 7.9.0
```

Dependencies:
---
```
"next": "^12.2.4",
"nextra": "2.0.0-beta.15",
"nextra-theme-docs": "2.0.0-beta.15",
"react": "^18.1.0",
"react-dom": "^18.1.0"
```

# Example for Route `/a/test`

Folder Structure
---
```
pages/
├─ a/
│  ├─ test.mdx
├─ _app.js
├─ index.mdx
```

Error
---
```console
Unhandled Runtime Error
Error: No content found for /a/test.

Call Stack
Layout
node_modules\.pnpm\nextra-theme-docs@2.0.0-beta.15_vyccxgqxcnj5vudpssiumaxlme\node_modules\nextra-theme-docs\dist\index.js (2367:0)
renderWithHooks
node_modules\.pnpm\react-dom@18.2.0_react@18.2.0\node_modules\react-dom\cjs\react-dom.development.js (16305:0)
updateFunctionComponent
node_modules\.pnpm\react-dom@18.2.0_react@18.2.0\node_modules\react-dom\cjs\react-dom.development.js (19588:0)
beginWork
node_modules\.pnpm\react-dom@18.2.0_react@18.2.0\node_modules\react-dom\cjs\react-dom.development.js (21601:0)
HTMLUnknownElement.callCallback
node_modules\.pnpm\react-dom@18.2.0_react@18.2.0\node_modules\react-dom\cjs\react-dom.development.js (4164:0)
Object.invokeGuardedCallbackDev
node_modules\.pnpm\react-dom@18.2.0_react@18.2.0\node_modules\react-dom\cjs\react-dom.development.js (4213:0)
invokeGuardedCallback
node_modules\.pnpm\react-dom@18.2.0_react@18.2.0\node_modules\react-dom\cjs\react-dom.development.js (4277:0)
beginWork$1
node_modules\.pnpm\react-dom@18.2.0_react@18.2.0\node_modules\react-dom\cjs\react-dom.development.js (27451:0)
performUnitOfWork
[...]

PR #663, opened 2022-08-09
fix `Unhandled Runtime Error: No content found for <route>`

B2o5T:
fixes https://github.com/shuding/nextra/issues/657

### DIFF
--- a/packages/nextra/src/loader.ts
+++ b/packages/nextra/src/loader.ts
@@ -169,7 +169,7 @@ export default MDXContent`.trimStart()
     ? `import __nextra_themeConfig__ from '${slash(path.resolve(themeConfig))}'`
     : ''
   const pageOpts: Omit<PageOpts, 'title'> = {
-    filename: slash(filename),
+    filename,
     route: slash(route),
     meta,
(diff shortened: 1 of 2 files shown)


packages/nextra/src/loader.ts
```diff
@@ -165,7 +165,7 @@ export default MDXContent`.trimStart()
 
   const layoutConfig = themeConfig ? slash(path.resolve(themeConfig)) : ''
   const pageOpts: Omit<PageOpts, 'titleText'> = {
-    filename: slash(filename),
+    filename,
```
B2o5T, line 168: don't need since filename can't contain slashes

packages/nextra/src/loader.ts
```diff
@@ -182,8 +182,7 @@ export default MDXContent`.trimStart()
 
   const pageNextRoute =
     '/' +
-    path
-      .relative(pagesDir, resourcePath)
+    slash(path.relative(pagesDir, resourcePath))
```
shuding, line 185: Good catch

shuding (approved):

PR #677, opened 2022-08-10
Add `nextra-body-typesetting-article` back

shuding:


packages/nextra-theme-docs/src/styles.css
```diff
@@ -500,6 +500,38 @@ article {
       @apply rotate-90;
     }
   }
+
+  /* Typesettings */
+  &.nextra-body-typesetting-article {
```
B2o5T, line 505: I guess it's me who removed it because I didn't found class usage due interpolation

B2o5T: made rollback in my PR with following change
```ts
      className={cn(
        'nextra-body relative flex w-full min-w-0 max-w-full justify-center pb-8 pr-[calc(env(safe-area-inset-right)-1.5rem)]',
        {
          default: '',
          article: 'nextra-body-typesetting-article'
        }[themeContext.typesetting]
      )}
```

B2o5T (approved):

packages/nextra-theme-docs/src/styles.css
shuding: Yeah I added that "typesetting" option just for long articles, as they kinda need a little different design than docs.

Issue #683, opened 2022-08-11
[theme-blog] Estimated Reading Time


numToStr:
### Problem

When visiting a blog post page, user doesn't get any clue about how long and how much time it will take to read the post.

### Solution

It would be nice if we can show estimated reading time like `10 min read` for blog post.

### Screenshot

![image](https://user-images.githubusercontent.com/24727447/184131363-23d4b3c7-a8e3-4ea6-876a-fda1ae73b69d.png)

Source: https://leerob.io/blog/rust

B2o5T:
Helpful proposal!

PR #684, opened 2022-08-11
add `nextraConfig.unstable_readingTime` option for blog theme

B2o5T:
fixes https://github.com/shuding/nextra/issues/683

## disabled
image
## enabled
image
## another example
image

### DIFF
--- a/.changeset/purple-melons-begin.md
+++ b/.changeset/purple-melons-begin.md
@@ -1,0 +1,6 @@
+---
+'nextra': patch
+'nextra-theme-blog': patch
+---
+
+add `nextraConfig.unstable_readingTime` option for blog theme
--- a/examples/blog/next.config.mjs
+++ b/examples/blog/next.config.mjs
@@ -4,7 +4,8 @@ const withNextra = nextra({
   theme: 'nextra-theme-blog',
   themeConfig: './theme.config.jsx',
   unstable_staticImage: true,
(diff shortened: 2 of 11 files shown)


numToStr:
@B2o5T Excellent work 🎉. May I ask you Is there is reason for putting the tags on next line? Isn't it possible to just inline the reading time between date and tags?

B2o5T:
didn't want to put too many things on one line, let's wait for the thought of @shuding

shuding:
Great feature, thanks @B2o5T!

When there're many tags it looks cleaner to break into a new line indeed. BTW can we make them top-aligned?

CleanShot 2022-08-11 at 23 21 01

numToStr:
Yes, I also agree. Moving tags on the next line only when there are many tags and aligning back-button and theme-switch with date will also work.

B2o5T:
@shuding rebased and fixed vertical alignment, what about adding the same in docs theme? which place will be the best to put it there?

shuding:
Not sure if we want to put this on the docs theme (maybe as an option that defaults to false) because usually when people look at docs, they only check the section that they need (hence https://tailwindcss.com uses the search bar as CTA), unlike blogs.

shuding:
We can ship this first and have a discussion about that feature in docs!

shuding (approved):

PR #686, opened 2022-08-12
simpify `tocEl` ternary

B2o5T:
The current ternary is hard to read
image

1. If we remove `hideToc` variable, and move his value inside ternary we'll see duplicated  `themeContext.layout === 'raw'` check

image

2. We can safely remove the first one check
image


shuding:
Thanks, good refactor!

PR #688, opened 2022-08-12
[docs] simplify `<Body />`

B2o5T:


packages/nextra-theme-docs/src/index.tsx
```diff
@@ -60,22 +60,19 @@ function useDirectoryInfo(pageMap: PageMapItem[]) {
 
 interface BodyProps {
   themeContext: PageTheme
-  breadcrumb?: ReactNode
-  toc?: ReactNode
```
B2o5T, line 64: was never passed

packages/nextra-theme-docs/src/index.tsx
```diff
@@ -60,22 +60,19 @@ function useDirectoryInfo(pageMap: PageMapItem[]) {
...
 const Body = ({
   themeContext,
   breadcrumb,
-  navLinks,
   timestamp,
+  navigation,
   children
 }: BodyProps): ReactElement => {
-  const config = useConfig()
-  const { locale = DEFAULT_LOCALE } = useRouter()
```
B2o5T, line 78: can be removed as we'll use `useConfig` and `useRouter` in `<InnerLayout />`

PR #697, opened 2022-08-14
match-sorter search should highlight every match like flexsearch, update `match-sorter`

B2o5T:
`<MemoizedStringWithMatchHighlights />` was extracted in `components` directory as `<HighlightMatches />` to be reused in `<MatchSorterSearch />`.

## before 

image

## after

image

### DIFF
--- a/packages/nextra-theme-docs/src/components/flexsearch.tsx
+++ b/packages/nextra-theme-docs/src/components/flexsearch.tsx
@@ -1,38 +1,11 @@
-import React, { memo, useState, useEffect, Fragment, ReactElement } from 'react'
+import React, { useState, useEffect, ReactElement, ReactNode } from 'react'
 import { useRouter } from 'next/router'
 import FlexSearch from 'flexsearch'
 import cn from 'clsx'
 import { Search } from './search'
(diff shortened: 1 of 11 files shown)


packages/nextra-theme-docs/src/components/flexsearch.tsx
```diff
@@ -165,92 +145,93 @@ export function Flexsearch() {
   }, [search])
 
   const load = async () => {
-    if (!indexes[locale] && !loading) {
```
B2o5T, line 168: below only earlier return was modified https://github.com/shuding/nextra/blob/2e892345fe27f0895c703f026c1d913e1d686a13/packages/nextra-theme-docs/src/components/flexsearch.tsx#L148-L150

packages/nextra-theme-docs/src/env.d.ts
```diff
@@ -1,5 +1,3 @@
-declare module 'match-sorter'
```
B2o5T, line 1: in the newer version match-sorter provides own type definitions

shuding:
Thank you this looks good to me!

Issue #634, opened 2022-08-04
Search input should have X icon to clear value


B2o5T:
![B15B1145-3D58-4531-ACBD-9A3A99AADAE1](https://user-images.githubusercontent.com/7361780/182917583-f120dac7-0e7e-44a5-8c81-7d425beb8612.jpeg)

PR #698, opened 2022-08-14
add clear button for search input

B2o5T:
https://github.com/shuding/nextra/issues/634

I guess `ESC` text is better than X icon and it will be not confusing to see triple X if there is banner on page
image

## ❗️❗️❗️ Also I changed to show `⌘ K` for mac os

### DIFF
--- a/packages/nextra-theme-docs/src/components/search.tsx
+++ b/packages/nextra-theme-docs/src/components/search.tsx
@@ -148,7 +148,7 @@ export function Search({
             ? 'cursor-pointer hover:opacity-70 z-20 flex'
             : 'hidden sm:flex pointer-events-none'
         )}
-        title="Clear"
+        title={value ? 'Clear' : undefined}
         onClick={() => {
           onChange('')
         }}
--- a/packages/nextra-theme-docs/src/components/search.tsx
+++ b/packages/nextra-theme-docs/src/components/search.tsx
@@ -12,7 +12,7 @@ import { Transition } from '@headlessui/react'
 import { SpinnerIcon } from 'nextra/icons'
 import { Input } from './input'
 import { Anchor } from './anchor'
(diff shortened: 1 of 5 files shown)


packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -117,17 +117,56 @@ export function Search({
...
+          'absolute ltr:right-1.5 rtl:left-1.5 my-1.5 select-none',
+          'rounded bg-white px-1.5 h-5 font-mono font-medium text-gray-500 text-[10px]',
+          'border dark:bg-dark/50 dark:border-gray-100/20',
+          'contrast-more:border-current contrast-more:text-current contrast-more:dark:border-current',
+          'items-center gap-1 transition-opacity',
+          value
+            ? 'cursor-pointer hover:opacity-70 z-20 flex'
+            : 'hidden sm:flex pointer-events-none'
+        )}
+        title="Clear"
```
B2o5T, line 151:
```suggestion
        title={value ? 'Clear' : undefined}
```

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -117,17 +117,56 @@ export function Search({
...
+      >
+        {value
+          ? 'ESC'
+          : mounted &&
+            (navigator.userAgent.includes('Macintosh') ? (
+              <>
+                <span className="text-xs">⌘</span>K
+              </>
+            ) : (
+              '/'
```
shuding, line 164: I think we can do CTRL+K for non-Mac browsers

B2o5T, line 164:
```suggestion
              'CTRL K'
```

B2o5T: image

PR #702, opened 2022-08-15
add `DocsThemeConfig.bodyExtraContent` option

B2o5T:
we need it to add Giscus comments in our websites


shuding:
Makes total sense

PR #705, opened 2022-08-17
add new `unstable_mermaid` loader option for replacing '''mermaid  code-blocks with `<Mermaid />` (nextra/mermaid) React component

B2o5T:
In The Guild, we use a lot of rendering mermaid diagrams, but unfortunately, there is no plugin for MDX2 for replacing \`\`\`mermaid code blocks so I wrote my own remark plugin and want to share with the community

https://nextra-theme-docs-dl1oz9i5p.vercel.sh/docs/advanced/mermaid


packages/nextra/package.json
```diff
@@ -86,9 +90,15 @@
     "slash": "^3.0.0"
   },
   "peerDependencies": {
+    "mermaid": ">=9.1.5",
     "next": ">=9.5.3",
     "react": ">=16.13.1",
-    "react-dom": ">=18.2.0"
+    "react-dom": ">=16.13.1"
```
B2o5T, line 96: mistake, peerDeps `react-dom` must be same as `react`!

packages/nextra/package.json
```diff
@@ -86,9 +90,15 @@
...
+  },
+  "peerDependenciesMeta": {
+    "mermaid": {
+      "optional": true
+    }
```
B2o5T, line 101: making this package optional so users that don't use mermaid will not download this package

packages/nextra/src/mdx-plugins/static-image.js
```diff
@@ -1,57 +1,28 @@
...
+export const remarkStaticImage = () => (tree, _file, done) => {
     const importsToInject = []
 
     visit(tree, 'image', visitor)
-    tree.children.unshift(...importsToInject)
-    tree.children.unshift(ASTNodeImport('$NextImageNextra', 'next/image'))
+    tree.children.unshift(
+      getASTNodeImport('$NextImageNextra', 'next/image'),
+      ...importsToInject
+    )
```
B2o5T, line 21: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1865

packages/nextra/src/mermaid.tsx
```diff
@@ -0,0 +1,30 @@
...
+let id = 0
+
+export default function Mermaid({ chart }: { chart: string }): ReactElement {
+  const [svg, setSVG] = useState('')
+
+  useEffect(() => {
+    mermaid.initialize({
+      startOnLoad: true,
+      theme:
+        typeof localStorage === 'undefined' ? undefined : localStorage.theme
```
B2o5T, line 16: any through on how to access `useTheme` hook inside this component and use `theme` value here?

shuding:
I think it will be definitely helpful for the community to publish it as a standalone plugin btw :)

packages/nextra/src/mermaid.tsx
shuding: I think you can directly use `useTheme` in this component

Issue #704, opened 2022-08-16
`useConfig` returns an empty structure in `theme.config.tsx` `head()`


robbear:
I find that adding gray matter title/description/image to a page does not result in a populated `config` object as accessed from `useConfig` in `theme.config.tsx`. 

Below, the defaults are always in effect. This seems to have regressed since the `theme.config.tsx` migrated from `theme.config.js` and the recent new accessibility of `useConfig` from the docs site.
```
  head() {
    const config = useConfig();
    const description =
      config.meta.description ||
      "SWR is a React Hooks library for data fetching. SWR first returns the data from cache (stale), then sends the fetch request (revalidate), and finally comes with the up-to-date data again.";
    const image =
      config.meta.image ||
      "https://assets.vercel.com/image/upload/v1572282926/swr/twitter-card.jpg";
```

robbear:
In case it's useful to note, in exploring I did find that `config` is properly populated in `nextra-theme-docs/src/index.tsx` in `Body` and `InnerLayout`:
https://github.com/shuding/nextra/blob/a8c588385f7a968125f5b9831f209292336b7674/packages/nextra-theme-docs/src/index.tsx#L77
https://github.com/shuding/nextra/blob/a8c588385f7a968125f5b9831f209292336b7674/packages/nextra-theme-docs/src/index.tsx#L158

But in both `examples/swr-site` and in my own site, the `config` object is unpopulated at the time you'd like to use the `config.meta.description` etc value in the theme `head` function:
https://github.com/shuding/nextra/blob/a8c588385f7a968125f5b9831f209292336b7674/examples/swr-site/theme.config.tsx#L73

B2o5T:
will be fixed in next release

Issue #679, opened 2022-08-10
Can't use `useRouter` inside `head`


shuding:
For the docs theme, when you have this in the config:

```jsx
  head: function Head() {
    const router = useRouter()
    return (
      <>
        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
        ...
```

`router` is going to be `null` because of https://github.com/vercel/next.js/issues/23559#issuecomment-828080139.

B2o5T:
will be fixed in next release

PR #706, opened 2022-08-17
fix `useConfig`/`useRouter` inside `head()`

B2o5T:
fixes https://github.com/shuding/nextra/issues/704  https://github.com/shuding/nextra/issues/679

### DIFF
--- a/.changeset/sweet-worms-visit.md
+++ b/.changeset/sweet-worms-visit.md
@@ -2,4 +2,4 @@
 'nextra-theme-docs': patch
 ---
 
-fix `useConfig`/`useRouter` inside `head()`, should be wrapped with `NextHead />`
+fix `useConfig`/`useRouter` inside `head()`, should be wrapped with `<NextHead/>`
--- a/.changeset/sweet-worms-visit.md
+++ b/.changeset/sweet-worms-visit.md
@@ -2,4 +2,4 @@
 'nextra-theme-docs': patch
 ---
 
-fix `useConfig`/`useRouter` inside `head()`, should be wrapped with `<NextHead/>`
+fix `useConfig`/`useRouter` inside `head()`
--- a/.changeset/sweet-worms-visit.md
+++ b/.changeset/sweet-worms-visit.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+fix `useConfig`/`useRouter` inside `head()`, should be wrapped with `NextHead />`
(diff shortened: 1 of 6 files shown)


shuding:
The potential solution I was thinking for #679 is we don't render `<Head/>` but instead call it directly in the layout component:

```jsx
const head = Head()
return <NextHead>{head}</NextHead>
```

B2o5T:
@shuding i guess it’s will still not works, I remember that I tested it

B2o5T:
Yeah, do you mean inside `head`? If yes, this should works!

B2o5T:
@shuding finally fixed! so now hooks can be used as well inside `head()` but return should be wrapped with `<NextHead />`
image
image

.changeset/sweet-worms-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+fix `useConfig`/`useRouter` inside `head()`, should be wrapped with `NextHead />`
```
shuding, line 5:
```suggestion
fix `useConfig`/`useRouter` inside `head()`, should be wrapped with `<NextHead/>`
```

packages/nextra-theme-docs/src/components/head.tsx
```diff
@@ -12,38 +11,41 @@ export function Head(): ReactElement {
...
+            />
+          </>
+        )}
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1.0, viewport-fit=cover"
+        />
+      </NextHead>
+      {/* since in NextHead we can't use hooks */}
+      {renderComponent(config.head)}
```
shuding, line 48: My original idea was to do this:

```suggestion
        {config.head()}
      </NextHead>
```

So it's required to be a functional component.

B2o5T: nice, now it's works 💪. config and router no longer empty
image

.changeset/sweet-worms-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+fix `useConfig`/`useRouter` inside `head()`, should be wrapped with `<NextHead/>`
```
B2o5T, line 5:
```suggestion
fix `useConfig`/`useRouter` inside `head()`
```

packages/nextra-theme-docs/src/components/head.tsx
shuding: Thanks!!

PR #713, opened 2022-08-19
`"layout": "raw"` should have unstyled `<a />` and `<p />` elements as well

B2o5T:
otherwise my pagination's buttons with `layout: raw` have color and underscore 
image

### DIFF
--- a/packages/nextra-theme-docs/src/components/breadcrumb.tsx
+++ b/packages/nextra-theme-docs/src/components/breadcrumb.tsx
@@ -10,21 +10,19 @@ export function Breadcrumb({
   activePath: Item[]
 }): ReactElement {
   return (
-    <div className="nextra-breadcrumb mt-2.5 flex cursor-default overflow-hidden text-sm font-normal text-gray-500 transition-colors">
+    <div className="nextra-breadcrumb mt-2.5 flex gap-1 overflow-hidden text-sm text-gray-500 contrast-more:text-current">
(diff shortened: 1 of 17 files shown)


packages/nextra-theme-docs/src/styles.css
```diff
@@ -319,12 +309,6 @@ p {
   .nextra-navigation-links {
     border-color: #999 !important;
   }
-  .nextra-breadcrumb {
-    @apply text-current;
-    .active {
-      @apply font-bold !text-current;
-    }
-  }
```
B2o5T, line 327: moved inside `breadcrumb.tsx` with `contrast-more:*` modifier

packages/nextra-theme-docs/src/components/breadcrumb.tsx
```diff
@@ -33,9 +31,7 @@ export function Breadcrumb({
               title={item.title}
             >
               {isLink && !isActive ? (
-                <Anchor href={item.route} className="text-current no-underline">
-                  {item.title}
-                </Anchor>
+                <Anchor href={item.route}>{item.title}</Anchor>
```
B2o5T, line 34: we no longer need to apply `no-underline` classes for links since:
1. there is no more global `a` class with color and underline setup
2. `@import 'tailwindcss/base';` do it for us in `styles.css`

shuding:
Thanks!

PR #714, opened 2022-08-19
prefer `ref.current.querySelector` over `document.querySelector`, remove `load` prop in `<Search />`

B2o5T:


packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -234,7 +243,8 @@ export function Search({
                     <Anchor
                       className="block no-underline scroll-m-12 !text-current hover:!bg-transparent !p-0"
                       href={route}
-                      onMouseMove={() => setActive(i)}
+                      data-index={i}
+                      onMouseMove={handleMouseMove}
```
B2o5T, line 247: 1 function instead of many

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -183,13 +196,8 @@ export function Search({
...
         onKeyDown={handleKeyDown}
-        onFocus={() => {
-          load?.()
-          setShow(true)
-        }}
-        onBlur={() => {
-          setShow(false)
-        }}
+        onFocus={handleFocusAndBlur}
+        onBlur={handleFocusAndBlur}
```
B2o5T, line 200: 1 function instead of 2

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -21,7 +21,6 @@ type SearchProps = {
   className?: string
   value: string
   onChange: (newValue: string) => void
-  load?: () => Promise<void>
```
B2o5T, line 24: removed since useless, can be executed in `onChange`

Issue #662, opened 2022-08-08
in some cases `edit on GitHub` link broken


B2o5T:
For some reason `edit on GitHub` on this page is broken
https://graphql-code-generator-e9g5x3mjg-theguild.vercel.app/docs/getting-started

shuding:
Yeah `getting-started/` is missing from that link.

PR #720, opened 2022-08-20
fix edit on github button for cases when filename named as `index`

B2o5T:
fixes https://github.com/shuding/nextra/issues/662


packages/nextra/src/loader.ts
```diff
@@ -169,10 +169,11 @@ export default MDXContent`.trimStart()
   const themeConfigImport = themeConfig
     ? `import __nextra_themeConfig__ from '${slash(path.resolve(themeConfig))}'`
     : ''
+
   const pageOpts: Omit<PageOpts, 'title'> = {
-    filename,
+    filePath: path.relative(process.cwd(), resourcePath),
```
shuding, line 174: This fixes the path for Windows:

```suggestion
    filePath: path.relative(process.cwd(), resourcePath).replace(/\\/g, '/'),
```

B2o5T: Thanks, I think we can use `slash` package instead

PR #726, opened 2022-08-20
move `withLayout` logic directly in nextra loader

B2o5T:


packages/nextra/src/loader.ts
```diff
@@ -213,17 +217,18 @@ const Content = props => (
   </__nextra_SSGContext__.Provider>
 )
 
-export default __nextra_withLayout__(
-  ${JSON.stringify(pageNextRoute)},
+globalThis.__nextra_pageContext__ ||= Object.create(null)
```
B2o5T, line 220: actually thanks to `||=` it's a possible

PR #727, opened 2022-08-20
fix extra space in flexsearch input after loading indexes

B2o5T:
@shuding I don't know why you added it [here](https://github.com/shuding/nextra/commit/41e7b915b79b1cfa5ddc8a8e4742009a2d00a730) but this provoke the following issue

https://user-images.githubusercontent.com/7361780/185764644-b113e309-4466-4744-be0c-a44d72ee7bcd.mov

### DIFF
--- a/.changeset/polite-nails-mix.md
+++ b/.changeset/polite-nails-mix.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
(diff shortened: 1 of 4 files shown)


shuding:
Yeah I think it was a bad hack to make refocusing trigger the search again - we should definitely get rid of it!

PR #734, opened 2022-08-22
[v2] Docs

shuding:
As you can see there is now a v2 docs deployment:

CleanShot 2022-08-22 at 01 58 08@2x


B2o5T:
This landing page looks like so modern

Follow elements should be adjusted
B286F071-C67A-479B-A7A7-12BF471FD457
57A38CBF-5104-430A-9D89-9D0FC5B467A6

docs/pages/docs/guide/organize-files.mdx
```diff
@@ -1 +1,89 @@
...
+]
+```
+
+And the global page map will be bundled to each page by Nextra. Then, configured theme will render the actual UI with that page map.
+
+## \_meta.json
+
+It’s very common to customize each page’s title, rather than just using the filename. For example it doesn’t make sense to have a page titled “Index”, but instead have a page titled “Home”.
+
+That’s where `_meta.json` comes in. You can have a `_meta.json` file in each directory, and it will be used to override the default configuration of each page:
```
B2o5T, line 51:
```suggestion
That’s where `_meta.json` comes in. You can have an `_meta.json` file in each directory, and it will be used to override the default configuration of each page:
```

Should be readed as an underscore meta

jermanuts:
image
image
It was a bit confusing seeing https://nextra.vercel.app/themes/docs "This website is built with the docs theme." while in the repo it says https://github.com/shuding/nextra#themes (swr-site theme) which is different from nextra.vercel.app current theme.

Nextra right now have 3 themes.
Also, do you think v2 Docs is stable when using it like the same features in swr-site?

B2o5T:
> Good luck, it looks amazing. Wishing the best.

@jermanuts thank you! We almost done 🤘

shuding:
Thanks for all the reviews, I also did some changes to the core in this PR that is also noted in #826. The docs is still unfinished but since it's still a preview deployment, I think we can ship this first.

Issue #539, opened 2022-07-14
Copy to Clipboard action for codeblocks


numToStr:
### Problem

Currently, if someone wants to copy the codeblock contents they have to manually select the content and press <kbd>Ctrl-c</kbd>. This could be cumbersome to do if the codeblock is large.

### Solution

Add a `Copy To Clipboard` action; something similar to github's.

![image](https://user-images.githubusercontent.com/24727447/178981000-c7505b4a-b9dc-4d80-ac6e-18d5616de9fe.png)

shuding:
Good thing to have! We'll need a syntax for that, maybe:

```markdown
```js copy
```

It can be used together with other options:

```markdown
```js copy showLineNumbers
```

numToStr:
Ohh! I didn't thought about it earlier but making it opt-in would be cool. Also I agree with your syntax suggestion

```

```js copy

```

B2o5T:
Yeah, and we can have theme config option to turn this option by default to `true`

PR #735, opened 2022-08-22
[Blog/Docs] Add copy to clipboard button for code-blocks.

B2o5T:
- [x] Add `NextraConfig.unstable_defaultShowCopyCode` option to show button by default
- [x] add `copy` and `copy=false` options for code-blocks

fixes https://github.com/shuding/nextra/issues/539

maybe rename `unstable_defaultShowCopyCode` to `unstable_defaultShowCopyButton` ?

### DIFF
--- a/.changeset/happy-crabs-hope.md
+++ b/.changeset/happy-crabs-hope.md
@@ -1,0 +1,7 @@
+---
+'nextra': patch
+'nextra-theme-blog': patch
+'nextra-theme-docs': patch
+---
+
+[Blog/Docs] Add copy to clipboard button for code-blocks. Add `NextraConfig.unstable_defaultShowCopyCode` option to show button by default, add `copy` and `copy=false` options for code-blocks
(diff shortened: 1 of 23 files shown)


numToStr:
@B2o5T Wonderful! Thanks for working on this. 

How about `unstable_copyCode`? the `boolean` value already implies show/hide and default/not-default. So, it's better to just omit that. We can also document this in jsdoc.

B2o5T:
Since we have `defaultMenuCollapsed` option, I wanted also be more explicit in naming of property to not lost context of default showing/hiding option of copy button

numToStr:
that's why suggested to document the behavior in jsdoc instead.

B2o5T:
Yeah, also we need setup some generated documentation for nextra config and theme config options

shuding:
Great work!

numToStr:
@B2o5T Sorry for the noise but the latest beta release doesn't have this change. Is that a mistake? And It seems that `better-loader-types` (to which this PR got merged) is also missing.

B2o5T:
@numToStr this PR has conflicts that must be resolved https://github.com/shuding/nextra/pull/718

PR #750, opened 2022-08-28
break words in navlinks, remove unneeded `<div />` wrappers, align text on right side for next link

B2o5T:
## before

image

## after

image

### DIFF
--- a/packages/nextra-theme-docs/src/components/callout.tsx
+++ b/packages/nextra-theme-docs/src/components/callout.tsx
@@ -39,9 +39,7 @@ export function Callout({
       >
         {emoji}
       </div>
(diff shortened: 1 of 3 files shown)


packages/nextra-theme-docs/src/components/callout.tsx
```diff
@@ -39,9 +39,7 @@ export function Callout({
       >
         {emoji}
       </div>
-      <span className="w-[calc(100%-2.5rem)]">
-        {children}
-      </span>
+      <div className="min-w-0">{children}</div>
```
B2o5T, line 42: improvement of https://github.com/shuding/nextra/pull/746 to avoid dependence on padding size (`2.5rem`)

PR #752, opened 2022-08-28
reuse table styles from docs in blog

B2o5T:
Should we use styles from `@tailwindcss/typography` or from docs?

## before

image

## after

image


shuding:
Personally I prefer the one from docs (which was copied from GitHub). 😄

Issue #747, opened 2022-08-26
Feedbacks on nextra-theme-docs


nihgwu:
I've been watching this project for a long time, and right now I'm experimenting to migrate a Docusaurous based React UI Library to Nextra, I've made most of the parts work seamlessly as before, some feedbacks:

- [x] Make heading as normal heading and only add `a` to `#`, the problem is that we have link in heading and nesting link is invalid and will cause SSR error https://github.com/shuding/nextra/pull/897

- [x] Can we extract title from markdown (both H1 and frontmatter.title), so we don't need to config for each file in `meta.json`, we have hundreds of markdown files, and it would a nightmare to maintain `meta.json` manually https://github.com/shuding/nextra/pull/755
    > **Note**: was made only for `frontMatter.title`

- [x] We also maintain blogs, however the articles are in ASC order by date, seems we have to use `meta.json` to change the order, can we respect the `frontmatter.date` and list in descending order https://github.com/shuding/nextra/pull/894
- [ ] Style isolation. My website is for Design System, so it's crucial that the style should not be affected by the external resources.
    a. The whole app should add `isolation: isolate` so the header and sidebar want overlay the modals from user space
    b. Try to avoid using `preflight`, and try to not using atomic classnames, as it will clash with user space style if they are using tailwind in their components with another config
- [x] Not Nextra related probably, it's super slow when navigating to a new page, seems my local package is not cached and rebuild every time if it's not visited, I'm using `next-transpile-modules` to work with local package, for comparison, it's hardly to feel the compiling for new page with Docusaurous
  > **Note**: was moved to track here https://github.com/shuding/nextra/issues/753

Thanks for the great project 👍 

UPDATE:
- [x]  Update style or TOC to use the same custom scrollbar as side menu https://github.com/shuding/nextra/pull/756
<img width="273" alt="image" src="https://user-images.githubusercontent.com/2595058/187099455-771b7e03-f37f-4fae-88a0-75922cce8262.png">

- [x]  We should skip navigating by tab if the submenu is hidden, right now if we tab on side menu items, it goes deep to collapsed items where I can't see the focus ring at all https://github.com/shuding/nextra/pull/805

- [x] Can we use agnostic terms, like `github`, we don't use Github but other hosting services https://github.com/shuding/nextra/pull/734
    > **Note**: `github` was removed here, more info here https://github.com/shuding/nextra/issues/826#issue-1368578794

[...]

PR #755, opened 2022-08-29
define page title in sidebar from `frontMatter.title` if page is not specified in `_meta.json`

B2o5T:
fixes https://github.com/shuding/nextra/issues/747

use `item.frontMatter.title` either `item.name` as sidebar title if page is not specified in `_meta.json`


packages/nextra-theme-docs/src/utils/normalize-pages.ts
```diff
@@ -225,11 +222,12 @@ export function normalizePages({
         underCurrentDocsRoot: underCurrentDocsRoot || isCurrentDocsTree,
         pageThemeContext: extendedPageThemeContext
       })
+    const title = extendedMeta.title || (type !== 'separator' && a.name)
 
     const getItem = (): Item => ({
       ...a,
       type,
-      title: title || (type === 'separator' ? undefined : getTitle(a.name)),
+      ...(title && { title }),
```
B2o5T, line 230: I remove `title` package since we no need it here
Example: `getTitle` will convert page name `getting-started` to `Getting-Started`

packages/nextra/src/context.ts
```diff
@@ -24,12 +23,12 @@ function filter(
 } {
   let activeLevelPages: Page[] = []
   const items: Page[] = []
-  const meta =
-    pageMap.find((item): item is MetaJsonFile => item.name === META_FILENAME)
-      ?.meta || {}
+  const meta = pageMap.find(
+    (item): item is MetaJsonFile => item.kind === 'Meta'
+  )!.data
```
B2o5T, line 28: now `_meta.json` fill be always exist in `pageMap` and in `fileMap` even it's not exist on disk, this `_meta.json` will contain all pages with the title as `item.frontMatter.title` or `item.name`

packages/nextra/src/context.ts
```diff
@@ -24,12 +23,12 @@ function filter(
...
 
   for (const item of pageMap) {
-    if (item.name === META_FILENAME) continue
+    if (item.kind === 'Meta') continue
```
B2o5T, line 31: I explicitly added `kind` property, so TypeScript will be much smarter to understand of which kind of file is processed

packages/nextra-theme-blog/src/utils/parent.ts
```diff
@@ -1,10 +1,11 @@
 import { MdxFile } from 'nextra'
 import { LayoutProps } from '../types'
 import traverse from './traverse'
+import { Folder } from 'nextra/src/types'
```
B2o5T, line 4:
```suggestion
import { Folder } from 'nextra'
```

PR #763, opened 2022-08-30
fix covered theme switch popup when i18n is not setup

B2o5T:
image


shuding:
Didn't notice that you already got this fixed, thanks! Still keeping #783 because I think it can be full-width and show the theme name there in this case.

PR #766, opened 2022-08-30
mark `react-children-utilities` as external dependency in tsup config

B2o5T:
for some reason, it not was imported 😏
image


B2o5T:
this package was removed here https://github.com/shuding/nextra/pull/774

PR #774, opened 2022-08-31
fix missing `Copy Code` button in code-blocks without language

B2o5T:
- [x] add default language for code blocks `language-text` so they will be processed by `rehypePrettyCode`, and copy code button will be present for them
- [x] remove `react-children-utilities` dependecy

### DIFF
--- a/packages/nextra/styles/code-block.css
+++ b/packages/nextra/styles/code-block.css
@@ -28,11 +28,6 @@ pre {
   contain: paint;
   code {
     @apply grid min-w-full rounded-none border-none !bg-transparent !p-0 text-sm leading-5 text-current dark:!bg-transparent;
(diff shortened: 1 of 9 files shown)


packages/nextra/styles/code-block.css
```diff
@@ -28,11 +28,6 @@ pre {
   contain: paint;
   code {
     @apply grid min-w-full rounded-none border-none !bg-transparent !p-0 text-sm leading-5 text-current dark:!bg-transparent;
-    /* code blocks without languages don't have children that have px-4 */
-    &:not([data-language]) {
-      @apply !px-4;
-    }
-
```
B2o5T, line 35: now these styles is no longer needed

PR #786, opened 2022-09-01
Fix edit link not showing

shuding:
You can see from the dev deployment that the edit link is missing.


B2o5T:
@shuding fixed with better solution here https://github.com/shuding/nextra/pull/765

shuding:
Wow thank you @B2o5T!

PR #798, opened 2022-09-02
better focus ui, use ring color as theme hue color

B2o5T:
also fixes https://github.com/shuding/nextra/issues/203

https://user-images.githubusercontent.com/7361780/188211094-18375fa0-4aec-43bf-855e-3d59969f0495.mov

https://user-images.githubusercontent.com/7361780/188211313-3e8236e6-bf0a-437a-8ed7-a2aa83930c71.mov

moved tailwind classes directly in className's, also `styles.css` now contains only 134 lines and 1 month ago it was 700+
image

PR #803, opened 2022-09-04
fix covered select options

B2o5T:
there is no way without using react portal

more info https://github.com/tailwindlabs/headlessui/issues/59


shuding:
I think that popup is right-aligned currently, did we run into an issue with that?

B2o5T:
@shuding it’s right aligned but in the latest PR with hided sidebar without this PR popup (switch language or theme) will be covered since container width is too small

shuding:
Ah that makes sense!

shuding (commented):
Love the new style! Can you make sure the first item is always expanded?

CleanShot 2022-09-05 at 20 06 03@2x

B2o5T:
> Love the new style! Can you make sure the first item is always expanded?
> 
> CleanShot 2022-09-05 at 20 06 03@2x

Do we really need it? Because on mobile this will be extra expanded, more info in my comment 

https://github.com/shuding/nextra/pull/798#discussion_r961907290

If you still prefer extra expanded I’ll change it

shuding:
Got it! I think then we only make it extend on desktop?

packages/nextra-theme-docs/src/index.tsx
```diff
@@ -188,11 +177,21 @@ const InnerLayout = ({
...
+
   return (
-    <div
-      dir={isRTL ? 'rtl' : 'ltr'}
-      className={cn('nextra-container transition-opacity', menu && 'opacity-80')}
-    >
+    <>
+      <script
+        dangerouslySetInnerHTML={{
+          __html: `document.documentElement.setAttribute('dir','${direction}')`
```
shuding, line 192: Makes sense! Does it work when the language changes like from LTR to RTL?

B2o5T: yeah it works, also there is no shrinking on initial loading

B2o5T:
will convert it to draft until I apply your suggestion

PR #811, opened 2022-09-05
Style improvements

shuding:
- Make sure that commands like `dev:theme-docs` don't trigger the v2 docs package
- Improve iOS Safari style as `100dvh` is now out
- Fix links with empty `href`
- Preload search index file
- Fix search result clickable area
- Fix theme icon size
- Fix `dir=ltr` missing before hydration (layout jumps)


B2o5T:
@shuding for rtl jumps I have a better solution here https://github.com/shuding/nextra/pull/803/files#diff-8310a54b1ec9da38ed7c74e8c1328e33f07848d25bd9779811614642e3268747R180-R194

since now switch/theme popups will render in body via react portals, and they are not children for root div

packages/nextra-theme-docs/css/styles.css
```diff
@@ -246,7 +255,7 @@ input[type='search']::-webkit-search-results-decoration {
   display: none;
 }
 .locale-switch[open] summary:before {
-  content: ' ';
+  content: '';
```
B2o5T, line 258: this class seems not used anywhere, not?

packages/nextra-theme-docs/src/mdx-components.tsx
```diff
@@ -181,6 +181,7 @@ const Details = ({
     <details
       className="my-4 rounded border border-gray-200 bg-white p-2 shadow-sm dark:border-neutral-800 dark:bg-neutral-900 first:mt-0 last:mb-0"
       {...props}
+      {...(openState && { 'data-expanded': true })}
```
B2o5T, line 184: `data-expanded={openState}` should works also

packages/nextra-theme-docs/src/mdx-components.tsx
```diff
@@ -195,8 +196,8 @@ const Summary = (props: ComponentProps<'summary'>): ReactElement => {
     <summary
       className={cn(
         'list-none cursor-pointer rounded p-1 outline-none transition-colors hover:bg-gray-100 dark:hover:bg-neutral-800',
-        "before:content-[''] before:inline-block before:transition-transform dark:before:invert",
-        '[[open]>&]:before:rotate-90 rtl:before:rotate-180'
+        "before:mr-1 before:content-[''] before:inline-block before:transition-transform dark:before:invert",
+        '[[data-expanded]>&]:before:rotate-90 rtl:before:rotate-180'
```
B2o5T, line 200: does previous didn't work? 🧐

packages/nextra-theme-docs/css/styles.css
shuding: Oh right, that seems to be legacy

packages/nextra-theme-docs/src/mdx-components.tsx
shuding: This makes sure that the arrow rotation animation isn’t delayed like the collapse animation

B2o5T: Nice catch, remember that saw this issue :)

PR #812, opened 2022-09-05
Make sure arrows are aligned with nested folders

shuding:
CleanShot 2022-09-05 at 21 16 37@2x

### DIFF
--- a/packages/nextra-theme-docs/src/components/collapse.tsx
+++ b/packages/nextra-theme-docs/src/components/collapse.tsx
@@ -70,7 +70,7 @@ export function Collapse({
     >
       <div
         ref={innerRef}
-        className="p-2 transform-gpu overflow-hidden transition-opacity duration-500 ease-in-out motion-reduce:transition-none"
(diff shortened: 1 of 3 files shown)


packages/nextra-theme-docs/src/components/collapse.tsx
```diff
@@ -70,7 +70,7 @@ export function Collapse({
     >
       <div
         ref={innerRef}
-        className="p-2 transform-gpu overflow-hidden transition-opacity duration-500 ease-in-out motion-reduce:transition-none"
+        className="p-2 ltr:pr-0 rtl:pl-0 transform-gpu overflow-hidden transition-opacity duration-500 ease-in-out motion-reduce:transition-none"
```
B2o5T, line 73: it's better to apply this class only for sidebar folders but not by default, what do you think?

`<Collapse className="ltr:pr-0 rtl:pl-0" ...`

shuding: Makes sense since this is a general component!

Issue #799, opened 2022-09-04
Table styling is highly inconsistent


ChristianIvicevic:
I am currently using the Beta 24 and run into the following issues that are really confusing:

1. HTML tables are almost fully unstyled. Only the `th` elements appear in bold per default:
```html
<table>
  <thead>
  <tr>
    <th>A</th>
    <th>B</th>
  </tr>
  </thead>
  <tbody>
  <tr>
    <td>C</td>
    <td>D</td>
  </tr>
  <tr>
    <td>E</td>
    <td>F</td>
  </tr>
  </tbody>
</table>
```
![image](https://user-images.githubusercontent.com/45569997/188311242-23bb88e0-4d91-40df-ae74-782a4d99ef32.png)

2. However tables with bodies that are the result of dynamic generation (or just wrapped with curlies) are correctly styled while the header is still almost completely unstyled:
```jsx
<table>
  <thead>
  <tr>
    <th>A</th>
    <th>B</th>
  </tr>
  </thead>
  <tbody>
  {[...Array(2)].map((__, index) => <tr key={index}>
    <td>C</td>
    <td>D</td>
  </tr>)}
  {<tr><td>C</td><td>D</td></tr>}
  </tbody>
</table>
```
![image](https://user-images.githubusercontent.com/45569997/188311310-8d48012f-5e13-4c7a-9b4b-d03638e24233.png)

3. GFM tables are corrrectly styled in their entirety:
```md
| A | B |
|---|---|
|  C|  D|
|  C|  D|
|  C|  D|
```
![image](https://user-images.githubusercontent.com/45569997/188311352-69215b65-f144-4173-95d1-dab182912dc3.png)

I'd like to ask whether you already have a plan whether or not custom tables will be styled or not in the final release of v2. I wasn't personally looking forward to writing custom styles (especially since I didn't get Tailwind working correctly) and would've liked to get the components automatically styled by `nextra-theme-docs` per default.

B2o5T:
Hi! Does there are some way replace HTML `table` / `tr` / `th` / `th` tags with MDX components?
Because currently only GFM tables are replaced

ChristianIvicevic:
Creating a table via a React component embedded into the MDX has the same behavior of being unstyled. My use-cases are that I dynamically want to generate tables at build time based on external data which is why GFM isn't an option for anything that is not hard-coded.

B2o5T:
@ChristianIvicevic I don't know why this happens and why table is not replace with MDX components table,
[...]

Issue #741, opened 2022-08-24
[nextra-theme-docs] PR #674 removed table styling that serviced authored markdown


robbear:
PR #674 removed `table` styling from `src/styles.css`. My site depended on that for authored html tables in markdown. What was the intention in removing this, and what is the recommendation for sites previously depending on it?

B2o5T:
Do you use `layout: raw`?  Intention of removing these styles - `layout: raw` should show all elements unstyled to avoid conflicts with user’s table styles for example

> what is the recommendation for sites previously depending on it

Explicitly define your classes, if you want have table with same styles from markdown pages, copy classes from here https://github.com/shuding/nextra/blob/core/packages/nextra-theme-docs/src/mdx-components.tsx

robbear:
> Do you use `layout: raw`?

No, not on that page. The author is inlining HTML in an otherwise regularly themed page.

B2o5T:
Can you show code example of your table ?

robbear:
A snippet from the markdown:

```
# Keywords and Symbols

Tables of keywords and other lexical symbols.

### Operations on Relations
<table>
  <thead>
    <tr>
      <th>Symbol</th>
      <th>Operator Name</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><code>
        <a href="/rel/ref/lib/stdlib#union">∪</a>
      </code></td>
      <td><code>
        <a href="/rel/ref/lib/stdlib#union">union</a>
      </code></td>
    </tr>
    <tr>
      <td><code>
        <a href="/rel/ref/lib/stdlib#intersect">∩</a>
      </code></td>
      <td><code>
        <a href="/rel/ref/lib/stdlib#intersect">intersect</a>
      </code></td>
    </tr>
    <tr>
      <td><code>
        <a href="/rel/ref/lib/stdlib#proper_superset">⊃</a>
      </code></td>
      <td><code>
        <a href="/rel/ref/lib/stdlib#proper_superset">proper_superset</a>
      </code></td>
    </tr>
    <tr>
      <td><code>
        <a href="/rel/ref/lib/stdlib#proper_subset">⊂</a>
      </code></td>
      <td><code>
        <a href="/rel/ref/lib/stdlib#proper_subset">proper_subset</a>
      </code></td>
    </tr>
    <tr>
      <td><code>
        <a href="/rel/ref/lib/stdlib#superset">⊇</a>
      </code></td>
      <td><code>
[...]

PR #813, opened 2022-09-06
explain why table is rendered differently

B2o5T:
@shuding I tried to explain confusing behaviour of rendering tables

https://nextra-v2-ggyzxlq24-shuding1.vercel.app/docs/guide/advanced/table

fixes https://github.com/shuding/nextra/issues/799 and https://github.com/shuding/nextra/issues/741


shuding:
This is so helpful, thank you! I'll copy this to the v2 docs as well

PR #814, opened 2022-09-06
use `title` package in nextra to determine sidebar title based on pagename

B2o5T:
1. bring back `title` package but use it not in `nextra-theme-docs` but in `nextra` directly

So follow sidebar links:
image

will be replaced with:

image

2. replace all `-` and `_` characters in pagename to make it looks even better!

image


shuding:
> replace all - and _ characters in pagename to make it looks even better

this is amazing!

Issue #810, opened 2022-09-05
Smarter image optimization


shuding:
Nextra currently modifies relative Markdown image paths to `next/image` such as `![](../../public/image.jpg)`. Another smart move we can do is to also handle `![](/image.jpg)` when it starts with `/` because it _should_ be from the public folder. Something like `join(relative(__dirname, appRoot), 'public', absoluteUrl)` may work.

There are exceptions though such as rewrites, so we can first check if that image exists in the public folder and if not, we don't do any modification there.

B2o5T:
If image doesn’t exist we can show at least console warning, this will be more helpful

PR #819, opened 2022-09-07
replace images with `<NextImage />` even when url not relative but that starts from `/` (public directory)

B2o5T:
fixes https://github.com/shuding/nextra/issues/810

### DIFF
--- a/packages/nextra/src/content-dump.ts
+++ b/packages/nextra/src/content-dump.ts
@@ -1,34 +1,24 @@
+import path from 'node:path'
 import fs from 'graceful-fs'
-import path from 'path'
-import { FrontMatter } from './types'
-import { CWD } from './constants'
-
-const { statSync, mkdirSync } = fs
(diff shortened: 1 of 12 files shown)


packages/nextra/src/content-dump.ts
```diff
@@ -61,24 +51,23 @@ export function addPage({
...
   // entire build.
   const content = JSON.stringify(asset[locale])
-  fs.writeFileSync(path.join(assetDir, dataFilename), content)
-  fs.writeFileSync(path.join(cacheDir, dataFilename), content)
+  fs.writeFileSync(path.join(ASSET_DIR, dataFilename), content)
+  fs.writeFileSync(path.join(CACHE_DIR, dataFilename), content)
 }
 
 // Copy cached data to the asset directory.
-export async function restoreCache() {
```
B2o5T, line 69: was never async

PR #832, opened 2022-09-11
do not add `placeholder="blur"` on `.svg` static images

B2o5T:


examples/swr-site/theme.config.tsx
```diff
@@ -198,12 +199,15 @@ const config: DocsThemeConfig = {
...
-      </div>
-    ),
+    titleComponent: ({ title, type }) =>
+      type === "separator" ? (
+        <div className="flex items-center gap-2">
+          <Logo className="h-1.5 shrink-0" />
+          {title}
+        </div>
+      ) : (
+        <>{title}</>
```
B2o5T, line 209: fixes 
image

packages/nextra/src/mdx-plugins/static-image.ts
```diff
@@ -82,7 +82,7 @@ export const remarkStaticImage: Plugin<
             name: 'alt',
             value: node.alt
           },
-          {
+          !url.endsWith('.svg') && {
```
B2o5T, line 85: without this, will get an error 
image

shuding: Good catch, thanks!

PR #843, opened 2022-09-13
Fix Nextra not able to run on WebContainers

shuding:
This fixes the issue that native addons cannot be loaded on WebContainers currently.


shuding:
I'd like the docs to have Stackblitz examples embedded :^)

PR #856, opened 2022-09-19
make `nextra-theme-docs` ESM compatible

B2o5T:
currently it has ESM issues, we use a patch to make it works 😅

https://github.com/dotansimha/graphql-yoga/blob/48794374dfd4a0f9b0544ab092194055dbc7ca6d/patches/nextra-theme-docs%2B2.0.0-beta.29.patch


B2o5T:
@shuding needs your help, these changes fix ESM issues but in `examples/swr-site` in dev and prod there are some hydration problems

## dev
image

## prod
image

I tried a lot of things but nothing helps, but in an applied patch on our website (`graphql-yoga`) there are no one hydration problems, I guess it goes from `pnpm` and its `dependenciesMeta`

image

promer94:
Could you explain more about what is `ESM issues`.  It would be batter if we could know more about the original problem

Current changes seems to make it harder for user to create their own theme.  
`import head from 'next/head.js'` is really wired syntax.  normally you dont need to add file extension for `bare specifiers` https://nodejs.org/api/esm.html#mandatory-file-extensions

Also, does the blog-themes have the same `ESM issues` ?

B2o5T:
@promer94

both `nextra-theme-docs` and `nextra-theme-blog` are ESM packages but they misses `"type": "module"` in their `package.json`'s

if you add `"type": "module"` in swr example and `nextra-theme-docs`, your app will be no longer works due misses `.js` extension in `next/head`, `next/router` and `next/link`

image

promer94:
I have two problems.
 
Although `nextra-theme-docs` itself is esm, it depends on `next` which is not esm. should we add `"type": "modules"` in this case ?

`swr example` is a applicatioon, is it necessary to add `"type": "modules"` for applications ?

B2o5T:
> swr example is a applicatioon, is it necessary to add "type": "modules" for applications ?

It up to you specify or not, both are fine with/without

> Although nextra-theme-docs itself is esm, it depends on next which is not esm. should we add "type": "modules" in this case ?

The problem was on our side, our package with components reexport all things from `nextra-theme-docs` but we also specified `type: module` in our `package.json` after removing it all things works without errors

PR #857, opened 2022-09-19
Allows regeneration of `pnpm-lock.yaml` by fixing build error

### LINKED ISSUE #844: V2: `pnpm install` fails when `pnpm-lock.yaml` is not present
On a fresh clone of branch `a0e5847`, if I remove `pnpm-lock.yaml` so as to regenerate it and run `pnpm install`, the installation fails. 

```
[NEXTRA-ESM] DTS Build start
│ [NEXTRA-TYPES] DTS Build start
│ src/plugin.ts(173,43): error TS2345: Argument of type 'string | undefined' is no
│   Type 'undefined' is not assignable to type 'string'.
│ Error: error occured in dts build
```

The committed lock file allows a successful `pnpm install`, but apparently can't be regenerated. This blocks my project from creating its own variant of the lock file.

robbear:
Fixes #844 

The lockfile can't currently be regenerated since the resulting build breaks due to TypeScript problems. This is a proposed fix.

1. `rm pnpm-lock.yaml`
2. `pnpm install`

Note that regenerating the lockfile appears to reveal peer dependency warnings due to the use of React 18.

### DIFF
--- a/packages/nextra/src/loader.ts
+++ b/packages/nextra/src/loader.ts
@@ -19,7 +19,7 @@ import {
   CWD
 } from './constants'
 
-const PAGES_DIR = findPagesDir(CWD).pages
+const PAGES_DIR = findPagesDir(CWD).pages as string
 
 // TODO: create this as a webpack plugin.
 const indexContentEmitted = new Set<string>()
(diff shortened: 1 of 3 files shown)


shuding:
Thank you!

PR #859, opened 2022-09-20
found some missing passHrefs to a tags,  reduce some warnings at DOM levels

Shubham567:
Just a small commit for now, however i have been looking into another issue on the console, which says,

> repeated forward or backslash are not allowed in routes


B2o5T:
Thanks, can you add changeset also?

B2o5T (approved):

PR #860, opened 2022-09-25
fix: ctrl+k check with e.ctrlKey

louisgv:
On chrome in non-mac, metaKey is not mapped into ctrlKey. We must use ctrlKey explicitly to check


B2o5T:
thanks, added changeset, will be included in next release

PR #865, opened 2022-09-30
use resolvedTheme instead renderedTheme + theme check

B2o5T:
just because resolvedTheme will be `dark`/`white`


packages/nextra-theme-blog/src/theme-switch.tsx
```diff
@@ -4,9 +4,9 @@ import { MoonIcon, SunIcon } from 'nextra/icons'
 import { useMounted } from 'nextra/hooks'
 
 export default function ThemeSwitch() {
-  const { theme, setTheme, resolvedTheme } = useTheme()
+  const { setTheme, resolvedTheme } = useTheme()
   const mounted = useMounted()
-  const isDark = theme === 'dark' || resolvedTheme === 'dark'
```
B2o5T, line 9: because here theme could be `system`

PR #872, opened 2022-10-04
Add `sidebar.defaultMenuCollapseLevel` to control sidebar menu collapse level

robbear:
The `sidebar.defaultMenuCollapseLevel` property allows the theme to control to which level in the content folder hierarchy the collapse should be set should `sidebar.defaultMenuCollapsed === true`.

Notice in `examples/swr-site`, the _Advanced_ sidebar menu is now open upon render, since the theme has been updated with `sidebar.defaultMenuCollapseLevel` set to `3`.


robbear:
Some background on motivation. In our site, we have (for now) a blank nav bar. Our content organization is exposed through the sidebar menu which we show everywhere including the site's home page. The organizational view is essentially the folders under the `pages` directory. We want an initial menu view that is not completely collapsed, but not completely expanded. Typically, a level 2 view - the top level items and their direct child folders - is a good compromise. But without more than the binary `sidebar.defaultMenuCollapsed` option, we can't do this. Hence, the desire for a `sidebar.defaultMenuCollapseLevel` property which becomes pertinent when `sidebar.defaultMenuCollapsed === true`.

packages/nextra-theme-docs/src/types.ts
```diff
@@ -94,7 +94,8 @@ export interface DocsThemeConfig {
     labels: string
   }
   sidebar: {
-    defaultMenuCollapsed: boolean
+    defaultMenuCollapseLevel: number,
+    defaultMenuCollapsed: boolean,
```
B2o5T, line 98: it's a definitely nice feature, but let's stay with only one option to configure this behaviour - `defaultMenuCollapseLevel: number`, `defaultMenuCollapsed` should be removed

packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -142,6 +142,7 @@ export const DEFAULT_THEME: DocsThemeConfig = {
     labels: 'bug'
   },
   sidebar: {
+    defaultMenuCollapseLevel: 1,
```
B2o5T, line 145: currenly setting `defaultMenuCollapseLevel` to 1 or 2 is the same, it should be fixed
```
page1
page2
page3
 - page 3-1
 - page 3-2 <- because I expected to collapse them with setting 2
   - page 3-2-1
   - page 3-2-2
```

examples/docs/src/pages/themes/docs/configuration.mdx
```diff
@@ -238,6 +238,25 @@ export default {
...
+
+```jsx filename="theme.config.js"
+/**
+ * @type {import('nextra-theme-docs').DocsThemeConfig}
+ */
+export default {
+  defaultMenuCollapseLevel: 2
+}
+```
+
```
B2o5T, line 259: Thanks for the docs! Can you move it to `./docs` folder, because `example/docs` it's not a real Nextra v2 docs but just an example

B2o5T:
@robbear I pushed my review fixes, you need only move your docs section!

B2o5T (approved):

examples/docs/src/pages/themes/docs/configuration.mdx
```diff
@@ -238,6 +238,25 @@ export default {
...
+**Default:** `1`
+
+**Example:**
+
+```jsx filename="theme.config.js"
+/**
+ * @type {import('nextra-theme-docs').DocsThemeConfig}
+ */
+export default {
+  defaultMenuCollapseLevel: 2
```
B2o5T, line 256: this should be
```
sidebar: {
  defaultMenuCollapseLevel: 2
}
```

robbear:
Does this PR make sense given https://github.com/shuding/nextra/pull/890? Doesn't that give finer control and add more clarity, achieving the same results?

B2o5T:
> Does this PR make sense given #890? Doesn't that give finer control and add more clarity, achieving the same results?

For me both PR makes sense, they make a configuration of `collapse` either globally or individually by page's `_meta.json#theme` option

examples/docs/src/pages/themes/docs/configuration.mdx
robbear: The actual docs page is laid out a bit differently, so this example here doesn't really matter, right? I've updated the docs page as I think I understood you to request.

packages/nextra-theme-docs/src/constants.tsx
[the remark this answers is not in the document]
robbear: I think this default value is sort of biased by how you organize your `pages` folder. If you do something like `pages/docs/blog-theme`, `pages/docs/docs-theme`, where `docs` is kind of a root, versus `pages/item-1/[stuff]`, `pages/item-2/[stuff]`, your expectations for the default value of `defaultMenuCollapseLevel` will differ. Seeing `2` as a default does feel a bit weird.

Issue #330, opened 2022-01-18
cant's use docsRepositoryBase with github enterprise


maticrivo:
Hi, when setting `docsRepositoryBase` with a self-hosted github enterprise version its throwing the error:
```
Invalid `docsRepositoryBase` URL!
```

After some investigation its coming from the `parse-git-url` package that does not handle self-hosted git server urls..

is there a way to add a workaround for that?

Thanks

johnbellone:
I am running into same issue here.

always-maap:
This is just a questionable design

B2o5T:
will be available in next release

maticrivo:
@B2o5T   Thanks for resolving this but honestly it has been 9 months since i opened the issue and as there was no response i have switched to use docusaurus 🙃

nkechr1s:
Is there any way to remove it ?

PR #879, opened 2022-10-11
feat: ability to use custom domain GitHub

trummerfeld:
I've stumbled across https://github.com/shuding/nextra/issues/330 myself and tried to get Nextra running for my Enterprise GitHub instances that use custom domains.

This change focuses on using making small changes to Nextra instead of `parse-git-url `.

Any feedback is welcome or otherwise put it to the bin.


B2o5T:
I don't think we need 2 more theme options  `docsRepositoryBaseCustom` and `docsRepositoryBaseCustomBranch`, this should be fixed directly in `parse-git-url` or another package should be used instead

trummerfeld:
That can be a solution too, the easy way was to add another configuration.
I will get my head around `gut-url-parse` as an alternative and see what I can do with that. 

Feel free to close this PR, thank you for your feedback.

Issue #838, opened 2022-09-12
Tabs should have default margin for inner content


shuding:
![image](https://user-images.githubusercontent.com/3676859/189666182-91617b48-37e8-469e-b6b8-6e01ceaa6f38.png)

B2o5T:
Yeah, you are right, saw this unexpected behaviour 2 months ago already 😅

Issue #863, opened 2022-09-27
Improve formatting (huge empty line)


raproid:
The formatting is partially broken and needs love.

1. An introductory sentence + list — the list is too far out after the sentence, i.e. separated by a huge empty line.
<img width="872" alt="192365558-7e783706-2be1-457a-a178-4bcfcd398f3c" src="https://user-images.githubusercontent.com/22454517/192404726-3c6c7594-83fb-4e7c-a7f2-88d88f7ff8ad.png">

2. Sublists — a sublist is too separated by a huge line that is higher than a normal one.
<img width="438" alt="192365561-8f0a5d9e-a580-462d-b152-a2bdb20b75ea" src="https://user-images.githubusercontent.com/22454517/192404767-5addfd35-df20-420f-ac74-2e66880181c2.png">

3. Paragraphs in callouts — separated by huge line.
<img width="872" alt="192365553-d732a6d3-a2f5-46f8-8b59-4bf92ff0bb91" src="https://user-images.githubusercontent.com/22454517/192404776-c1fca2c2-90c8-4e30-8d57-04ee1fa7c184.png">

raproid:
@shuding

raproid:
@shuding

raproid:
May I suggest a solution vector for the nested lists problem? 

ul.mt-6,
ol.mt-6 {
  margin-top: 0.6em;
  margin-bottom: 0.5em;
}

ben-z:
Ran into this today. In my use case, only the nested lists where the parent has an extra element were affected:

By default, Nextra uses `:first-child`. And in the case above, the first child is the `a` element.

<img width="1413" alt="image" src="https://github.com/shuding/nextra/assets/5977478/7cae1080-7b53-43d8-9cbe-0aa27314ac7c">

Handled by using `:first-of-type` instead. I added this to my stylesheet:

```css
ul.first\:nx-mt-0:first-of-type,
ol.first\:nx-mt-0:first-of-type {
  margin-top: 0;
}
```

Before:

<img width="846" alt="image" src="https://github.com/shuding/nextra/assets/5977478/9f200025-c047-4a7d-9c07-9100fafb50a8">

After:

<img width="845" alt="image" src="https://github.com/shuding/nextra/assets/5977478/3ec54109-9c67-403a-a56d-d39d84f0fa07">

dimaMachina:
upstream https://github.com/shuding/nextra/issues/16

Issue #16, opened 2020-11-12
Use @tailwind/typography


shuding:
Let's use https://tailwindcss.com/docs/typography-plugin for article styling.

ng-hai:
I think I can contribute to this, but not sure where to put the plugin. Should it live in its `theme` plugin or project's default `tailwind.config.js`?

pacocoursey:
Best to add it to each theme individually @ng-hai, meaning here:
https://github.com/shuding/nextra/blob/core/packages/nextra-theme-docs/tailwind.config.js

Remember that `core` is the development branch.

ng-hai:
I have added the plugin locally, and noticed these UI bugs

<img width="788" alt="image" src="https://user-images.githubusercontent.com/19520278/108674873-8b336c00-7518-11eb-8abe-2f94b9fbbe1d.png">

The code block background changed to dark.

<img width="184" alt="image" src="https://user-images.githubusercontent.com/19520278/108674962-aef6b200-7518-11eb-9fd0-68bd46d2332b.png">

<img width="279" alt="image" src="https://user-images.githubusercontent.com/19520278/108675210-0c8afe80-7519-11eb-9920-4a9aeab2e66a.png">

List items appended new prefix.

<img width="747" alt="image" src="https://user-images.githubusercontent.com/19520278/108675262-22002880-7519-11eb-8313-29f8f60275a2.png">

<img width="323" alt="image" src="https://user-images.githubusercontent.com/19520278/108675294-30e6db00-7519-11eb-8fbe-b9bf65ec3a84.png">

Spacing and colors get wrong.

Should I change the MDX Components styles, or customize the plugin styles itself?

shuding:
I think it's better to change the styles of the components @ng-hai

thien-do:
Oh I just noticed that the "blog" theme already has this "typography" plugin :D only the "docs" is needed

AntoineKM:
@thien-do since `@tailwind/typography` was only for articles, I think this issue is fixed.

PR #891, opened 2022-10-13
fix margin-top for `<Tab />` content

B2o5T:
fixes https://github.com/shuding/nextra/issues/838

### DIFF
--- a/packages/nextra-theme-docs/src/mdx-components.tsx
+++ b/packages/nextra-theme-docs/src/mdx-components.tsx
@@ -179,7 +179,7 @@ const Details = ({
 
   return (
     <details
-      className="my-4 rounded border border-gray-200 bg-white p-2 shadow-sm dark:border-neutral-800 dark:bg-neutral-900 first:mt-0 last:mb-0"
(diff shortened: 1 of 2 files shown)


packages/nextra-theme-docs/src/mdx-components.tsx
```diff
@@ -179,7 +179,7 @@ const Details = ({
 
   return (
     <details
-      className="my-4 rounded border border-gray-200 bg-white p-2 shadow-sm dark:border-neutral-800 dark:bg-neutral-900 first:mt-0 last:mb-0"
+      className="my-4 rounded border border-gray-200 bg-white p-2 shadow-sm dark:border-neutral-800 dark:bg-neutral-900"
```
B2o5T, line 182: does there any reason to have `first:mt-0` class for details, ul, blockquote, table, p elements

shuding: I think that makes it look better when there is nothing before the list and the title, e.g.:

```md

### title

<details>...</details>

```

B2o5T: I think we can merge it and add tailwind typography (that was proposed in https://github.com/shuding/nextra/issues/16 issue) in another PR to fix https://github.com/shuding/nextra/issues/863

Issue #113, opened 2021-03-27
Generate OpenGraph Tags from Frontmatter


LucaNerlich:
It would be greate, if (for each page) the opengraph tags could be generated based on the frontmatter from the corresponding markdown file. This would then generate the "share" preview card, when the blog url gets posted on social media, discord, slack etc.

Example: 
```markdown
---
title: my blog test
date: 2021/3/27
tag: code, tech
author: Luca Nerlich
description: test description
image: /images/blog/test.jpg
---
```

Example OG Tags

https://ogp.me/

```html
<html prefix="og: https://ogp.me/ns#">
<head>
<title>my blog test</title>
<meta property="og:title" content="The Rock" />
<meta property="og:type" content="article" />
<meta property="og:url" content="<get this from next router>" />
<meta property="og:image" content="<get this from next router>/images/blog/test.jpg" />
...
</head>
...
</html>
```

RayhanADev:
I agree, that is a good idea!

shuding:
Great feature to have, thanks!

buccalon:
It would be nice to have an option to define custom meta tags too.
In my case, I'd like to have each page accessible to citation managers like Zotero.
See: https://www.zotero.org/support/dev/exposing_metadata

B2o5T:
Integration of `next-seo` will be available in the future beta release.

you can take a look at an example of the configuration of `getNextSeoProps()` theme option here https://github.com/shuding/nextra/blob/core/examples/docs/src/theme.config.js#L20-L73

Also you can access to parsed frontMatter data in `getNextSeoProps` function via `useConfig` hook.

PR #892, opened 2022-10-14
setup `next-seo`

B2o5T:
fixes https://github.com/shuding/nextra/issues/113
- setup `next-seo`
- add new theme option `getNextSeoProps`
- remove `titleSuffix` theme option in favor of `getNextSeoProps.titleTemplate`
- by default pass `description`, `canonical`, `openGraph` values to `<NextSeo />` component  from page `frontMatter`, values can be overridden with return value of `getNextSeoProps`

### DIFF
--- a/.changeset/silver-students-return.md
+++ b/.changeset/silver-students-return.md
@@ -1,0 +1,8 @@
+---
+'nextra-theme-docs': patch
+---
+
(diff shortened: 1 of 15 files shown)


shuding (approved):
This is amazing, I didn't expect it to be so smooth to add `next-seo` 👍

Issue #874, opened 2022-10-06
Switching between tabs causes jumping to the last clicked Heading in the On This Page menu


galleon8:
URL to check the behavior: https://www.ankr.com/docs/advanced-api/query-methods/

1. At the URL provided, click any Heading in the On This Page menu (the one on the far right).
2. Scroll to any code snippet with tabs (for better illustration, use the one a bit further from your current position on the page).
3. Switch between tabs.

Here you are — you've just jumped to the Heading previously clicked in the menu!

promer94:
https://github.com/shuding/nextra/pull/567 causes this problem. 

Switching tabs changes the document's height and triggers the ResizeObserver
https://github.com/shuding/nextra/blob/f7856a188b124c44cbe69e856a5c9c3311ac5833/packages/nextra-theme-docs/src/index.tsx#L29-L36

PR #893, opened 2022-10-14
remove `resizeObserver` for dynamic content

B2o5T:
fixes https://github.com/shuding/nextra/issues/874
revert of https://github.com/shuding/nextra/pull/567

Dynamic content in documentation is not very often, but `#` anchors in URL + clicking for `<Tab />` more more often.

This PR remove `resizeObserver` for dynamic content since it provokes jumps on the tabs switches when there is `#` anchor in URL


robbear:
This kind of content, as demonstrated in the scroll-with-dynamic-height example, is pervasive in our documentation site. Without the `resizeObserver`, the page scroll resolves to just about anywhere but the targeted anchor. This would be a significant breaking change for us.

B2o5T:
@robbear I am sorry, but your feature introduce more confused behaviour for other users that don’t use dynamic content but uses Tabs instead, if you want to work on fixing it - feel free, but I think there is no solution without removing this observable

robbear:
@B2o5T agreed - I hadn't considered the user-driven dynamic resizing possibility such as the tabs semantics. I think we have alternate approaches to our issue. Thanks for the feedback.

PR #768, opened 2022-08-31
add `DocsThemeConfig.header` config option for overriding navbar

B2o5T:

Issue #745, opened 2022-08-25
Make it easier to fork/customize `nextra-theme-docs`


gkaemmer:
We've been using [Nextra at Oso](https://www.osohq.com/docs) and it's been working well. Thanks for putting so much effort into it :). 

There are increasingly many requirements that pull us farther from the off-the-shelf docs theme, but there's still a lot of functionality from the docs theme that we don't want to maintain ourselves. The main things we value about `nextra-theme-docs`
- The sidebar nav: parsing nextra's file tree and `meta.json` to build a usable tree in the sidebar
- Parsing the markdown to build a TOC on the right
- (soon) flexsearch functionality

But there are a lot of things we want to customize, like:
- Buttons in the top navbar (both on the left and the right sides)
- Which folders in the sidebar are default open/closed (as well as hiding some behind a "More" toggle)
- Sidebar styling things, specifically to more clearly differentiate sections of docs ("separators" don't really work for us)
- "Index" pages that link to all of their children

I'm sure that list will grow as time goes on. To do that stuff, right now we maintain [a fork of `nextra`](https://github.com/osohq/nextra) but it's a bit tedious to maintain. We only really need a fork of `nextra-theme-docs`, but because nextra is a monorepo we have to fork the whole thing. (you can see the diff [here](https://github.com/shuding/nextra/compare/core...osohq:nextra:master)). I suspect most users of `nextra-theme-docs` will eventually need to similarly fork the theme to add little customizations here and there.

Some things that might make this easier:
- Put `nextra-theme-docs` in it's own repo so that it's easier to fork (that way, we'd only need to maintain a fork of the code that we're actually interested in customizing, not the whole `nextra` library)
- Somehow make it easier to completely override sections of the page (by providing, say, a custom `Navbar` component, or a custom `Folder` component for the sidebar). This seems like it'd be hard, but would make `nextra-theme-docs` more powerful without needing to fork.

Also interested in other folks' solutions to this problem!

B2o5T:
For the things to customization, please create separate issues with screenshots if possible, this will be easy to keep track and approve

And for about

> Put nextra-theme-docs in it's own repo...

![](https://c.tenor.com/3EIa6ycgpsUAAAAM/no-noooo.gif)

Because it’s a step back, this will be harder for us to maintain

shuding:
[...]

PR #907, opened 2022-10-18
allow adding additional navbar content via `navbar.extraContent`

Toalaah:
_tldr_: This PR introduces the ability to add additional content to the navbar in the theme config, analogously to `main.extraContent` or `toc.extraContent`.
<br>
We found the need for this functionality while migrating our existing documentation system to nextra at work. We require an additional login button—ideally in the navbar—for allowing certain clients to authenticate in order to access restricted content. While searching for a solution I stumbled upon #768, however this is not quite what I was looking for as we simply want to augment the navbar with one or two additional components, not rewrite it in it's entirety as its functionality actually suits are needs quite well (other than the aforementioned issue). As far as I could tell this is currently not possible.

I can imagine that this situation is not too uncommon / niche, as this already seems to have been raised before in #745.

### DIFF
--- a/.changeset/angry-balloons-end.md
+++ b/.changeset/angry-balloons-end.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+allow adding additional navbar content via `navbar.extraContent`
--- a/packages/nextra-theme-docs/src/components/navbar.tsx
+++ b/packages/nextra-theme-docs/src/components/navbar.tsx
@@ -184,6 +184,8 @@ export function Navbar({ flatDirectories, items }: NavBarProps): ReactElement {
           </Anchor>
         ) : null}
 
+        {renderComponent(config.navbar.extraContent)}
+
         <button
           className="nextra-hamburger rounded active:bg-gray-400/20 p-2 -mr-2 md:hidden"
           onClick={() => setMenu(!menu)}
--- a/packages/nextra-theme-docs/src/constants.tsx
+++ b/packages/nextra-theme-docs/src/constants.tsx
@@ -90,7 +90,10 @@ export const DEFAULT_THEME: DocsThemeConfig = {
   main: {
     extraContent: null
   },
-  navbar: Navbar,
+  navbar: {
+    component: Navbar,
+    extraContent: null
+  },
   navigation: {
     next: true,
     prev: true
--- a/packages/nextra-theme-docs/src/index.tsx
+++ b/packages/nextra-theme-docs/src/index.tsx
@@ -198,7 +198,7 @@ const InnerLayout = ({
       <Head />
       <Banner />
       {themeContext.navbar &&
-        renderComponent(config.navbar, {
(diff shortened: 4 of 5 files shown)


B2o5T:
I like it, definitely, we should have a flexible way to configure extra elements in navbar.

The only thing I see to improve is a similar refactor as I did here: https://github.com/shuding/nextra/pull/886 (to avoid adding a new theme option for _prefix_ content in the future for some cases). And let's still uses `navbar` theme option name instead of `navbar.extraContent`.

Also seems like your change and [this merged PR](https://github.com/shuding/nextra/pull/908/files) by @tknickman make the same things but in a more abstract way, because their custom project/chat element will be rendered only if:

1. project.link/chat.link should be empty string `''`
2. project.icon/chat.icon should be defined

In your case will just put customizable content in `navbar.extraContent` and it all.

B2o5T (approved):

Toalaah:
Hi, first of all sorry for the long incoming wall of text, just had some thoughts on what you mentioned. You can read my comments specifically in regards to **this** PR in the last paragraph in case you want to skip.

You make a good point with the similarities the merged #908 PR. I do think that supporting both is still valuable as in _most_ situations I can see users being happy with simply changing a chat links icons, and in this situation having an easy way to do so without forcing them to create a whole new component is definitely a plus in terms of the plugin's ergonomics.

The `extraContent` to me is more of an additional option to allow _just enough_ more flexibility with navbar customization. IMO if both `extraContent` and `chatLink` / `chatIcon` / etc. don't cover your needs anymore, you are probably better off doing the whole navbar from scratch.

It almost makes me wonder though whether it wouldn't make more sense to put these navbar-related config options (`project.Icon`, `chat.Icon`, etc.) into the `config.navbar` opts to consolidate what the available customization options are. This would also allow you to make the whole choice of a completely custom navbar component and and choosing these options mutally exclusive. As an example, it doesn't really make sense to allow both `config.project.icon` and `navbar.Component` to be set, as the custom nav-component will ignore the custom project icon. So something like this:

```typescript
// ...
navbar: ReactNode | React.FC<NavBarProps> | {
    chat: {
        icon: ReactNode | React.FC
        link: string
    },
    extraContent: ReactNode | React.FC,
    project: {
        icon: ReactNode | React.FC
        link: string
    },
     ...
}
// ...
```

But as you already mentioned something like this would require serious refactoring and I don't really know if this is worth it.

In terms of this MR, I am pretty much satisfied with the added functionality. The only thing that perhaps could be changed is that currently these components would probably be hidden on mobile devices as they are not injected into the hamburger menu, but TBH I havn't spent enough time with the code to judge how much work this would be. Unless you have any other concerns or change request, feel free to merge if you like, am looking forward to hopefully using this in the future!

B2o5T:
@Toalaah I like your proposal, but let's merge/release this feature first

Toalaah:
@B2o5T sounds good. If there is anything else that needs to be done before this can be merged, feel free to let me know, thanks!

PR #908, opened 2022-10-19
chore(docs): export ThemeSwitch & custom social

tknickman:
1. Export ThemeSwitch to allow using this in custom docs footers
2. Support fully custom social links


shuding (approved):
Makes sense, thanks!

raproid:
@B2o5T how to use the custom social links: what param should I declare and where? And can I add a custom icon per link?

PR #911, opened 2022-10-20
Add `nx-` tailwind prefix

shuding:
We need to isolate these classnames in case the website itself also uses Tailwind.

~_Seems like dark + prefix doesn't always work: `dark:nx-shadow-[0_-12px_16px_#111]`_~ _changing to `darkMode: ['class', 'html[class~="dark"]']` fixes it, see https://github.com/tailwindlabs/tailwindcss/issues/5754_

### DIFF
--- a/.changeset/hot-rockets-sip.md
+++ b/.changeset/hot-rockets-sip.md
@@ -1,0 +1,7 @@
+---
+'nextra': patch
+'nextra-theme-blog': patch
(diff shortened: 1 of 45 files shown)


shuding:
cc @B2o5T this is a pretty big refactoring and it will affect some other PRs 🙏 (however this should not break users' code unless they're doing something wrong)

PR #947, opened 2022-10-28
fix project/chat icon could appear without provided link

B2o5T:
I was discouraged by the appearance of the discord icon without specifying the discord link in `theme.config` 

Custom elements in NavBar should be added by this new option https://github.com/shuding/nextra/pull/907, current behaviour changed by @tknickman is confused

image

### DIFF
--- a/.changeset/metal-laws-joke.md
+++ b/.changeset/metal-laws-joke.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+fix project/chat icon could appear without provided link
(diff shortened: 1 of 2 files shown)


B2o5T:
> i.e. you're in control of rendering the link, or not if you so choose. This is helpful because you may not want to render the icon at some breakpoints for example.

you can do the same with the proposed `main.extraContent` option https://github.com/shuding/nextra/pull/907

The problem is that now if we don't specify the `project.link` or `chat.link`, we'll still see them in NavBar, and it confuses because we didn't configure them, similar issue https://github.com/shuding/nextra/issues/950

B2o5T:
@shuding I will merge this PR to avoid confusing unclickable chat icon, because https://github.com/shuding/nextra/pull/907 was merged and customization additional content is available via `navbar.extraContent`

PR #952, opened 2022-10-29
remove redundant `<div data-rehype-pretty-code-fragment />` element that wrap `<pre />` element

B2o5T:


B2o5T:
implemented in https://github.com/shuding/nextra/pull/1253

PR #978, opened 2022-11-22
Support LaTeX

### LINKED ISSUE #416: LaTeX support
Planning on using nextra with the docs template for a new project, and I was wondering:

- Is it possible to get LaTeX support inside the docs?
- The font seems to have changed. Would it be possible to get back to the old one (Inter)? 

Thanks!

robbear:
Fixes #416 

Screen Shot 2022-11-23 at 2 21 10 PM

Could we do something like this? We might want a theme flag to avoid the download of `katex.min.css` if a theme-user is not interested in math rendering.

By the way, please note that I had to pin `"flexsearch": "0.7.21"` on a clean rebuild of the `pnpm-lock.yaml` to get around a failure in `pnpm install`.


Subham-Maity:
@robbear Sir, does this work with yarn?

robbear:
> does this work with yarn?

@Subham-Maity I have not built with yarn, only pnpm, but if the project has otherwise been working for you with yarn, I don't see any reason why these changes would not.

packages/nextra-theme-docs/package.json
```diff
@@ -48,7 +48,7 @@
     "@popperjs/core": "^2.11.6",
     "@reach/skip-nav": "^0.17.0",
     "clsx": "^1.2.1",
-    "flexsearch": "^0.7.21",
+    "flexsearch": "0.7.21",
```
hisamafahri, line 51: > By the way, please note that I had to pin "flexsearch": "0.7.21" on a clean rebuild of the pnpm-lock.yaml to get around a failure in pnpm install.

@robbear would you mind share the error you've been getting? Since I tried on my machine everything works well.

package.json
```diff
@@ -38,6 +38,7 @@
   ],
   "pnpm": {
     "overrides": {
+      "flexsearch": "0.7.21",
```
hisamafahri, line 41: CMIIW, but I don't think this is package is necessary in the root `package.json`

```suggestion

```

shuding:
I think for this to land, we need to find a way to inject the KaTeX CSS dynamically instead of relying on a CDN. Thanks for opening this!!

robbear:
> I think for this to land, we need to find a way to inject the KaTeX CSS dynamically instead of relying on a CDN.

@shuding I replaced the CDN reference with `import "katex/dist/katex.min.css"` in `swr-site/pages/_app.js`. It was helpful to discover that `katex.min.css` is bundled with `katex`.

robbear:
> Someone is attempting to deploy a commit to a Personal Account owned by @shuding on [Vercel](https://vercel.com/).
@shuding first needs to [authorize it](https://vercel.com/git/authorize?type=github&job=%7B%22headInfo%22%3A%7B%22sha%22%3A%226e6c90b7c5bbeb09402b207a7558621ccdad3271%22%7D%2C%22id%22%3A%22QmXpdk5BjsfeShmE3uPmj6FNNrrkvcZBoimbpMGp4ZywMu%22%2C%22org%22%3A%22shuding%22%2C%22prId%22%3A978%2C%22repo%22%3A%22nextra%22%7D).

This sounds scary, and was most likely me. Do I have something configured improperly with regard to my fork?

packages/nextra-theme-docs/package.json
robbear: I was seeing this when I:

1. Start from a fresh clone (specifically, no `node_module` contents)
2. Remove `pnpm-lock.yaml` so as to regenerate it
3. `pnpm install`

image

Pinning `flexsearch` to `0.7.21` resolves it.

package.json
robbear: What this line in the root `package.json` under `pnpm/overrides` does is force the workspaces to use the specified `flexsearch` version. I didn't actually edit the `packages/nextra-theme-docs/package.json` line. The act of running `pnpm install` did so as a result of the override instruction.

shuding:
Ah @robbear not because of your fork, it was me misconfigured the Git Fork Protection option on Vercel - resolved now! Sorry for the inconvenience!

shuding:
Implemented by #1143, also appreciate your input here! 🙌

PR #991, opened 2022-12-02
Polish the docs theme; rename options

shuding:
This PR removes the `unstable_` prefix for flexsearch, staticImage and the copy button because they're pretty stable and safe to ship in 2.0.

Also I adjusted some shadow, border, and scrollbar styles. And fixed a bug about separators in the sidebar.

### DIFF
--- a/packages/nextra/src/types.ts
+++ b/packages/nextra/src/types.ts
@@ -92,7 +92,7 @@ export type NextraConfig = {
   defaultShowCopyCode?: boolean
   flexsearch?: Flexsearch
   staticImage?: boolean
-  unstable_readingTime?: boolean
+  readingTime?: boolean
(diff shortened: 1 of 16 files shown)


packages/nextra/src/types.ts
```diff
@@ -89,9 +89,9 @@ type Flexsearch = boolean | { codeblocks: boolean }
 export type NextraConfig = {
   theme: Theme
   themeConfig?: string
-  unstable_defaultShowCopyCode?: boolean
-  unstable_flexsearch?: Flexsearch
-  unstable_staticImage?: boolean
+  defaultShowCopyCode?: boolean
+  flexsearch?: Flexsearch
+  staticImage?: boolean
   unstable_readingTime?: boolean
```
B2o5T, line 95: this can be renamed too :)

shuding: OK!

PR #979, opened 2022-11-22
feat(app): add CODE_OF_CONDUCT

bmuenzenmeyer:
noticed this project does not have a CODE OF CONDUCT. It seems large enough to warrant one, though @shuding shoulders the decision and the moderation responsibility

PR #1009, opened 2022-12-03
feat(app): add CODE_OF_CONDUCT

bmuenzenmeyer:
This file was introduced within #979 but I think it got blown away somehow with the recent launch of the new nextra.

Hoping it is still welcome here.


shuding:
Thanks!

Issue #1010, opened 2022-12-03
a11y issues


mayank99:
I was scrolling through the docs and a few different accessibility issues immediately stood out to me:

![annotated screenshot of a section of the nextra docs highlighting a few issues: sticky header will hide focusable elements. color used for conveying state in header nav. poor contrast in header search and side nav. focus stays on link after navigating. links open in new tab without informing screen readers. copy to clipboard button in code snippets is not keyboard accessible](https://user-images.githubusercontent.com/9084735/205448227-fd8c0580-4556-4e5a-9f74-3afe495dc07d.png)

Running automated a11y tools like [axe devtools](https://www.deque.com/axe/devtools/) reveal a few dozen more issues.

A lot of these issues are simple fixes, but some of them might require rethinking the UI. Also some of these issues are confined to the nextra docs, but many of them will likely affect other sites built using nextra.

Additionally, the landing page has its own set of issues but I'd imagine the landing page is fully custom (not built using Nextra for the most part), so it's less important for users. In the screenshot below, the left card has low contrast and the right card has inaccessible text (it's cut off at ~400px browser width).
![one card from the nextra home page has white text on a light blue background. another card next to it reads "A11y as a top priority." in big bold letters. the next line attempts to say "Nextra respect system options such as Increase Contrast and Reduce Motion" but the second line gets cut off by an image.](https://user-images.githubusercontent.com/9084735/205448527-7a473739-98b0-486c-a4e9-553d43eeb0de.png)

GeekyMax-Ytiam:
The poor contrast was the main issue with this nextra I faced the same issue

PR #1013, opened 2022-12-03
make code-blocks buttons focusable if they are visible on page

B2o5T:
also, I removed all `unstable_` suffixes in repo

related https://github.com/shuding/nextra/issues/1010


shuding:
Thanks, although they don't seem to be tab-focusable?

Issue #641, opened 2022-08-05
[theme-blog] Empty tag name in title


numToStr:
### Problem

When you visit `type: tag` page; the tag name is missing from the page title. You can visit the https://site-k83dk8pms-numtostr.vercel.app/tags/neovim to see the missing tag name from the title

Corresponding MDX: https://github.com/numToStr/site/blob/main/pages/tags/%5Btag%5D.mdx

- Expected: `Tagged Posts - "neovim"`
- Actual: `Tagged Posts - ""`

### Packages
```json
"next": "^12.2.3",
"nextra": "2.0.0-beta.14",
"nextra-theme-blog": "2.0.0-beta.14",
"react": "^18.2.0",
"react-dom": "^18.2.0"
```

PR #1014, opened 2022-12-03
feat: static tag title 

promer94:
statically generate tags for blog theme
close https://github.com/shuding/nextra/issues/641


shuding:
Wow this is very interesting... thanks!

PR #1026, opened 2022-12-04
[Docs] Fixing key name for footer content

liarco:
The documentation uses `footer.content` while the actual code is using `footer.text`.

Here is the code reference:
https://github.com/shuding/nextra/blob/e1fc506a0bdf73fe9fcf679e3de39918697aa352/packages/nextra-theme-docs/src/components/footer.tsx#L28

---

I also noticed an extra `$` in the code example and I removed it, here is the code reference:
https://github.com/shuding/nextra/blob/73241b3fee58f0318e9349a5eb7650a7cd55a4e3/docs/pages/docs/docs-theme/theme-configuration.mdx?plain=1#L470

### DIFF
--- a/docs/pages/docs/docs-theme/theme-configuration.mdx
+++ b/docs/pages/docs/docs-theme/theme-configuration.mdx
@@ -467,7 +467,7 @@ You can add some simple content, such as copyright information to the default fo
 export default {
   footer: {
     text: <span>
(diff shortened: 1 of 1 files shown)


B2o5T (approved):

B2o5T:
nice catch, ty

PR #1071, opened 2022-12-11
fix missing border in navbar

B2o5T:
@shuding on a dark theme we no longer need a border?

as border is missing on latest preview 
image

### DIFF
--- a/.changeset/odd-pants-grow.md
+++ b/.changeset/odd-pants-grow.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+fix missing border in navbar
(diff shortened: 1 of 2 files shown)


shuding:
Ah @B2o5T I intentionally changed that to be not inset, to show the box shadow on top of the underlying content. For example `inset`:

CleanShot 2022-12-11 at 20 30 20@2x

Without `inset` (looks much better):

CleanShot 2022-12-11 at 20 30 43@2x

B2o5T:
@shuding did you take a look at dark mode? Because in light mode there is a border and in dark - not

shuding:
I'm fixing it now on this PR!

B2o5T:
Thanks 🙏❤️

PR #1101, opened 2022-12-19
fix(nav-links.tsx): fix word breaks and alignment in Safari

jaredLunde:
I might need some clarity on the intent of the ltr/rtl selectors, but this fixes an issue where the flex children do not grow in Safari, causing words to break in odd places.

Additionally, because the `<a>` elements have a `flex` display, `text-align` has no effect.

### Before

Screen Shot 2022-12-19 at 9 39 42 AM

### After

Screen Shot 2022-12-19 at 9 41 41 AM

### DIFF
--- a/packages/nextra-theme-docs/src/components/nav-links.tsx
+++ b/packages/nextra-theme-docs/src/components/nav-links.tsx
@@ -13,7 +13,7 @@ interface NavLinkProps {
 
 const classes = {
   link: cn(
-    'nx-flex nx-grow nx-max-w-[50%] nx-items-center nx-gap-1 nx-py-4 nx-text-base nx-font-medium nx-text-gray-600 nx-transition-colors [word-break:break-word] hover:nx-text-primary-600 dark:nx-text-gray-300 md:nx-text-lg'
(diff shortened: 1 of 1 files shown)


B2o5T:
Hi! I can't reproduce it on my Safari 16.1, in which version you saw it? 
image

jaredLunde:
Whoops, should have included. This happens in macOS Safari 15.6.1 and iOS 16.0 Safari for me. Here are the configurations I'm using: 

```js
// postconfig.config.js

module.exports = {
  plugins: {
    tailwindcss: {},
    autoprefixer: {},
  },
};
```

```js
// next.config.js

const withNextra = require("nextra")({
  theme: "nextra-theme-docs",
  themeConfig: "./theme.config.tsx",
  flexsearch: {
    codeblocks: true,
  },
  staticImage: true,
  defaultShowCopyCode: true,
});
```

```js
// tailwind.config.js
module.exports = {
  darkMode: "class",
  content: [
    "./components/**/*.{js,jsx,ts,tsx}",
    "./pages/**/*.{js,jsx,ts,tsx,md,mdx}",
    "./theme.config.tsx",
  ],
  theme: {
    extend: {
      fontFamily: {
        sans: `"Roboto", system-ui, -apple-system, sans-serif`,
        mono: `"Roboto Mono", "Lucida Console", Monaco, monospace`,
      },
    },
  },
};
```

I'm also importing tailwind in a global stylesheet. Not seeing any differences in the computed styles though 🤔

jaredLunde:
Hmm strangely this has something to do with changing the root font size from `1rem` to `0.925rem` i.e.

```css
html[lang="en"] {
  font-size: 0.925rem;
}
```

PR #1113, opened 2022-12-23
fix(nextra-theme-docs): add aria label for banner dismiss button

boostvolt:
- Add aria label for banner dismiss button


B2o5T:
Thanks!

Issue #1118, opened 2022-12-24
Hydration failed because the initial UI does not match what was rendered on the server on version2


donghhan:
I am still having hydration error whenever I insulted my local image although I have recently upgraded to version 2. 

![image](https://user-images.githubusercontent.com/84304078/209435230-444cf21f-880f-40c1-999f-4ccb0aea23e9.png)

Here is an error log 

```
Unhandled Runtime Error
Error: Hydration failed because the initial UI does not match what was rendered on the server.

See more info here: https://nextjs.org/docs/messages/react-hydration-error

Call Stack
throwOnHydrationMismatch
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (12507:0)
tryToClaimNextHydratableInstance
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (12520:0)
updateHostComponent
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (19902:0)
beginWork
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (21618:0)
HTMLUnknownElement.callCallback
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (4164:0)
Object.invokeGuardedCallbackDev
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (4213:0)
invokeGuardedCallback
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (4277:0)
beginWork$1
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (27451:0)
performUnitOfWork
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (26557:0)
workLoopSync
node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js (26466:0)
renderRootSync
[...]

PR #1119, opened 2022-12-24
fix hydration for default `gitTimestamp` component

B2o5T:
fixes https://github.com/shuding/nextra/issues/1118


shuding (approved):
Thanks! P.S. Happy holidays

PR #1127, opened 2022-12-26
better performance of `collectFiles`

B2o5T:
add concurrency while reading files


shuding (approved):
Thanks! I believe this will be extremely helpful for cases like #1131. I'm thinking about making a perf test as well.

PR #1143, opened 2022-12-29
Add LaTeX support

### LINKED ISSUE #150: Math typesetting
Hi, I just wanted to ask if there was a plan to enable math equations? It seems to me this is not a trivial issue. 

Thanks!

jacobk999:
Closes #150, #416, and #978
image
- I used `KaTeX` with `remark-math` and `rehype-katex`
- It works by adding `latex: true` within `next.config.js`
- I tried writing some documentation shown in the picture above, they are located in `examples/docs`.


shuding:
Thanks, this basically looks good to me (comparing to #978 this one bundles the CSS and loads it when latex is enabled)! 

Do you mind adding it to the docs? Under https://nextra.site/docs/guide. Then it would be good to ship this!

jacobk999:
image
Currently, it causes a crash since MDX tries to parse the LaTeX as JavaScript. I pushed what docs should be but I was unsure how to go about making the docs use the local version of Nextra instead of `2.0.1`. The docs aren't part of the workspace so I don't think the `workspace:*` syntax would work.

shuding:
Oh right, that would require a separate PR to the docs after this change is released. Could you revert the commit? I can get this merged first.

jacobk999:
yup, done!

PR #1144, opened 2022-12-29
feat: Refactor internal APIs; improve HMR for meta and layout data

### LINKED ISSUE #1090: Hot reload of TOC sidebar not working
First of all thanks for this great framework. It is so easy and I love it. 💜

While working with it I noticed that the TOC sidebar on the right is not updating when I make changes to the headlines in `*.mdx` files. Not a big deal, but would be a nice touch if that would work when running `next dev`. :)

I use the latest version 2.0.1 with the docs template, the repo is here: https://github.com/reactjs/react-docgen/tree/main/packages/website

Thanks

shuding:
Fixes #1090, also things like titles in `_meta.json` can have HMR now.

Inside the loader, we track the checksum for `pageOpts`. When that changes, e.g. `headings` or `pageMap` got an update, we need to trigger a re-render of the layout via Webpack's HMR. For content update (`<Content>`), that's handled by fast refresh itself.

Also removed some unused fields from the `headings` AST node, such as `position`. That's also because they always affect the checksum even if headings are not edited.


packages/nextra-theme-docs/src/index.tsx
```diff
@@ -236,9 +236,11 @@ const InnerLayout = ({
 }
 
 export default function Layout(props: any): ReactElement {
-  const { route } = useRouter()
-  const context = globalThis.__nextra_pageContext__[route]
-  if (!context) throw new Error(`No content found for ${route}.`)
+  const context = usePageContext()
+  if (!context)
```
B2o5T, line 240: I guess this check we can add directly to use-page-context.ts? To avoid repetition in docs and blog also it will be available for custom themes :)

packages/nextra/src/hooks/use-page-context.ts
```diff
@@ -0,0 +1,41 @@
...
+  const rerender = useState({})[1]
+  const { route } = useRouter()
+
+  // TODO: Remove the HMR logic for prod builds
+  useEffect(() => {
+    const trigger = () => rerender({})
+
+    const listeners = __nextra_internal__.refreshListeners
+
+    if (!listeners[route]) listeners[route] = []
```
B2o5T, line 32:
```suggestion
    listeners[route] ||= []
```

packages/nextra/src/loader.ts
```diff
@@ -208,39 +231,55 @@ export default MDXContent`
...
+__nextra_internal__.context[${JSON.stringify(pageNextRoute)}] = {
   Content,
   pageOpts: __nextra_pageOpts__,
-  themeConfig: ${themeConfigImport ? '__nextra_themeConfig__' : 'null'}
+  themeConfig: ${themeConfigImport ? '__nextra_themeConfig__' : 'null'},
+}
+${'' /* TODO: Do not include all the HMR logic for prod */}
+if (module.hot) {
+  const checksum = "${pageOptsChecksum}"
+  if (!module.hot.data) module.hot.data = {}
```
B2o5T, line 274:
```suggestion
  module.hot.data ||= {}
```

packages/nextra/src/mdx-plugins/remark.ts
```diff
@@ -37,7 +37,8 @@ export const remarkHeadings: Plugin<[], Root> = function (this: Processor) {
               (child: { type: string }) => child.type === 'mdxJsxTextElement'
             )
           const heading = {
-            ...(node as Heading),
+            type: node.type,
```
B2o5T, line 40: Do we need `type`? I guess it always `type: heading`?

B2o5T (commented):
Could test only after 03/01

packages/nextra/src/hooks/use-page-context.ts
```diff
@@ -0,0 +1,41 @@
+import { useRouter } from 'next/router'
```
B2o5T, line 1: Test fails because we compile hooks to mjs extension, we should compile hooks to js or add .js extension here

packages/nextra/src/hooks/use-page-context.ts
```diff
@@ -0,0 +1,41 @@
...
+    >
+    refreshListeners: Record<string, (() => void)[]>
+  }
+
+  const rerender = useState({})[1]
+  const { route } = useRouter()
+
+  // TODO: Remove the HMR logic for prod builds
+  useEffect(() => {
+    const trigger = () => rerender({})
```
B2o5T, line 28:
```suggestion
    if (process.env.NODE_ENV === 'production') return
    const trigger = () => rerender({})
```

packages/nextra/src/loader.ts
```diff
@@ -208,39 +231,55 @@ export default MDXContent`
...
-// the route to identify it.
-globalThis.__nextra_pageContext__[${JSON.stringify(pageNextRoute)}] = {
+__nextra_internal__.context[${JSON.stringify(pageNextRoute)}] = {
   Content,
   pageOpts: __nextra_pageOpts__,
-  themeConfig: ${themeConfigImport ? '__nextra_themeConfig__' : 'null'}
+  themeConfig: ${themeConfigImport ? '__nextra_themeConfig__' : 'null'},
+}
+${'' /* TODO: Do not include all the HMR logic for prod */}
+if (module.hot) {
```
B2o5T, line 272: To remove todo comment?
```suggestion
if (process.env.NODE_ENV !== 'production' && module.hot) {
```

packages/nextra/src/loader.ts
```diff
@@ -208,39 +231,55 @@ export default MDXContent`
       // Remove the only `index` route
       .replace(/^index$/, '')
 
+  const stringifiedPageOpts = JSON.stringify(pageOpts)
+  const pageOptsChecksum = hashFnv32a(stringifiedPageOpts)
```
B2o5T, line 235: To avoid calling function on production
```suggestion
  const pageOptsChecksum = process.env.NODE_ENV !== 'production' && hashFnv32a(stringifiedPageOpts)
```

packages/nextra/src/loader.ts
```diff
@@ -208,39 +231,55 @@ export default MDXContent`
...
   pageOpts: __nextra_pageOpts__,
-  themeConfig: ${themeConfigImport ? '__nextra_themeConfig__' : 'null'}
+  themeConfig: ${themeConfigImport ? '__nextra_themeConfig__' : 'null'},
+}
+${'' /* TODO: Do not include all the HMR logic for prod */}
+if (module.hot) {
+  const checksum = "${pageOptsChecksum}"
+  if (!module.hot.data) module.hot.data = {}
+  if (module.hot.data.prevPageOptsChecksum !== checksum) {
+    __nextra_internal__.refreshListeners[${JSON.stringify(
```
B2o5T, line 276: We already call json stringify on line 266, we can assign it to a variable

packages/nextra/src/mdx-plugins/remark.ts
shuding: Yeah I think we can simplify this!

shuding:
Thanks @B2o5T! You can merge this whenever you think it's good to go.

B2o5T (approved):
It works like a charm, the one downside that I saw is that the header/footer is rerendered too now

PR #1160, opened 2023-01-01
Fixes dynamic head tags example

MikeCarbone:
While a key `frontMatter` does exist inside the object returned from `useConfig()`, the object is empty. The rest of the values we expect further in the example (like `frontMatter.title` and `frontMatter.description`) are present in the object returned from `useConfig()`. So we can do two things to improve this:

1. Not destructure and have `frontMatter` represent the values returned from `useConfig()` or

2. Destructure the values directly like so: `const { title, description } = useConfig()`

### DIFF
--- a/docs/pages/docs/docs-theme/theme-configuration.mdx
+++ b/docs/pages/docs/docs-theme/theme-configuration.mdx
@@ -117,7 +117,7 @@ import { useConfig } from 'nextra-theme-docs'
 export default {
   head: () => {
     const { asPath } = useRouter()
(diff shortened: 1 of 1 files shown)


B2o5T (changes requested):
sorry but I don't understand your change, `frontMatter` property exists on useConfig's return value
https://github.com/shuding/nextra/blob/e5262d08d8ee3c4e0ce9707546b0658297b6e31b/packages/nextra-theme-docs/src/contexts/config.tsx#L18-L19

Issue #1031, opened 2022-12-05
hydration failed at preview deploy


promer94:
[Preview deploy](https://nextra-v2-git-fork-promer94-fix-doc-preview-shuding1.vercel.app/docs) is not working currently. 

MDX component does not hydrate correctly in preview deploy.

<img width="1250" alt="Screenshot 2022-12-05 at 20 01 25" src="https://user-images.githubusercontent.com/15878786/205632637-2bbcd467-dc6a-4562-983d-05bf9821e6c5.png">

Production:
<img width="1128" alt="Screenshot 2022-12-05 at 20 01 06" src="https://user-images.githubusercontent.com/15878786/205632631-cd87dc83-15e6-4fa8-be40-704c14e01d8c.png">

B2o5T:
Closed https://github.com/shuding/nextra/pull/1168

Issue #852, opened 2022-09-16
Proposal: replace `_meta.json` in favor of `_meta.{js,jsx,ts,tsx}`


B2o5T:
First of all why?

1. to have the possibility to include repeated top content in `main` block

![image](https://user-images.githubusercontent.com/7361780/190829024-019c06ed-ba81-4daf-9ff8-651c2d809736.png)

2. to have possibility to include bottom content in `main` block

> currently it's possible via `main.extraContent` property but the source code became ugly since we have to check in which page we are and what we need to show

https://github.com/dotansimha/graphql-yoga/blob/1b018c3d2d347d3ca02eb0ca0175025325db6886/website/theme.config.tsx#L51-L64

3. to have possibility to sort sidebar items with the default eslint rule `sort-keys` and with eslint comment `/* eslint sort-keys: error */` and omit installation of some JSON plugin to do it

4. have type safe `_meta` file since we can import `Meta` type and have autocomplete

Example: current `_meta` file with included some top and bottom content will looks like

### before

https://github.com/shuding/nextra/blob/1a029e615a851d6ad7c1ad4238f455fee551b030/examples/swr-site/pages/docs/advanced/_meta.en-US.json#L1-L13

### after

```tsx
export default {
  "*": {
    theme: {
      footer: false,
      main: {
        topContent: <div>foo</div>,
        bottomContent: <span>bar</span>
      }
    },
  },
  "--- yoo": {
    title: "Do Not Use",
    type: "separator",
  },
  more: "More: A Super Super Super Super Long Directory",
  "file-name.with.DOTS": "Filenames with Dots",
};
```

@shuding any thoughts? :)

B2o5T:
💡 Also, we need to find a way to ignore accessing `_meta` file since Next.js will treat it as page, example /docs/_meta should get a 404 error

shuding:
I like the idea! Allowing `_meta` to be JavaScript brings more dynamicity indeed. And I'm a big fan of the type safety feature. Although I assume that it will be a bit tricky to support anything other than `_meta.js`, because we need the data inside the loader logic which is Node.js and can't make it easy to support TS and JSX there (and ESM too).

B2o5T:
> Although I assume that it will be a bit tricky to support anything other than _meta.js

You are right, I guess `_meta.mjs` will be a better extension since Next.js will treat `.js` files under the `/pages` directory as Page and since it's not exporting React component, Next.js throw an error

[...]

Issue #20, opened 2020-11-20
Versioned docs


jaredpalmer:
Kind of like next.js docs load the route manifest file from github version tags, would be rad to use same technique somehow.

helloworld:
+1 it would be very useful to see versioning implemented. Our team had to spend a lot of time building a custom documentation site that supported versioning since we couldn't find any other existing solutions.

B2o5T:
We missing this feature for our websites and I can work on this next week! Cc @shuding

shuding:
@B2o5T I have some thoughts regarding this, let's find some time to discuss next week 👍. In general it should be similar to remote MDX but more

B2o5T:
will break the silence 🤐

@shuding I think we can support different kinds of versioning docs

## 1. a simple way, via regular folders like below

<img width="219" alt="image" src="https://user-images.githubusercontent.com/7361780/189003712-754a632d-8480-4a2c-88c8-7047b69435d2.png">

## 2. via some advanced technique like remote downloading from some specific commit history hash

The WIP for the first one was implemented today and we'll test it soon on GraphQL Yoga v3

Let's start by small and improve it!

https://user-images.githubusercontent.com/7361780/189004955-15ece872-b416-4e00-8604-0c3635063be1.mov

I'll prepare soon my initial implementation PR

shuding:
I think 2 is the way to go, but curious about your solution for the meta information. 👀 @B2o5T 

In the long term I believe that “remote MDX” should be solved by React Server Components, but that would require a big refactoring of the framework (looking forward to that future).

B2o5T:
@shuding 
> but curious about your solution for the meta information

source code is [here](https://github.com/shuding/nextra/pull/840/files#diff-10a6386078f0b61870dacfdc58b870e91da110ce041b5d4eeab275569e249975) ;)

> In the long term I believe that “remote MDX” should be solved by React Server Components, but that would require a big refactoring of the framework (looking forward to that future).

What yours through about future problems with remote versioning:
- how to edit remote docs if legacy docs are stored in `main` branch and not in a separate branch
- `Edit on GitHub` link will not work if legacy docs are stored in `main` branch
- static images should be disabled? or content from `/public` directory of legacy docs should be copied

shuding:
@B2o5T Please don’t mind that I’m dumping all my thoughts over the years!

[...]

PR #1168, opened 2023-01-03
experimental: Dynamic meta data

shuding:
With this change, we can finally implement versioning and remote meta + content. And all these can be flexible enough to use SSR, SSG, or ISR.

An example of versioning:

```
/docs
  ... → current version
/v1
  [slug].mdx → use MDX remote + gSSP to build these from external APIs
  _meta.js → use an async function to generate the meta structure
```

So it will have the full sidebar, ISR and everything. I'm very excited about this change.

To-dos:

- [ ] Middleware should not add `.locale` to dynamic routes (need a loader to inject page map?)
- [ ] Add caching logic to `getDynamicMeta` to avoid over-fetching
- [x] Implement a loader for _meta to make sure it doesn't error as a page
- [x] Merge #1144 first (logic of `__nextra_pageMap` can be moved into the new hook)
- [x] Rewrite _meta routes to 404
- [ ] Support catch-all routes `[...slug].mdx`
- [ ] Allow _meta keys to contain `/` so it can apply to catch-all dynamic routes
- [x] `getDynamicMeta` need to also provide the MDX title. It should be either parsed from the MDX or provided by the user inside `getStaticProps`.
- [x] Similarly, it needs to parse and provide headings too.
- [x] The `pageMap` object seems to be out of order. Need to investigate as I refactored a bunch of things related.
- [x] We might want to include MDX remote in Nextra, and provide an out of the box experience.
- [x] Fix hydration error caused by `MDXProvider` and `useMDXComponents` coming from different instances.
- [x] Shiki errors in lambdas
- [x] The footer navigation should not across doc areas

Related:
#20, #852


packages/nextra/src/ssg.ts
```diff
@@ -7,3 +7,12 @@ export const useSSG = (key = 'ssg') => useContext(SSGContext)?.[key]
 
 export const DataContext = SSGContext
 export const useData = useSSG
+
+export const getDynamicMeta = async () => {
+  if ((globalThis as any).__nextra_resolvePageMap__) {
+    return {
+      __nextra_pageMap: await (globalThis as any).__nextra_resolvePageMap__()
```
B2o5T, line 14:
```suggestion
  const resolvePageMap = (globalThis as any).__nextra_resolvePageMap__
  if (resolvePageMap) {
    return {
      __nextra_pageMap: await resolvePageMap()
```

packages/nextra/src/plugin.ts
```diff
@@ -86,36 +98,75 @@ export async function collectFiles(
         fileMap[fp] = {
           kind: 'Meta',
           ...(locale && { locale }),
-          data: parseJsonFile(content, fp)
+          data: JSON.parse(content)
```
B2o5T, line 101: here `parseJsonFile` is used to avoid throwing an error in case of incorrect JSON file

packages/nextra/src/plugin.ts
```diff
@@ -86,36 +98,75 @@ export async function collectFiles(
...
+            path.relative(CWD, filePath)
+          )
+        }
+        return fileMap[fp]
+      } else if (fileName === 'meta.json') {
         console.warn(
           '[nextra] "meta.json" was renamed to "_meta.json". Rename the following file:',
           path.relative(CWD, filePath)
         )
+      } else if (/_meta\.(jsx|ts|tsx)$/.test(fileName)) {
```
B2o5T, line 137:
```suggestion
      } else if (/^_meta\.(jsx|ts|tsx)$/.test(fileName)) {
```

shuding:
OK so the next blocker we have is that Shiki doesn't work during SSR when deployed:

CleanShot-2023-01-04-czbrBxu8@2x

(It works during build time though, but not inside a serverless function)

shuding:
I think other than that, it's almost working perfectly on local. A couple of things I noticed:
- `getDynamicMeta` need to also provide the MDX title. It should be either parsed from the MDX or provided by the user inside `getStaticProps`.
- The `pageMap` object seems to be out of order. Need to investigate as I refactored a bunch of things related.
- We might want to include MDX remote in Nextra, and provide an out of the box experience. But I'm not sure. This can be done later.

shuding:
How it looks like locally (you might need to provide an `GITHUB_TEST_REMOTE_MDX` env to bypass rate limiting):

https://user-images.githubusercontent.com/3676859/210462972-b3d54f03-fc3b-4ea0-a5be-ea4fc4c98813.mp4

Since it's all gSSP-based, you have ISR and `fallback` available!

B2o5T:
> How it looks like locally (you might need to provide an `GITHUB_TEST_REMOTE_MDX` env to bypass rate limiting):
> 
>  CleanShot-2023-01-04-l07Y7aO3.mp4 
> Since it's all gSSP-based, you have ISR and `fallback` available!

Wow, amazing progress 😍🎉

packages/nextra-theme-docs/tsup.config.ts
```diff
@@ -6,6 +6,7 @@ export default defineConfig({
   entry: ['src/index.tsx'],
   format: 'esm',
   dts: true,
+  external: ['nextra'],
```
shuding, line 9: @promer94 this fixes #1031 :D

shuding:
I see why there's the Shiki problem. This dynamic require assumes that all languages are located in the corresponding directory: https://github.com/shikijs/shiki/blob/c7da91c857ed9f86b262e259215e46d95381aba7/packages/shiki/src/loader.ts#L117-L118 but when deployed to a serverless environment, however Next.js' [OFT](https://nextjs.org/docs/advanced-features/output-file-tracing) can't collect them fully at build time.

promer94:
https://github.com/vercel/next.js/pull/44605

shuding:
🥲 we can't use that for older Next.js versions sadly

packages/nextra-theme-docs/tsup.config.ts
promer94: I think we should nextra as theme's `peerDependencies`

shuding:
Got obvious bugs fixed. I think we can land this as an undocumented unstable feature - it doesn't work especially with i18n and catch-all routes!

shuding:
BTW I think our example/swr-site has too many pages, which makes build very slow. Maybe we remove some locales but only keep like 2 for testing?

B2o5T:
@shuding let's keep 1 locale for ltr and 1 for rtl

PR #1178, opened 2023-01-03
Simplify the theme layout interface

shuding:
After #1144, I realized that we can simplify the theme layout component even further and get rid of the `usePageContext()` call (so it can be a 100% internal implementation detail).

With this change, the theme layout is just a component that receives the following props:
- `children`: The page content
- `pageOpts`: Related meta information for the current page
- `themeConfig`: The global theme configuration

Which is very straightforward and easy to understand. This will make custom theme easier to create too.

Another reason for this change is, in https://github.com/shuding/nextra/pull/1168/files#diff-8310a54b1ec9da38ed7c74e8c1328e33f07848d25bd9779811614642e3268747R247-R249 we are manipulating the page's props (received from SSG/SSR) and the `pageOpts` object. It would be great to move all these to an internal layer as we can have more changes like that in the future.

### DIFF
--- a/packages/nextra/src/layout.ts
+++ b/packages/nextra/src/layout.ts
@@ -1,0 +1,21 @@
+import React from 'react'
+
+import { SSGContext } from './ssg'
+import { useInternals } from './use-internals'
+
+export default function Nextra(props: any): React.ReactElement {
+  const { context, Layout } = useInternals()
+  const { Content } = context
+
+  return React.createElement(
+    Layout,
+    context,
+    React.createElement(
+      SSGContext.Provider,
+      {
+        value: props
+      },
+      React.createElement(Content, props)
+    )
+  )
+}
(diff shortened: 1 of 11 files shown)


packages/nextra/src/layout.ts
```diff
@@ -0,0 +1,21 @@
...
+    Layout,
+    restContext,
+    React.createElement(
+      SSGContext.Provider,
+      {
+        value: props
+      },
+      React.createElement(Content, props)
+    )
+  )
```
B2o5T, line 20: any reason to not use JSX markup here?

B2o5T (approved):

packages/nextra/src/layout.ts
shuding: We don't have JSX enabled for the `nextra` package so I just went with this :D

PR #1184, opened 2023-01-04
fix(theme-blog): `•` separator is missing if tags are empty

numToStr:
This PR fixes the missing `•` separator in meta heading when `readingTime` is enabled and tags are empty.

- Before
image

- After
image

---

Also wrapped the separator in `<span className="nx-px-1">` to prevent uneven spacing when using monospace font like Fira Code.

- Before
image

- After
image


shuding:
Thanks!

PR #1185, opened 2023-01-04
feat(theme-blog): allow changing prefix in `<TagTitle />`

numToStr:
This makes `<TagTitle/>` little more customizable and allows user to change title how they want. Let me know if you have any other ways to solve this, I would really appreciate it.

---

Also, currently `TagTitle` doesn't respect `config.titleSuffix`. I tried to fix it using `useBlogContext()` but I was running into some weird error whenever I tried. So, do you have any idea how we can fix this?


shuding:
> Also, currently TagTitle doesn't respect config.titleSuffix

I think the blog theme isn't well documented yet, we need to work on a full list and finalize them a bit

Issue #1154, opened 2022-12-31
LCH


shuding:
We currently use HSL to theme, but ideally we want to switch to LCH and eventually, OKLCH.

The blocker right now is we'll have to use a polyfill for that. But we use CSS variables to provide the hue value which makes it very tricky to polyfill.

PR #1189, opened 2023-01-05
Oklch

B2o5T:
fixes https://github.com/shuding/nextra/issues/1154

Unfortunately, there is no way to polyfill oklch functions that include CSS variables, like following code, so we still should use hue for theme color
image


package.json
```diff
@@ -45,11 +45,10 @@
       "@types/react-dom": "^18.0.6",
       "postcss": "^8.4.14",
       "postcss-cli": "^8.3.1",
-      "autoprefixer": "^10.4.8",
-      "cssnano": "^5.1.12",
```
B2o5T, line 49: we no longer use `cssnano`

packages/nextra-theme-blog/package.json
```diff
@@ -69,17 +68,16 @@
   },
   "devDependencies": {
     "@tailwindcss/nesting": "^0.0.0-insiders.565cd3e",
-    "@tailwindcss/typography": "^0.5.7",
+    "@tailwindcss/typography": "^0.5.8",
     "@types/react": "^18.0.15",
     "@types/react-dom": "^18.0.6",
-    "autoprefixer": "^10.4.8",
```
B2o5T, line 75: no need to use `autoprefixer` since we use `lightningcss`

packages/nextra-theme-docs/tailwind.config.js
```diff
@@ -63,7 +64,6 @@ module.exports = {
         750: makePrimaryColor(35),
         800: makePrimaryColor(32),
         900: makePrimaryColor(24),
-        1000: makePrimaryColor(12)
```
B2o5T, line 66: tailwind palet includes dark suffix from 50 to 900

packages/nextra-theme-docs/package.json
```diff
@@ -65,20 +64,18 @@
   },
   "devDependencies": {
     "@tailwindcss/nesting": "^0.0.0-insiders.565cd3e",
-    "@tailwindcss/typography": "^0.5.4",
```
B2o5T, line 68: We don’t use it in docs

shuding:
LGTM, although the main reason I want to use OKLCH is for theming. Looking forward to a day that we can use CSS variables with OKLCH :(

B2o5T:
> LGTM, although the main reason I want to use OKLCH is for theming.

tried a lot, but unfortunately without success 🥲

B2o5T:
> Looking forward to a day that we can use CSS variables with OKLCH :(

@shuding to avoid misunderstanding. Note: we can use CSS variables with `oklch`, but we can't polyfill these properties that includes CSS variables, so bg color of color block becomes unstyled in Chrome (since it doesn't support oklch), but in Safari it's works

PR #1201, opened 2023-01-07
feat: add custom component configuration

timmywheels:
This adds functionality to allow for overriding the default MDX component configuration which makes the theme design more flexible.

As you can see, in `examples/blog/theme.config.jsx`, the `h1` element is overridden using a custom implementation.

#### Screenshot of example overriden H1 style:
image


B2o5T (approved):
Nice feat, we have the same in docs theme

PR #1206, opened 2023-01-08
Add docs for custom themes; tweak styling; fix Zod config checks

shuding:


packages/nextra-theme-docs/css/styles.css
```diff
@@ -34,6 +34,19 @@ summary {
...
+
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    @apply nx-text-slate-900 dark:nx-text-slate-100;
+  }
+}
```
shuding, line 48: I think our current content color has too much contrast (#000/#fff), adjusting them slightly will make a huge difference. Many well-designed docs do that.

packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -141,10 +141,12 @@ export const themeSchema = z
...
-const publicThemeSchema = themeSchema.deepPartial().extend({
-  // to have `locale` and `text` as required properties
-  i18n: i18nSchema
-})
+const publicThemeSchema = themeSchema
+  .extend({
+    // to have `locale` and `text` as required properties
+    i18n: i18nSchema
+  })
+  .deepPartial()
```
B2o5T, line 149: previously if `i18n` was set it requires `locale` and `text` to be set, now both are optional
✅ valid
```js
i18n: [{ locale: "en-US", text: "English" }]
```

❌ invalid
```js
i18n: [{ locale: "en-US" }]
```

packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -296,7 +298,8 @@ export const DEFAULT_THEME: DocsThemeConfig = {
   toc: {
     component: TOC,
     float: true,
-    title: 'On This Page'
+    title: 'On This Page',
+    extraContent: <React.Fragment />
```
B2o5T, line 302:
```suggestion
    title: 'On This Page'
```
this is redundant

packages/nextra-theme-docs/css/styles.css
```diff
@@ -34,6 +34,19 @@ summary {
...
+.nextra-content {
+  @apply nx-text-slate-700 dark:nx-text-slate-200;
+
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    @apply nx-text-slate-900 dark:nx-text-slate-100;
```
B2o5T, line 46: you can add them directly in `mdx-components.tsx`
here
https://github.com/shuding/nextra/blob/0cb1f35999b75bda5406e6212e4a40a5e9111cc0/packages/nextra-theme-docs/src/mdx-components.tsx#L234
and here
https://github.com/shuding/nextra/blob/0cb1f35999b75bda5406e6212e4a40a5e9111cc0/packages/nextra-theme-docs/src/mdx-components.tsx#L110

docs/package.json
```diff
@@ -13,13 +13,14 @@
     "clsx": "^1.2.1",
     "framer-motion": "^8.0.0",
     "next": "^13.1.1",
-    "nextra": "2.0.3",
-    "nextra-theme-docs": "2.0.3",
+    "nextra": "workspace:*",
+    "nextra-theme-docs": "workspace:*",
```
B2o5T, line 17: this is a good improvement to add docs in the workspace ❤️ wanted to propose it too

packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -126,7 +126,7 @@ export const themeSchema = z
       labels: z.string()
     }),
     sidebar: z.object({
-      defaultMenuCollapseLevel: z.number().min(2).int(),
+      defaultMenuCollapseLevel: z.number().min(0).int(),
```
B2o5T, line 129:
```suggestion
      defaultMenuCollapseLevel: z.number().min(1).int(),
```
this was my mistake 🙈, but it should be 1 (because behaviour is the same for 1 and 0 - collapse all)

packages/nextra-theme-docs/src/constants.tsx
shuding: Ohh! I was running into an error without this change for docs/, it seems that the `i18n` field is required

packages/nextra-theme-docs/src/constants.tsx
shuding: Running into an error too for the docs theme

packages/nextra-theme-docs/src/constants.tsx
B2o5T: Ah, will take a look today at why this happened

B2o5T: here `.optional()` was missing - `i18n: i18nSchema.optional()`

packages/nextra-theme-docs/src/constants.tsx
B2o5T: found why we didn't get type errors in this case:

the problem was in mismatching `z.custom` validation function and `ReactNode` type, `ReactNode` accept a value to be `null` or `undefined` but our custom `isReactNode` didn't allowed it, so I adjusted the function by checking `value == null`

I pushed a fix in this branch

packages/nextra-theme-docs/css/styles.css
B2o5T: moved to `mdx-components.tsx`

packages/nextra-theme-docs/src/constants.tsx
B2o5T: set to `1`

B2o5T (approved):
I will merge it even `docs` deploy is failed, seems something is broken due to recent versioning changes.

image

PR #1214, opened 2023-01-08
fix: adjust padding when banner is visible

### LINKED ISSUE #1183: Nextra theme docs wrong when show banner
At Nextra theme docs, I found a bug when using the banner top menu

This error causes the search box to be lost

Before I turn it off:

image

After I turn it off:

image

mocherfaoui:
closes #1183


B2o5T (changes requested):
thanks for trying to fix it, but now there is an extra space when we scroll and open the menu :(
https://user-images.githubusercontent.com/7361780/211222661-4d2363a5-b00a-4140-a36a-9518da019e99.mov

mocherfaoui:
@B2o5T 
oops, I totally missed that!
do you recommend using something like intersection observer to know if the banner is in view? or would that be too much?

B2o5T:
> @B2o5T oops, I totally missed that! do you recommend using something like intersection observer to know if the banner is in view? or would that be too much?

I think it is too much for this simple case :)

Maybe we should add an `onClick` handler for the menu bottom to scroll a little bit only in case if banner exists and we are at the top of the viewport (`window.scrollY === 0`)

cc @shuding any thoughts ?

mocherfaoui:
@B2o5T 

I went with option#2, let me know what you think

B2o5T (changes requested):
now banner styles on rtl direction are broken (you can test it by switching to Spanish language)

before:
image

now:
image

shuding:
This seems like a tricky one, I think option 2) sounds good to me

fang-w-shen:
nice!

> @B2o5T
> 
> I went with option#2, let me know what you think

dszeto:
I just run into the same issue. Thanks for taking a shot at fixing this!

B2o5T (approved):

PR #1228, opened 2023-01-10
support `getStaticProps` from theme config

B2o5T:
`getStaticProps` from the MDX page take priority over `getStaticProps` from the theme config

fixes https://github.com/shuding/nextra/issues/1227

### DIFF
--- a/packages/nextra/src/loader.ts
+++ b/packages/nextra/src/loader.ts
@@ -157,7 +157,7 @@ async function loader(
 
   const katexCssImport = latex ? "import 'katex/dist/katex.min.css'" : ''
   const cssImport = OFFICIAL_THEMES.includes(
-    theme as typeof OFFICIAL_THEMES[number]
+    theme as (typeof OFFICIAL_THEMES)[number]
   )
     ? `import '${theme}/style.css'`
     : ''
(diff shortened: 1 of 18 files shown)


shuding:
Gonna improve this part a bit:

```js
  // `getStaticProps` from the MDX page take priority over `getStaticProps` from the theme config
  result.includes('getStaticProps')
```

shuding:
I thought again about this, two major concerns I have:
- What should we do if the page itself has `getStaticProps`/`getServerSideProps`/`getStaticPaths`?
- Since we attach the same gSSP to all pages, we can't know which page is being rendered inside gSSP.

I have another similar proposal that can probably avoid those.

B2o5T:
> Gonna improve this part a bit:
> 
> ```js
>   // `getStaticProps` from the MDX page take priority over `getStaticProps` from the theme config
>   result.includes('getStaticProps')
> ```

Since getStaticProps can be expected differently 

`export function getStaticProps`
`export const getStaticProps`
`export { getStaticProps } from './foo'`

The better way will be to convert the `result` into AST and parse it to find getStaticProps, but I guess it’s overkill

shuding:
I pushed an commit to implement a `transform` method which can be very extendable

packages/nextra/src/loader.ts
```diff
@@ -134,6 +135,15 @@ async function loader(
   // so we can generate the correct page map.
   context.addContextDependency(PAGES_DIR)
 
+  // Add local theme as a dependency
+  if (theme.startsWith('.') || theme.startsWith('/')) {
+    context.addDependency(path.resolve(theme))
+  }
+  // Add theme config as a dependency
+  if (themeConfig) {
+    context.addDependency(path.resolve(themeConfig))
```
B2o5T, line 144: This is a nice improvement

PR #1230, opened 2023-01-10
Allow to use MD (instead of MDX) syntax

shuding:
Sometimes people want to only use standard Markdown files to build a simple page, if they don't need JSX at all. One difference between MD and MDX is that `<!-- comment -->` will be invalid in MDX.

Hence we are now accepting the `format` option for `mdxOptions`. It can be defined via both `next.config.js` globally or front matter on a per file basis.

### DIFF
--- a/packages/nextra/src/compile.ts
+++ b/packages/nextra/src/compile.ts
@@ -92,10 +92,14 @@ export async function compileMdx(
     ]
   })
   try {
-    const vFile = await compiler.process({
-      value: source,
-      path: filePath
-    })
+    const vFile = await compiler.process(
+      filePath
+        ? {
+            value: source,
+            path: filePath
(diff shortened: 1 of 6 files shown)


packages/nextra/src/compile.ts
```diff
@@ -91,7 +92,14 @@ export async function compileMdx(
...
   try {
-    const vFile = await compiler.process(source)
+    const vFile = await compiler.process(
+      filePath
+        ? {
+            value: source,
+            path: filePath
+          }
+        : source
+    )
```
shuding, line 102: According to https://github.com/orgs/mdx-js/discussions/2219#discussioncomment-4649484 it's good to also include the filepath for plugins (if they need it).

PR #1241, opened 2023-01-11
fix: align footer text to center

lotyp:


lotyp:
Will look like this, after changes take effect:

image

B2o5T (changes requested):
we don't need `nx-` prefixes in docs, it's only for encapsulating `nextra-theme-docs` and `nextra-theme-blog` styles, you should have lint errors

lotyp:
@B2o5T yes, I understand, but looks like without them, style of footer is broken, or align to left side is intended? It looks broken for me

PR #1243, opened 2023-01-11
Allow disabling code highlighting & memory improvements

shuding:


packages/nextra/src/compile.ts
```diff
@@ -52,45 +54,57 @@ export async function compileMdx(
...
     | 'readingTime'
     | 'latex'
+    | 'codeHighlight'
   > & {
     mdxOptions?: LoaderOptions['mdxOptions'] &
       Pick<ProcessorOptions, 'jsx' | 'outputFormat' | 'format'>
   } = {},
-  filePath = ''
+  filePath = '',
+  useCachedCompiler = false
```
shuding, line 63: For dev and prod builds, when called from the loader all options here will not change. Hence we can just reuse the compiler.

packages/nextra/src/index.js
```diff
@@ -41,7 +41,10 @@ const nextra = (...config) =>
       ],
       webpack(config, options) {
         config.plugins ||= []
-        config.plugins.push(nextraPlugin)
+
+        if (options.nextRuntime !== 'edge' && options.isServer) {
+          config.plugins.push(nextraPlugin)
+        }
```
shuding, line 47: Only need to be included in one compiler (Next.js has 3).

packages/nextra/src/loader.ts
```diff
@@ -126,22 +126,24 @@ async function loader(
...
-    context.addDependency(path.resolve(themeConfig))
+    // Add local theme as a dependency
+    if (theme.startsWith('.') || theme.startsWith('/')) {
+      context.addDependency(path.resolve(theme))
+    }
+    // Add theme config as a dependency
+    if (themeConfig) {
+      context.addDependency(path.resolve(themeConfig))
+    }
   }
```
shuding, line 147: No need to add dependencies for prod builds.

packages/nextra/src/compile.ts
```diff
@@ -18,11 +18,8 @@ import { truthy } from './utils'
 import remarkMath from 'remark-math'
 import rehypeKatex from 'rehype-katex'
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
 globalThis.__nextra_temp_do_not_use = () => {
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
+  // @ts-expect-error -- ignore error - File is not a module
```
B2o5T, line 22: to avoid adding `eslint-disable` comment, just use `@ts-expect-error` instead of `@ts-ignore` and describe why the error occurs, this will be less annoying

packages/nextra/src/compile.ts
```diff
@@ -77,45 +81,50 @@ export async function compileMdx(
...
-    outputFormat: mdxOptions.outputFormat || 'function-body',
-    providerImportSource: 'nextra/mdx',
-    format: mdxOptions.format || 'mdx',
-    // https://github.com/hashicorp/next-mdx-remote/issues/307#issuecomment-1363415249
-    development: false,
-    remarkPlugins: [
-      ...(mdxOptions.remarkPlugins || []),
-      remarkGfm,
-      remarkHeadings,
-      loaderOptions.staticImage && ([remarkStaticImage, { filePath }] as any),
```
B2o5T, line 94: since filePath will be cached, I removed this parameter from `remarkStaticImage `

packages/nextra/src/index.js
```diff
@@ -56,6 +52,15 @@ const nextra = (...config) =>
...
     }
 
+    const nextraLoaderOptions = {
+      ...nextraConfig,
+      locales: nextConfig.i18n?.locales || [''],
+      defaultLocale: nextConfig.i18n?.defaultLocale || DEFAULT_LOCALE,
+      pageMapCache,
+      newNextLinkBehavior: nextConfig.experimental?.newNextLinkBehavior,
+      distDir: nextConfig.distDir
+    }
```
B2o5T, line 62: we can put this object outside of `webpack` function

packages/nextra/src/compile.ts
```diff
@@ -61,11 +63,13 @@ export async function compileMdx(
     | 'defaultShowCopyCode'
     | 'readingTime'
     | 'latex'
+    | 'codeHighlight'
   > & {
     mdxOptions?: LoaderOptions['mdxOptions'] &
-      Pick<ProcessorOptions, 'jsx' | 'outputFormat' | 'format'>
+      Pick<ProcessorOptions, 'jsx' | 'outputFormat'>
```
B2o5T, line 69: I removed `format` from here because we already specified it in `mdxOptions.format` and there are conflicts

B2o5T (approved):

PR #1253, opened 2023-01-13
fix hydration error, broken code-blocks styles with `nextraConfig.codeHighlight: false` setting, refactor

B2o5T:


packages/nextra/src/compile.ts
```diff
@@ -23,15 +23,7 @@ globalThis.__nextra_temp_do_not_use = () => {
   import('./__temp__')
 }
 
-const createCompiler = (mdxOptions: ProcessorOptions): Processor => {
-  const compiler = createProcessor(mdxOptions)
-  compiler.data('headingMeta', {
-    headings: []
-  })
-  return compiler
-}
```
B2o5T, line 32: since now the compiler is cached it produces hydration error on the client because `headingMeta` are appended on each navigation
image

packages/nextra/src/compile.ts
```diff
@@ -124,24 +121,17 @@ export async function compileMdx(
...
 
     const headingMeta = compiler.data('headingMeta') as Pick<
       PageOpts,
       'headings' | 'hasJsxInH1' | 'title'
     >
-    if (headingMeta.title) {
-      result += `\nconst __nextra_title__ = ${JSON.stringify(
-        headingMeta.title
-      )}`
-    }
```
B2o5T, line 139: since we no longer use the previous plugin, this code is no longer needed we can get `title` (`headingMeta.title`) in loader

packages/nextra/src/components/pre.tsx
```diff
@@ -33,36 +25,26 @@ export const Pre = ({
     }
   }, [])
 
-  useEffect(() => {
-    const codeEl = preRef.current?.querySelector('code')
-
-    if (codeEl?.textContent) {
-      setCodeString(codeEl.textContent)
-    }
-  }, [preRef])
```
B2o5T, line 42: we don't need this anymore since this component was refactored by @shuding to take `getValue` fn and not `value` prop

packages/nextra/src/constants.ts
```diff
@@ -8,11 +9,12 @@ export const LOCALE_REGEX = /\.([a-z]{2}(-[A-Z]{2})?)$/
...
 export const DEFAULT_LOCALE = 'en-US'
 
-export const DEFAULT_CONFIG = {
+export const DEFAULT_CONFIG: Omit<NextraConfig, 'theme'> = {
   staticImage: true,
   flexsearch: {
     codeblocks: true
-  }
+  },
+  codeHighlight: true
```
B2o5T, line 17: it was missing, because currenly all code blocks are unstyled by default

packages/nextra/src/layout.tsx
```diff
@@ -28,13 +28,11 @@ export default function Nextra({
...
-      SSGContext.Provider,
-      { value: props },
-      createElement(Content, props)
-    )
+  return (
+    <Layout {...restContext} pageProps={props}>
+      <SSGContext.Provider value={props}>
+        <Content {...props} />
+      </SSGContext.Provider>
+    </Layout>
```
B2o5T, line 36: we can use jsx markup here

packages/nextra/src/loader.ts
```diff
@@ -127,10 +128,13 @@ async function loader(
   // mdx is imported but is outside the `pages` directory
   if (!fileMap[mdxPath]) {
     fileMap[mdxPath] = await collectMdx(mdxPath)
-    context.addMissingDependency(mdxPath)
+    if (!IS_PRODUCTION) {
+      context.addMissingDependency(mdxPath)
+    }
```
B2o5T, line 133: I guess it should be called only on dev

packages/nextra/src/loader.ts
```diff
@@ -172,15 +176,16 @@ async function loader(
       defaultShowCopyCode,
       staticImage,
       flexsearch,
-      latex
+      latex,
+      codeHighlight
```
B2o5T, line 180: `codeHighlight` was not passed, it `false` by default in our `compileMdx`

packages/nextra/src/loader.ts
```diff
@@ -172,15 +176,16 @@ async function loader(
...
     },
     mdxPath,
-    true
+    false // TODO: produce hydration errors or error - Create a new processor first, by calling it: use `processor()` instead of `processor`.
```
B2o5T, line 183: caching compiler is not ready for the moment, it produces hydration error on the client or an error
```bash
Create a new processor first, by calling it: use `processor()` instead of `processor`
```
while trying call `compiler.data('headingMeta', { headings: [] })` 2nd time

packages/nextra/src/loader.ts
```diff
@@ -278,15 +288,7 @@ export default MDXContent`
...
-          }) => `import('${slash(
-            (() => {
-              const relative = path.relative(
-                path.dirname(mdxPath),
-                metaFilePath
-              )
-              return relative.startsWith('.') ? relative : './' + relative
-            })()
-          )}').then(m => {
+          }) => `import('${convertToRelativePath(metaFilePath)}').then(m => {
```
B2o5T, line 291: extracted as function for better readability

packages/nextra/src/page-map.ts
```diff
@@ -19,39 +19,35 @@ function getDynamicMeta(
...
+      const { __nextra_src, ...newItem } = item
+      dynamicMetaItems.push({
+        metaFilePath: __nextra_src,
+        metaObjectKeyPath: `${path}[${i}]`,
+        metaParentKeyPath: path
+      })
+      newItems.push(newItem)
+      continue
+    }
+    newItems.push(item)
```
B2o5T, line 50: this was refactored for better readability too

packages/nextra/src/utils.ts
```diff
@@ -18,21 +18,6 @@ export function parseFileName(filePath: string): {
...
-    return JSON.parse(content)
-  } catch (err) {
-    console.error(
-      `[nextra] Error parsing ${path}, make sure it's a valid JSON`,
-      err
-    )
-    return {}
-  }
-}
-
```
B2o5T, line 35: no longer used

packages/nextra/styles/code-block.css
```diff
@@ -33,6 +33,10 @@ pre {
     }
   }
 
+  &:not([data-theme]) {
+    @apply nx-px-4;
+  }
```
B2o5T, line 38: fix broken styles with `codeHighlight: false`
image

B2o5T: also, I made this PR https://github.com/shuding/nextra/pull/952 works here because `codeHighlight: false` doesn't wrap `<pre/>` with `<div data-rehype-pretty-code-fragment />`

shuding (approved):
Looks good!

PR #1268, opened 2023-01-15
loader refactor, type-safe `__nextra_resolvePageMap`, avoid code interpolation in loader.ts

B2o5T:
I didn't think about this refactoring before, but now it looks much much better + everything is typesafe, also @shuding this should help you finish versioning

I tested remote and hot refresh - both works 🎉


packages/nextra/src/setup-page.ts
```diff
@@ -0,0 +1,116 @@
...
+}) {
+  if (typeof window === 'undefined') {
+    globalThis.__nextra_resolvePageMap = async () => {
+      const clonedPageMap = JSON.parse(JSON.stringify(pageOpts.pageMap))
+      await Promise.all(
+        dynamicMetaItems.map(
+          async ({ metaFilePath, metaObjectKeyPath, metaParentKeyPath }) => {
+            const mod = await import(metaFilePath)
+            const metaData = await mod.default()
+            const meta = get(clonedPageMap, metaObjectKeyPath)
```
B2o5T, line 62: @shuding if you'll refactor this code https://github.com/shuding/nextra/blob/5f5e06bb8257cc4a3fa533cd03927f855aaba222/packages/nextra/src/page-map.ts#L29 we can remove `lodash.get`

packages/nextra/src/setup-page.ts
```diff
@@ -0,0 +1,116 @@
...
+  for (let i = 0; i < str.length; i++) {
+    hval ^= str.charCodeAt(i)
+    hval += (hval << 1) + (hval << 4) + (hval << 7) + (hval << 8) + (hval << 24)
+  }
+
+  // Convert to 8 digit hex string
+  return ('0000000' + (hval >>> 0).toString(16)).substring(-8)
+}
+
+export function setupNextraPage({
```
B2o5T, line 37: we can choose a better name

packages/nextra/src/remote.ts
```diff
@@ -14,7 +14,7 @@ export const buildDynamicMDX = async (content: string) => {
 }
 
 export const buildDynamicMeta = async () => {
-  const resolvePageMap = (globalThis as any).__nextra_resolvePageMap__
+  const resolvePageMap = (globalThis as any).__nextra_resolvePageMap
```
B2o5T, line 17: adding a double underscore at the end is redundant

packages/nextra/src/loader.ts
```diff
@@ -242,8 +221,8 @@ export default MDXContent`
   const layout = isLocalTheme ? path.resolve(theme) : theme
 
   const themeConfigImport = themeConfig
-    ? `import __nextra_themeConfig__ from '${slash(path.resolve(themeConfig))}'`
-    : ''
+    ? `import __nextra_themeConfig from '${slash(path.resolve(themeConfig))}'`
+    : 'const __nextra_themeConfig = null'
```
B2o5T, line 225: to avoid error - variable not defined :)

packages/nextra/package.json
```diff
@@ -112,10 +116,12 @@
   "devDependencies": {
     "@types/github-slugger": "^1.3.0",
     "@types/graceful-fs": "^4.1.5",
+    "@types/lodash.get": "^4.4.7",
     "@types/mdast": "^3.0.10",
     "@types/react": "^18.0.15",
     "@types/react-dom": "^18.0.6",
     "@types/webpack": "^5.28.0",
+    "@types/webpack-env": "^1.18.0",
```
B2o5T, line 124: to have `module.hot` types

packages/nextra/src/setup-page.ts
shuding: Thanks, love putting loader content inside another file, makes things super clean and typed! One thing I'm not sure is all the content here (`lodash.get`, `hashFnv32a`) will be part of the page bundle that make the page larger. They should be done at build time I think. Maybe `hashFnv32a(JSON.stringify(pageOpts))` can be passed inline?

```js
setupNextraPage({
  pageOptsHash: "${pageOptsChecksum}"
})
```

Another thing is probably Webpack's dead code elimination problem, using `if (!IS_PRODUCTION && hot)` and `if (process.env.NODE_ENV !== 'production' && hot)` could be different. The former needs to actually run the code to tell if that condition will be executed or not, but the latter can be determined when building the module via static code analyzation so that code will be removed completely after build.

shuding (commented):
🤩 Great improvement of life, thanks!

packages/nextra/src/setup-page.ts
B2o5T: you are right, after adding `process.env.NODE_ENV !== 'production'` or `process.env.NODE_ENV === 'development'` `.next/static/chunks` folder doesn't include `module.hot` code, and `hashFnv32a` is automatically tree-shaked, and about `lodash.get` let's keep for the moment

B2o5T: before/after
image

package.json
```diff
@@ -1,7 +1,6 @@
 {
   "private": true,
   "scripts": {
-    "dev": "turbo run dev",
```
B2o5T, line 4: this should be never used

B2o5T:
@shuding I applied your suggestions, I will merge this one because I wanna push another pr with remote `[...catchAll]` support

packages/nextra/src/loader.ts
```diff
@@ -269,102 +248,27 @@ export default MDXContent`
...
-      ${dynamicMetaItems
-        .map(
-          ({
-            metaFilePath,
-            metaObjectKeyPath,
-            metaParentKeyPath
-          }) => `import('${convertToRelativePath(metaFilePath)}').then(m => {
-        const getMeta = Promise.resolve(m.default()).then(metaData => {
-          const meta = clonedPageMap${metaObjectKeyPath}
-          meta.data = metaData
```
shuding, line 294: This change causes a bunch of errors like the following:

```
swr-site:dev: warn  - ../../node_modules/.pnpm/file+packages+nextra_q76c2b4vyoegvsbrcwkfvimnai/node_modules/nextra/dist/setup-page.mjs
swr-site:dev: Critical dependency: the request of a dependency is an expression
```

Because previously all the imports are static and inlined as code, but now they're dynamic so webpack can't figure out the module graph without running the code.

shuding: I think I'll try to refactor it to make it both type safe and static. Although it's tricky due to the nature of webpack loaders.

PR #1286, opened 2023-01-17
remove `github-slugger` from docs

B2o5T:
since we attach heading id here https://github.com/shuding/nextra/blob/091b77b2fb947a58752ec676cb84e4d0fbd764c2/packages/nextra/src/mdx-plugins/rehype.js#L37-L38

everything should be consistent and we should not use directly `github-slugger` in docs

also after this change https://github.com/shuding/nextra/pull/553 would be easily to implement


shuding (approved):
Nice!

Issue #1239, opened 2023-01-11
Support for sidebar control via single root _meta file


KRRISH96:
I have been trying nextra recently and loving it. Thank you for building this :)

I feel a bit limited in terms of controlling the sidebar,
Currently, I can only structure the items in the sidebar for a particular directory and have to create `_meta` in the nested sub-directory to control respective items. Can easily get hectic as the nested level goes deep and there are many directories and sub-directories as well.

I do understand why managing _meta at a respective directory will make things easier when moving around. But I think a single `_meta` control should also be allowed. Similar to menus where we can add as children.

Similar to [menu with items](https://nextra.site/docs/docs-theme/page-configuration#menus)

```js
{
  "company": {
    "title": "Company",
    "type": "category" // <== maybe a new type as "category"?,
    "items": { // <== Like this
      "about": "About",
      "contact": "Contact Us",
      "policies": {
        "title": "Policies",
          "items": { // <== With Nested items too
            "terms": "Terms of Service",
            "legal": "Legal Policies"
          }
      }
    }
  }
}
```

Any suggestions/workarounds in achieving this would be really helpful. Thank you 🙌

KRRISH96:
Similar behavior. I'm wondering if there is a way to just have one `_meta.json` file at the root to customize (titles, order, etc.,)

Something like this from docusaurus (https://docusaurus.io/docs/sidebar/items#sidebar-item-category).

the use case is I want to have a single source of truth for my sidebar (`docs/_meta.json`) that has options to customize everything from root level to nested doc files and render the sidebar accordingly.

shuding:
In our initial design, _meta.json at each folder level is great when you move docs around. For example when moving an entire folder to another place, everything just works.

Another reason is we want some kind of "dynamic meta" in specific levels. You will see that in our upcoming updates.

shuding:
> But I think a single _meta control should also be allowed. Similar to menus where we can add as children.

Agreed! It's just that we need to be careful about conflicts etc.

KRRISH96:
> In our initial design, _meta.json at each folder level is great when you move docs around. For example when moving an entire folder to another place, everything just works.
> 
> Another reason is we want some kind of "dynamic meta" in specific levels. You will see that in our upcoming updates.


+1 on this approach as it's quite common that docs are moved around, especially with major version changes. However, root-level control should still be allowed IMHO.
Also, I did try the dynamic meta out and liked it 💯 .

KRRISH96:
[...]

PR #1294, opened 2023-01-18
add support for remote [...catchAll] routes 

B2o5T:
how look like catch all routes for `graphql-eslint` repository fetched from git

image

- [x] add support for remote `[...catchAll]` routes
- [x] sanitize remote mdx by removing `import` statements 
- [x] test `collectCatchAll` that adds to pageMap meta files and mdx pages 
- [x] fix hydration error
   for some reason remote mdx wrapped with `<p/>`
image

> was caused by extra semi 

image
- [x] be capable provide missing components .`<RemoteContent />` now accept `components` props
- [x] be capable provide `codeHighlight: true` and `defaultShowCopyCode: true` while compiling remote mdx
- [x] be capable auto-generate meta links for sidebar from dynamic `_meta` file with `createCatchAllMeta()`
- [x] be capable of simple reordering above 👆 
- [x] be capable override internal URLs e.g. `[hello](/docs/foo)` => `[hello](/docs/v2/foo)` in remote mdx content
- [x] folder renaming works for both folders with/without index page
- [x] better folder structure
```json
{
  "$YOUR_FOLDER_NAME": {
    "type": "folder",
    "title":  "$YOUR_NEW_FOLDER_TITLE",
    "items": {
      "$NESTED_PAGE": "$NESTED_PAGE_TITLE"
    }
  }
}
```  

- [ ] <s>support meta keys with `/` </s> ❗️this no longer needed

### DIFF
--- a/examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
+++ b/examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
@@ -10,7 +10,7 @@ module.exports = async () => {
   return createCatchAllMeta(
     // Ensure you didn't forget define some file by providing all paths of remote files
     files,
-    // Next you can override order of your meta files, folders should end with / and be nested
+    // Next you can override the order of your meta files, folders should have `type: 'folder'` and have `items` property
     {
       index: 'Introduction',
       'getting-started': {
--- a/examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
+++ b/examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
@@ -1,0 +1,25 @@
+/* eslint-env node */
+const { listFiles } = require('../../../components/remote-utils')
+const { createCatchAllMeta } = require('nextra/catch-all')
+
+module.exports = async () => {
+  const files = await listFiles({
+    repo: 'https://github.com/B2o5T/graphql-eslint',
+    rootDir: 'website/src/pages/docs/'
+  })
+  return createCatchAllMeta(
+    // Ensure you didn't forget define some file by providing all paths of remote files
+    files,
+    // Next you can override order of your meta files, folders should end with / and be nested
+    {
+      index: 'Introduction',
+      'getting-started': {
+        type: 'folder',
+        title: '🦄 GETTING Started',
+        items: {
+          parser: ''
+        }
+      }
+    }
+  )
+}
--- a/packages/nextra/src/compile.ts
+++ b/packages/nextra/src/compile.ts
@@ -4,7 +4,7 @@ import { createProcessor, ProcessorOptions } from '@mdx-js/mdx'
 import { Processor } from '@mdx-js/mdx/lib/core'
 import remarkGfm from 'remark-gfm'
 import rehypePrettyCode from 'rehype-pretty-code'
-import readingTime from 'remark-reading-time'
+import remarkReadingTime from 'remark-reading-time'
 import grayMatter from 'gray-matter'
 
 import {
@@ -13,7 +13,8 @@ import {
   remarkReplaceImports,
   structurize,
   parseMeta,
-  attachMeta
+  attachMeta,
+  remarkRemoveImports
 } from './mdx-plugins'
 import { LoaderOptions, PageOpts, ReadingTime } from './types'
 import theme from './theme.json'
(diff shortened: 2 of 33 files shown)


packages/nextra/src/compile.ts
```diff
@@ -83,29 +84,36 @@ export async function compileMdx(
...
+
   const compiler =
     (useCachedCompiler && cachedCompilerForFormat[format]) ||
     (cachedCompilerForFormat[format] = createProcessor({
       jsx,
       format,
       outputFormat,
       providerImportSource: 'nextra/mdx',
-      // https://github.com/hashicorp/next-mdx-remote/issues/307#issuecomment-1363415249
-      development: false,
```
B2o5T, line 94: seems like this is no longer needed on 4.2.1 version https://github.com/hashicorp/next-mdx-remote/issues/307#issuecomment-1377131951

examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
```diff
@@ -0,0 +1,39 @@
...
+  // TODO: return meta from metaMap
+  return {
+    "/": {
+      configs: "configs",
+      "custom-rules": "custom-rules",
+      "getting-started": "getting-started",
+      "/getting-started": {
+        "parser-options": "parser-options",
+        parser: "parser",
+      },
```
shuding, line 35: Have you considered the `items` proposal from #1239? Or supporting `/` in keys like:

```json
{
  "configs": "configs",
  "getting-started": "getting-started",
  "getting-started/parser-options": "parser-options",
  "getting-started/parser": "parser"
}
```

shuding (commented):
Amazing work! Left one thing for discussion

examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
B2o5T: I think we can support boths?

examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
```diff
@@ -0,0 +1,21 @@
...
+    // Next you can override order of your meta files, folders should end with / and be nested
+    {
+      index: 'Introduction',
+      'getting-started': 'Getting Started',
+      'getting-started/': {
+        parser: 'Parser'
+      }
+    }
+  )
+}
```
B2o5T, line 21: @shuding I still think that nested keys with slashes (I changed the order, now folders should end with slashes, like in `.gitignore`) are the most friendly way since less code is written.

With the above helper function, I was able to change the original order of items with a few lines coded

## from

image

## to
image

examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
shuding: I think they need to be merged right? It's still 1 item on the sidebar 🤔

B2o5T:
@shuding I found a way to remove slash from folder keys, also now folders renaming works for folders with **and** without the folder index page

## folder with an index page
image
image

## folder without index page

image
image

examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
B2o5T: they should not be merged so nextra under the hood will know if folder has index page or not

examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
```diff
@@ -0,0 +1,25 @@
...
+
+module.exports = async () => {
+  const files = await listFiles({
+    repo: 'https://github.com/B2o5T/graphql-eslint',
+    rootDir: 'website/src/pages/docs/'
+  })
+  return createCatchAllMeta(
+    // Ensure you didn't forget define some file by providing all paths of remote files
+    files,
+    // Next you can override order of your meta files, folders should end with / and be nested
```
B2o5T, line 13:
```suggestion
    // Next you can override the order of your meta files, folders should have `type: 'folder'` and have `items` property
```

examples/swr-site/pages/remote/graphql-eslint/_meta.en-US.js
shuding: The tricky thing was i18n. If users can choose `foo.en.mdx` or `foo/index.en.mdx` (they point to the same URL), and when accessing `/foo`, the middleware doesn't know if it's `foo.en` or `foo/index.en`. So in the past I documented it as you'll always have to avoid `index.en.mdx`.

But I think overall the middleware based i18n solution is too hacky and I'd like to fix it properly. I have an idea for that and I'll probably start working on it soon!

shuding:
Wow amazing result!! Code LGTM!

PR #1296, opened 2023-01-18
feat: add automatic format detection

### LINKED ISSUE #709: Skip `mdxCompiler` for regular `md` files
Hi, first of all thanks for the great framework, really enjoyed it!

## Issue
At the moment both `md` and `mdx` files are treated as MDX and get prepared by `compileMdx` (`@mdx-js`) process. This leads to 2 problems:

1) increased memory consumption. For ~2000 small md files ([sample file](https://gist.github.com/dj-stormtrooper/d223d4748acc8d0d43fa0287434509ae)) we need over 8GB for next build which make it impossible [to build on Vercel for example](https://vercel.com/docs/concepts/deployments/troubleshoot-a-build#cancelled-builds-due-to-limits)
2) we have a lot of black-listed words and symbols in md (e.g. `import`, `export`, `{`, `<`) which make it hard to develop or migrate documentation 

## Solution
Opt-out / opt-in option for skipping MDX processing for md files and use simple md-html convertion by`remark-html` instead

-- OR --

Workaround to get this to work. Perhaps custom webpack loader for different types of files (might be too abtract IMO) or any other way to exclude `md` files from MDX processing and apply simple md-html conversion without extracting a lot of internal implementation details of Nextra

Thanks!

### LINKED ISSUE #1295: Support format: "detect" for the MDX compiler
https://mdxjs.com/packages/mdx/#optionsformat

It should be already supported via #1230 but we're missing the type and documentation.

Closes #709

balazsorban44:
The file path needs to be checked since `createProcessor` cannot detect the format:

https://mdxjs.com/docs/troubleshooting-mdx/#incorrect-format-detect

closes #709, closes #1295


packages/nextra/src/compile.ts
```diff
@@ -82,6 +82,8 @@ export async function compileMdx(
     // You can override MDX options in the frontMatter too.
     ...(frontMatter.mdxOptions as Record<string, unknown>)
   }
+  
+  const format = _format = "detect" ? filePath.endsWith(".mdx") ? "mdx" : "md" : _format
```
B2o5T, line 86:
```suggestion
  const format = _format === "detect" ? filePath.endsWith(".mdx") ? "mdx" : "md" : _format
```

balazsorban44: 🤦‍♂️

B2o5T (approved):

PR #1298, opened 2023-01-18
feat: Add `search.useResultModifier` support

shuding:
Although we can improve our search algorithm, it might still be helpful to make it customizable. Keeping this a draft as I'm not convinced yet...

It's interesting that due to our multi-index implementation and the natural of doc sites, we can just collect the full content and then do search on either a server or on the client completely (will make it configurable in the future, or automatically). I don't see a bottleneck in the near future (1mb gzipped index is a lot, probably can fit 5,000~10,000 pages). And if that becomes the bottleneck some day, we can find a way to convert them into vectors instead.

Ref: list of libraries that we can take a look:
- https://github.com/lithammer/fuzzysearch
- https://github.com/facebookresearch/faiss 
- https://github.com/leeoniya/uFuzzy
- https://github.com/nextapps-de/flexsearch
- https://github.com/EthanRutherford/fast-fuzzy
- https://github.com/LyraSearch/lyra


dimaMachina:
Closed by https://github.com/shuding/nextra/pull/2569

PR #1301, opened 2023-01-18
fix: Warning of dynamic imports

shuding:
See detailed explanation here: https://github.com/shuding/nextra/pull/1268/files#r1080669178

### DIFF
--- a/packages/nextra/src/loader.ts
+++ b/packages/nextra/src/loader.ts
@@ -4,7 +4,7 @@ import type { LoaderContext } from 'webpack'
 import path from 'node:path'
 import slash from 'slash'
 
-import { pageTitleFromFilename, parseFileName } from './utils'
+import { hashFnv32a, pageTitleFromFilename, parseFileName } from './utils'
 import { compileMdx } from './compile'
 import { resolvePageMap } from './page-map'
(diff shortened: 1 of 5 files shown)


B2o5T:
oops, thanks for the explanation!

packages/nextra/src/loader.ts
```diff
@@ -256,12 +258,20 @@ ${finalResult}
...
 setupNextraPage({
   pageNextRoute: ${JSON.stringify(pageNextRoute)},
-  pageOpts: ${JSON.stringify(pageOpts)},
+  pageOpts: ${stringifiedPageOpts},
   nextraLayout: __nextra_layout,
   themeConfig: ${themeConfigImport ? '__nextra_themeConfig' : 'null'},
   Content: MDXContent,
   hot: module.hot,
-  dynamicMetaItems: ${JSON.stringify(dynamicMetaItems)}
+  pageOptsChecksum: ${JSON.stringify(hashFnv32a(stringifiedPageOpts))},
```
B2o5T, line 266: anyway, `module.hot` code is no longer included and tree shaked due `process.env.NODE_ENV !== 'production'` check

PR #1310, opened 2023-01-19
chore: Fix dev errors

shuding:
Errors like `swr-site:dev: Module not found: Can't resolve './chunk-HIDP27A7.mjs'` are extremely annoying, had to run `pnpm i` every time to copy over all the files. Not sure which part is wrong but I went with the way of transpiling everything in Nextra and disabling `splitting`.

### DIFF
--- a/.changeset/tender-plants-greet.md
+++ b/.changeset/tender-plants-greet.md
@@ -1,0 +1,5 @@
+---
+'nextra': patch
+---
+
+disable code splitting and fix dev errors
--- a/packages/nextra/src/__temp__.js
+++ b/packages/nextra/src/__temp__.js
@@ -1,5 +1,3 @@
-/* eslint-disable @typescript-eslint/no-var-requires, @typescript-eslint/ban-ts-comment */
-
 // This file has to be targeted to CJS, to keep `require.resolve` untranspiled.
 // Otherwise, the file tracing will not work.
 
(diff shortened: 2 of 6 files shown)


shuding:
Hmm. Our Nextra package is very special which is a mix of ESM and CJS, with a bunch of webpack / esbuild hacks. Our linter rules are too strict and they just don't understand these.

shuding:
```
/home/runner/work/nextra/nextra/packages/nextra/src/types.ts
Error:   138:24  error  Irregular whitespace not allowed  no-irregular-whitespace
```

I don't see anything wrong on that line...

B2o5T:
> `@typescript-eslint/ban-ts-comment` is funny. If people manually add a comment to disable type checking, there is a reason for that. Human shouldn't be controlled by rules.

It’s a rule that is included in the typescript-eslint/recommended config with the right reason do not use ts-ignore in favor ts-expect-error - to describe why you ignoring type error

shuding:
Yeah but we are doing some hacky things in this package, but the recommended ruleset only covers normal usage... :(

I was just a bit said because I spent 90% of time fixing these linting errors (which are not real problems) than writing the actual code...

shuding:
Why did the CI error with a file I didn't edit... 🤔

B2o5T:
> Why did the CI error with a file I didn't edit... 🤔

I take look later today, maybe due eslint `--cache` flag

B2o5T:
it happened due to this PR https://github.com/shuding/nextra/pull/1296
image

shuding:
Wow... thank you @B2o5T 🙏

B2o5T:
btw it is very nice improvement, saw these problems too 😅

Issue #1273, opened 2023-01-15
Page titles not automatically capitalized with Title package


Ragura:
The docs state the following about pages that are not mentioned inside `_meta.json`:

_By default, the page map contains all .md and .mdx filenames and the directory structure, sorted alphabetically. Then, Nextra will use the [title(opens in a new tab)](https://github.com/vercel/title) package to get formatted page names from filenames._

This doesn't seem to be working (or I'm misunderstanding it). The sidebar still shows the lower-cased titles. Even in the official docs, under the `Docs Theme -> Built-ins` you can see the names of `callout` and `tabs` not being capitalized correctly.

PR #1312, opened 2023-01-19
fix capitalizing of undefined `_meta.json` files

B2o5T:
was broken by https://github.com/shuding/nextra/pull/1168 fixes https://github.com/shuding/nextra/issues/1273


docs/pages/docs/_meta.json
```diff
@@ -1,13 +1,13 @@
...
   "-- Themes --": {
     "type": "separator",
     "title": "Themes"
   },
-  "docs-theme": "Docs Theme",
-  "blog-theme": "Blog Theme",
-  "custom-theme": "Custom Theme",
+  "docs-theme": "",
+  "blog-theme": "",
+  "custom-theme": "",
```
B2o5T, line 10: I propose to capitalize undefined files in `_meta.json` and files with empty titles because in 99.9% cases, you don't want to render empty titles, even if you'll want you can use `" "`

in the above case we DRY and don't repeat `Docs Theme`, `Blog Theme`, etc. this will be automatically handled by `title` package

shuding (approved):

PR #1317, opened 2023-01-19
Add transformPageOpts option for advanced use cases

shuding:
This is an option that mostly helps custom themes. For example if the site has 1000 pages, every page will bundle the meta information (such as front matter) for all other pages and that's 1000x1000 copies. If the theme doesn't show a global sidebar for everything (unlike our `nextra-theme-docs`), that's totally unnecessary.

### DIFF
--- a/packages/nextra/src/use-internals.ts
+++ b/packages/nextra/src/use-internals.ts
@@ -19,6 +19,7 @@ export function useInternals() {
   // should be removed after compilation and it's fine to put the effect under
   // if, because hooks' order is still stable.
   if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
     useEffect(() => {
       const trigger = () => rerender({})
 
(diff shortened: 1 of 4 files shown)


packages/nextra/src/use-internals.ts
```diff
@@ -15,19 +15,23 @@ export function useInternals() {
...
   const rerender = useState({})[1]
 
-  useEffect(() => {
-    if (IS_PRODUCTION) return
-    const trigger = () => rerender({})
+  // The HMR handling logic is not needed for production builds, the condition
+  // should be removed after compilation and it's fine to put the effect under
+  // if, because hooks' order is still stable.
+  if (process.env.NODE_ENV !== 'production') {
+    useEffect(() => {
```
B2o5T, line 23: you can use `eslint-disable-next-line` comment for this case since this code will be tree shaked

B2o5T (approved):

PR #1320, opened 2023-01-19
Fix build error of resolving subpath imports

shuding:
I'm not 100% sure why there's this error, but it seems that some times the webpack resolver doesn't handle `./*` correctly:

CleanShot 2023-01-19 at 23 51 33@2x


B2o5T:
Nice!

B2o5T (approved):

Issue #1315, opened 2023-01-19
The Sidebar is loading all tsx files under page folder


medkrimi:
I am using Nextra into an existing NextJS app. I upgraded from 2.0.1 to 2.2.7. 

Now, all tsx files within the page folder, are displayed as an item into the left sidebar menu. 
In 2.0.1 it was only taking mdx files.

onceLearner:
Same issue with version 2.2.9
Normally only `/dev` folder should be on the sidebar.
<img width="210" alt="image" src="https://user-images.githubusercontent.com/41970812/213674421-ddc50be9-15d2-4304-9f76-891c48f0a3bd.png">

PR #1331, opened 2023-01-20
fix when sidebar show non-md folders

B2o5T:
fixes https://github.com/shuding/nextra/issues/1315

### DIFF
--- a/packages/nextra/__test__/__snapshots__/page-map.test.ts.snap
+++ b/packages/nextra/__test__/__snapshots__/page-map.test.ts.snap
@@ -130,10 +130,6 @@ exports[`Page Process > pageMap en-US 1`] = `
           "name": "team",
           "route": "/about/team",
         },
-        {
-          "data": {},
-          "kind": "Meta",
-        },
       ],
       "kind": "Folder",
       "name": "about",
@@ -162,10 +158,6 @@ exports[`Page Process > pageMap en-US 1`] = `
           "locale": "en-US",
           "name": "swr-v1",
           "route": "/blog/swr-v1",
-        },
-        {
-          "data": {},
-          "kind": "Meta",
         },
       ],
       "kind": "Folder",
@@ -322,18 +314,10 @@ exports[`Page Process > pageMap en-US 1`] = `
                       "name": "two",
                       "route": "/docs/advanced/more/tree/two",
                     },
-                    {
-                      "data": {},
-                      "kind": "Meta",
-                    },
                   ],
                   "kind": "Folder",
                   "name": "tree",
                   "route": "/docs/advanced/more/tree",
-                },
-                {
-                  "data": {},
-                  "kind": "Meta",
                 },
               ],
               "kind": "Folder",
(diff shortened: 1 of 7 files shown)


packages/nextra/__test__/__snapshots__/page-map.test.ts.snap
```diff
@@ -663,8 +624,19 @@ exports[`Page Process > pageMap ru 1`] = `
...
+              "theme": {
+                "layout": "raw",
+              },
+              "type": "page",
+            },
+            "acknowledgement": "🧩 Acknowledgement",
+            "team": "👥 Team",
+          },
           "kind": "Meta",
+          "locale": "en-US",
```
B2o5T, line 639: I guess it fix even silent bug :))

packages/nextra/__test__/page-map.test.ts
```diff
@@ -54,4 +54,12 @@ describe('Page Process', () => {
...
     expect(gettingStartData.pageMap).toEqual(indexData.pageMap)
   })
+
+  it("should not add to pageMap folders that doesn't contain markdown files", async () => {
+    const { items } = await collectFiles(
+      path.join(PAGES_DIR, 'docs/test-non-md-folders'),
+      ['']
+    )
+    expect(items.length).toBe(0)
+  })
```
B2o5T, line 64: small test (:

packages/nextra/src/plugin.ts
```diff
@@ -152,15 +152,11 @@ export async function collectFiles(
...
 
-  const mdxPagesAndFolders: (MdxFile | Folder)[] = []
   const mdxPages: MdxFile[] = []
   const metaLocaleIndexes = new Map<string, number>()
 
   for (let i = 0; i < items.length; i++) {
     const item = items[i]
-    if (item.kind === 'MdxPage' || item.kind === 'Folder') {
-      mdxPagesAndFolders.push(item)
-    }
```
B2o5T, line 163: collection was updated but never used => removed

mattvb91:
Is there any chance we could add a config option to say "subpath" that the sidebar will load only files based FROM that subpath. For example I want to host my docs under `/docs` but I want the sidebar to START at that level. Right now the sidebar would have `docs` as a collapsible but I want that to be the root.

Probably not relevant for this PR?

shuding:
@mattvb91 you can use `"display": "children"` for that case.

packages/nextra/src/plugin.ts
```diff
@@ -183,15 +179,7 @@ export async function collectFiles(
...
 
-    if (metaIndex === undefined) {
-      // Create a new meta file if it doesn't exist.
-      fileMap[metaPath] = {
-        kind: 'Meta',
-        ...(locale && { locale }),
-        data: Object.fromEntries(defaultMeta)
-      }
-      items.push(fileMap[metaPath])
-    } else {
```
shuding, line 194: @B2o5T It seems like this change caused `callout` and `tabs` to be lowercased:

CleanShot 2023-01-24 at 14 23 13@2x

https://nextra.site/docs/docs-theme/built-ins

B2o5T: Saw it yesterday, will fix it !

Issue #1303, opened 2023-01-19
Import markdown from parent directory


andreaswilli:
Thanks for the nice project!

I'm currently migrating from v1 to v2 and I'm getting this error when I try to import a markdown file from outside of the docs directory:
```
Module not found: Can't resolve 'nextra-theme-docs/style.css'
> 1 | import 'nextra-theme-docs/style.css'
  2 | /*@jsxRuntime automatic @jsxImportSource react*/
  3 | import {useMDXComponents as _provideComponents} from "nextra/mdx";
  4 | import $NextImageNextra from "next/image";
```
Here is my [repo](https://github.com/andreaswilli/react-verification-input). The nextra site is located in `/docs-src` while I want to include a part of the `README.md` file.
With v1 this worked fine, but I can't get it working with the latest version.
Any help is appreciated.

shuding:
I saw you are using `2.0.0` in your repo, did you try latest version?

andreaswilli:
It is the latest version. `^2.0.0` means install the latest version < v3
You can see it in the [lock file](https://github.com/andreaswilli/react-verification-input/blob/master/docs-src/yarn.lock#L1522).

B2o5T:
I didn't get any errors from `next build` and in `next dev` I was able to render `README.md` 
<img width="1538" alt="image" src="https://user-images.githubusercontent.com/7361780/213791049-41f2afb2-45fa-4887-a5ed-5c561b15cad9.png">

<img width="619" alt="image" src="https://user-images.githubusercontent.com/7361780/213791157-ca818f57-d4c9-486f-883d-197cea01d257.png">


closing it, more information is required

B2o5T:
oh I need to change `README.md` path

andreaswilli:
Yes, I added a copy script as a workaround.
Sorry, should have mentioned that.

andreaswilli:
Thanks for the quick patch! My problem is solved 🎉

B2o5T:
Thanks for reporting it 😍

PR #1332, opened 2023-01-21
fix `Module not found: Can't resolve 'nextra-theme-docs/style.css'` for imported markdown files that located outside of CWD

B2o5T:
fixes https://github.com/shuding/nextra/issues/1303

### DIFF
--- a/packages/nextra/src/compile.ts
+++ b/packages/nextra/src/compile.ts
@@ -109,7 +109,7 @@ export async function compileMdx(
 
   // https://github.com/shuding/nextra/issues/1303
   const isFileOutsideCWD =
-    !isPageImport && path.relative(CWD, filePath).startsWith('../')
+    !isPageImport && path.relative(CWD, filePath).startsWith('..')
 
   const compiler =
     (useCachedCompiler && cachedCompilerForFormat[format]) ||
--- a/packages/nextra/src/compile.ts
+++ b/packages/nextra/src/compile.ts
@@ -1,3 +1,5 @@
+import path from 'node:path'
+import { createRequire } from 'node:module'
 import { createProcessor, ProcessorOptions } from '@mdx-js/mdx'
 import { Processor } from '@mdx-js/mdx/lib/core'
 import remarkGfm from 'remark-gfm'
@@ -8,6 +10,7 @@ import grayMatter from 'gray-matter'
 import {
   remarkStaticImage,
   remarkHeadings,
+  remarkReplaceImports,
   structurize,
   parseMeta,
   attachMeta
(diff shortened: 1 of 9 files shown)


packages/nextra/src/loader.ts
```diff
@@ -172,10 +175,8 @@ async function loader(
     : ''
 
   // Imported as a normal component, no need to add the layout.
-  if (!pageImport) {
-    return `${cssImport}
-${result}
-export default MDXContent`
+  if (!isPageImport) {
+    return result
```
B2o5T, line 179: we don't need import styles for imported markdowns since it will be available from file which imports it

packages/nextra/src/compile.ts
```diff
@@ -103,13 +107,19 @@ export async function compileMdx(
...
+
   const compiler =
     (useCachedCompiler && cachedCompilerForFormat[format]) ||
     (cachedCompilerForFormat[format] = createProcessor({
       jsx,
       format,
       outputFormat,
-      providerImportSource: 'nextra/mdx',
+      providerImportSource: isFileOutsideCWD
+        ? require.resolve('nextra').replace(/index\.js$/, 'mdx.mjs') // fixes Package subpath './mdx' is not defined by "exports"
```
B2o5T, line 121: For some reason I couldn't use `require.resolve('nextra/mdx')` 🤔, got error Package subpath './mdx' is not defined by "exports"

packages/nextra/tsup.config.ts
```diff
@@ -16,43 +55,20 @@ export default defineConfig([
...
+      '!src/mdx-plugins/remark-replace-imports.ts'
+    ],
+    ...esmOptions
+  },
+  {
+    name: 'nextra-esm-import.meta',
+    entry: ['src/compile.ts', 'src/mdx-plugins/remark-replace-imports.ts'],
+    ...esmOptions,
+    // import.meta is available only from es2020
+    target: 'es2020'
```
B2o5T, line 71: should we use es2020 for all esm files?

packages/nextra/src/compile.ts
shuding: I think that's because we don't have a cjs export

packages/nextra/src/compile.ts
```diff
@@ -103,13 +107,19 @@ export async function compileMdx(
   const format =
     _format === 'detect' ? (filePath.endsWith('.mdx') ? 'mdx' : 'md') : _format
 
+  // https://github.com/shuding/nextra/issues/1303
+  const isFileOutsideCWD =
+    !isPageImport && path.relative(CWD, filePath).startsWith('../')
```
shuding, line 112:
```suggestion
    !isPageImport && path.relative(CWD, filePath).startsWith('..')
```

Probably breaks windows

Issue #1335, opened 2023-01-21
Error in config ``: invalid theme config


ChristianIvicevic:
Recently I have been running into a reported zod error during validation of my custom theme config. The structure of my file is the following:
```jsx
// @ts-check
/** @type {import('nextra-theme-docs').DocsThemeConfig} */
const config = {
  banner: { text: '...' },
  editLink: { component: () => null },
  feedback: { content: () => null },
  footer: { component: () => null },
  head: (<>{/* ... */}</>),
  logo: (<>{/* ... */}</>),
  navigation: { next: false, prev: false },
  search: { placeholder: '...' },
  titleSuffix: '...',
  toc: { float: true },
  useNextSeoProps: () => ({ titleTemplate: '...' })
};

module.exports = config;
```
Do you have any suggestions how I might debug this to deduce what's wrong with the config? To clarify, the offending key in the error message is literally an empty string:
![image](https://user-images.githubusercontent.com/45569997/213875563-5abe118a-a06a-444f-89ba-f4f0e9cd54fd.png)

ChristianIvicevic:
The `titleSuffix` property I had was superseded by the `useNextSeoProps` as far as I can tell and was the cause for the error. Keeping this issue open though since the error message isn't really pointing out the correct key.

B2o5T:
Yeah, error message should be fixed

PR #1341, opened 2023-01-22
use `z.strictObject` instead `z.object` in zod theme validation, improve zod error messages

B2o5T:
fixes https://github.com/shuding/nextra/issues/1335

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -29,7 +29,10 @@ module.exports = {
         '@typescript-eslint/prefer-optional-chain': 'error',
         'no-else-return': ['error', { allowElseIf: false }],
         'no-lonely-if': 'error',
-        'prefer-destructuring': ['error', { VariableDeclarator: { object: true } }],
+        'prefer-destructuring': [
+          'error',
+          { VariableDeclarator: { object: true } }
+        ],
         // todo: enable
         '@typescript-eslint/no-explicit-any': 'off',
         '@typescript-eslint/no-non-null-assertion': 'off',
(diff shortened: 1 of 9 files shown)


.eslintrc.cjs
```diff
@@ -59,6 +62,12 @@ module.exports = {
               'CallExpression[callee.name=useMemo][arguments.1.type=ArrayExpression][arguments.1.elements.length=0]',
             message:
               "`useMemo` with an empty dependency array can't provide a stable reference, use `useRef` instead."
+          },
+          {
+            // ❌ z.object(…)
+            selector:
+              'MemberExpression[object.name=z] > .property[name=object]',
+            message: 'Use z.strictObject is more safe.'
```
B2o5T, line 70: z.strictObject is safer since it's disallow unknown object keys

packages/nextra-theme-docs/src/contexts/config.tsx
```diff
@@ -58,21 +54,17 @@ export const ConfigProvider = ({
   }
   if (!isValidated) {
     try {
-      theme = themeSchema.parse(theme, {
-        errorMap: () => ({ message: 'Invalid theme config' })
-      })
```
B2o5T, line 63: we should not hide the original error message, otherwise, for example if we set `defaultCollapseLevel: -1` we'll not understand what's wrong, now it's clearer
image

.eslintrc.cjs
B2o5T: image

packages/nextra-theme-docs/src/contexts/config.tsx
B2o5T: I adjusted the path in https://github.com/shuding/nextra/pull/1343 PR by

```js
`Path: "${issue.path.join('.')}"`
```

like it was before, so the first error will end with `Path: "sidebar.defaultMenuCollapseLevel"`

PR #1360, opened 2023-01-22
validate `_meta.json` with zod

B2o5T:


packages/nextra-theme-docs/src/utils/normalize-pages.ts
```diff
@@ -9,18 +9,10 @@ function extendMeta(
...
 
-/**
- * An option to control how an item should be displayed in the sidebar:
- * - `normal`: the default behavior, item will be displayed
- * - `hidden`: the item will not be displayed in the sidebar entirely
- * - `children`: if the item is a folder, itself will be hidden but all its children will still be processed
- */
-export type Display = 'normal' | 'hidden' | 'children'
-
 export interface Item extends MdxFile {
```
B2o5T, line 16: what is `Item` and which `type` should it have? Since we have `PageItem` that should have `type: page` and DocsItem with `type: doc` (that is by default). I am only confused by `interface Item` 😅

Issue #1032, opened 2022-12-05
raw string `__esModule` will break nextra


promer94:
## Link to reproduction
https://github.com/promer94/nextra-evil-string

## To Reproduce

```
pnpm i
pnpm dev
```
then navigate to about page, the error will show up

<img width="1250" alt="Screenshot 2022-12-05 at 20 07 49" src="https://user-images.githubusercontent.com/15878786/205633690-173c5a8b-61e5-47f6-beb5-27da4b44efb2.png">

promer94:
root cause: https://github.com/vercel/next.js/pull/38181/files

MDX page will treadted as `cjs` module when it contains `__esModule`. Then webpack will try to resolve the `require` field of `next/ssg` which will cause the error

shuding:
Oh this is blocking another site that I'm helping moving...

shuding:
Is there a chance to workaround it inside Nextra? 💭

shuding:
Super dirty hack: replace `__esModule` with `__es[zero width space]Module` in the content, lol.

B2o5T:
Bundle `next/ssg` to both cjs and esm?

PR #1368, opened 2023-01-23
fix raw `__esModule` string ☠️

B2o5T:
fixes https://github.com/shuding/nextra/issues/1032

### DIFF
--- a/packages/nextra/src/compile.ts
+++ b/packages/nextra/src/compile.ts
@@ -171,7 +171,8 @@ export async function compileMdx(
     const readingTime = vFile.data.readingTime as ReadingTime | undefined
 
     return {
-      result: String(vFile),
+      // https://github.com/shuding/nextra/issues/1032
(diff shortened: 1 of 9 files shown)


packages/nextra/src/compile.ts
```diff
@@ -171,7 +171,8 @@ export async function compileMdx(
     const readingTime = vFile.data.readingTime as ReadingTime | undefined
 
     return {
-      result: String(vFile),
+      // https://github.com/shuding/nextra/issues/1032
+      result: String(vFile).replaceAll('__esModule', '_\\_esModule'),
```
B2o5T, line 175: 😎

shuding: Wow why would that work?! Isn't that `\`...?

B2o5T: \ is unnecessary escape

PR #1386, opened 2023-01-24
fixes hydration error for /404 route, allow define theme from _meta.json for /404 route

B2o5T:
## before (the theme was not applied even though we defined it in _meta.json)

image

## now

image


packages/nextra-theme-docs/src/index.tsx
```diff
@@ -21,22 +21,7 @@ import { getComponents } from './mdx-components'
 import { ActiveAnchorProvider, ConfigProvider, useConfig } from './contexts'
 import type { PageTheme } from './constants'
 import { DEFAULT_LOCALE, PartialDocsThemeConfig } from './constants'
-import { getFSRoute, normalizePages, renderComponent } from './utils'
-
-function useDirectoryInfo(pageMap: PageMapItem[]) {
-  const { locale = DEFAULT_LOCALE, defaultLocale, asPath } = useRouter()
-
-  return useMemo(() => {
-    // asPath can return redirected url
```
B2o5T, line 30: since asPath will be redirected URL, normalize pages will never return correct pageItem

packages/nextra-theme-docs/src/utils/use-fs-route.ts
```diff
@@ -0,0 +1,20 @@
...
+import { DEFAULT_LOCALE, ERROR_ROUTES } from '../constants'
+
+const template = 'https://nextra.vercel.app'
+
+export const useFSRoute = () => {
+  const { locale = DEFAULT_LOCALE, asPath, route } = useRouter()
+
+  return useMemo(() => {
+    // asPath can return redirected url
+    const clientRoute = ERROR_ROUTES.has(route) ? route : asPath
```
B2o5T, line 12: use `route` only if it 404/500 route, `asPath` for other ones

packages/nextra/src/locales.ts
```diff
@@ -38,7 +38,7 @@ export function locales(request: NextRequest) {
 
   const shouldHandleLocale =
     !/^\/(api|_next)\//.test(nextUrl.pathname) &&
-    !/\.(jpe?g|svg|png|webmanifest|xml)$/.test(nextUrl.pathname) &&
+    !/\.(jpe?g|svg|png|webmanifest|xml|ico)$/.test(nextUrl.pathname) &&
```
B2o5T, line 41: do not redirect `favicon.ico`

package.json
```diff
@@ -25,6 +25,7 @@
     "@typescript-eslint/eslint-plugin": "5.49.0",
     "@typescript-eslint/parser": "5.49.0",
     "eslint": "8.32.0",
+    "eslint-config-prettier": "^8.6.0",
```
B2o5T, line 28: to avoid conflicts with eslint plugins

packages/nextra-theme-docs/__test__/use-fs-route.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(value).toEqual('/404')
+  })
+
+  it('should return /500', () => {
+    const value = mockAndRenderHook({
+      asPath: '/error',
+      route: '/500'
+    })
+    expect(value).toEqual('/500')
+  })
```
B2o5T, line 70: added tests for it

packages/nextra-theme-docs/src/utils/use-fs-route.ts
```diff
@@ -0,0 +1,20 @@
...
+import { useRouter } from 'next/router'
+import { DEFAULT_LOCALE, ERROR_ROUTES } from '../constants'
+
+const template = 'https://nextra.vercel.app'
+
+export const useFSRoute = () => {
+  const { locale = DEFAULT_LOCALE, asPath, route } = useRouter()
+
+  return useMemo(() => {
+    // asPath can return redirected url
```
B2o5T, line 11:
```suggestion
    // because for the 404 route `asPath` will be redirected URL and `normalizePages` will never return correct pageItem
```

PR #1406, opened 2023-01-26
Updated SEO options for i18n

ramozdev:
- Changed example code to conditionally to avoid adding the suffix to the homepage to fix TypeScript error.

```
Type 'undefined' is not assignable to type 'NextSeoProps'.
```

- Added an example code to conditionally to avoid adding the suffix to the homepage when using i18n.


B2o5T (changes requested):
we can simplify to use `asPath` instead

B2o5T:
closed in favor https://github.com/shuding/nextra/pull/1414

PR #1408, opened 2023-01-27
fix custom class names fully overriding default component styles

### LINKED ISSUE #1398: Nextra components don't allow extending the class names
https://github.com/shuding/nextra/blob/a404ef7c6ea321995cff431a915c2bc407a45c8a/packages/nextra/src/components/td.tsx#L3-L8

Recently I am trying reuse as many components from Nextra to not hard code classes or come up with own styling for my components. Now I was attempting to add Tailwind to pass on minor customizations to existing components such as the `Td` component linked above and noticed that `<Td className="my-custom-class" />` fully overrides your own styles.

I'd propose to discuss whether you want to support consumers like us to reuse these components. If so, then it might be nice to rewrite the components s.t. custom class names get appended instead of fully overwriting the existing ones. As a workaround I have to write inline styles instead of being able to use class names.

ChristianIvicevic:
Just some low hanging fruit.
Fixes #1398

### DIFF
--- a/.changeset/spotty-poets-count.md
+++ b/.changeset/spotty-poets-count.md
@@ -2,4 +2,4 @@
 'nextra': patch
 ---
 
-fix custom class names fully overriding default component styles
+do not override `className` for `<Td />`,  `<Th />` and  `<Tr />` components
--- a/.changeset/spotty-poets-count.md
+++ b/.changeset/spotty-poets-count.md
@@ -1,0 +1,5 @@
+---
+'nextra': patch
+---
+
+fix custom class names fully overriding default component styles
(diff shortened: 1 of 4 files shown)


.changeset/spotty-poets-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'nextra': patch
+---
+
+fix custom class names fully overriding default component styles
```
B2o5T, line 5:
```suggestion
do not override `className` for `<Td />`,  `<Th />` and  `<Tr />` components
```

B2o5T (approved):

Issue #1389, opened 2023-01-24
"darkMode": false is not working


gaurav-pantelwar:
I want to have a light theme only on my API docs platform, but I am not able to do using "darkMode": false flag.

gaurav-pantelwar:
Is this flag supposed to force the light theme or just for hiding the theme toggle button?

B2o5T:
> Is this flag supposed to force the light theme or just for hiding the theme toggle button?

just for hiding the theme switch button

> I want to have a light theme only on my API docs platform, but I am not able to do using "darkMode": false flag.

use `nextThemes.forcedTheme` for it

I think we'll fix it to by setting `nextThemes.forcedTheme` to `light` when `darkMode` is `false`

B2o5T:
see my comment on how to fix it https://github.com/shuding/nextra/issues/1630#issuecomment-1467079236

Issue #478, opened 2022-06-07
Dark mode can't be deactivated


zunkelty:
When you disable `darkMode` in `theme.config.js`, I expect Nextra to be in light mode only. However, when you activate `prefers-color-scheme: dark`, the setting is overridden and the page is shown in dark mode. This is a problem if you have explicitly turned off dark mode before.
The error is due to [`next-themes`](https://github.com/pacocoursey/next-themes) that automatically detects such settings and sets the color scheme to dark. 

My suggestion: Dark mode is an opt-out feature which means that if `darkMode` is disabled, no theming logic whatsoever is active. It doesn't make sense to include `next-themes` still (and use it's`ThemeProvider`) when dark mode is deactivated.

Unfortunately, right now the option to disable dark mode is not how you expect it to be (and probably not how it's wanted?).

(I'm on version `2.0.0-alpha.54` if that makes any difference)

smakosh:
Running into the same issue on the following project: https://github.com/smakosh/ontwik-ui

darkMode is turned off and default theme is set to `light` here: https://github.com/smakosh/ontwik-ui/blob/master/apps/public-docs/theme.config.js#L16

Still I'm getting dark mode when visiting the docs website: https://ontwik-ui.smakosh.com/

erikmoyaers:
You can only allow light mode by setting the localStorage theme to 'light' whenever the localStorage theme has a different value. And only render the page when localStorage is 'light'.

in _app.js:
```js
export default function Nextra({ Component, pageProps }) {
    const [canRender, setCanRender] = useState(false)

    useEffect(() => {
        if (window.localStorage.getItem("theme") !== "light") {
            window.localStorage.setItem("theme", "light")
        }
        setCanRender(true)
    }, [])

    return (
        canRender ? <Component {...pageProps} /> : <></>
    )
}
```

B2o5T:
Let’s track fixing this unexpected behavior here https://github.com/shuding/nextra/issues/1389

PR #1417, opened 2023-01-28
force theme to `light` if `darkMode: false` was set

B2o5T:
fixes https://github.com/shuding/nextra/issues/1389 https://github.com/shuding/nextra/issues/478

### DIFF
--- a/packages/nextra-theme-blog/src/index.tsx
+++ b/packages/nextra-theme-blog/src/index.tsx
@@ -42,7 +42,10 @@ export default function Layout({
   const extendedConfig = { ...DEFAULT_THEME, ...context.themeConfig }
 
   return (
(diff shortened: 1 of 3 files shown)


packages/nextra-theme-blog/src/index.tsx
```diff
@@ -42,7 +42,10 @@ export default function Layout({
   const extendedConfig = { ...DEFAULT_THEME, ...context.themeConfig }
 
   return (
-    <ThemeProvider attribute="class" defaultTheme="system" enableSystem>
+    <ThemeProvider
+      attribute="class"
+      forcedTheme={extendedConfig.darkMode ? undefined : 'light'}
+    >
```
B2o5T, line 48: `defaultTheme="system"` and `enableSystem=true` are default values

Issue #1207, opened 2023-01-08
Lots of code smells


R-Bower:
Hey there, great package.  I've spent some time to migrate over from Docusaurus and am loving the ergonomics provided by nextra-theme-docs.  I'm also a huge Next.js fan, so this was a no-brainer.  

In using this package, I've stumbled on some errors that prompted me to check the source code.  There are several code smells in the nextra and nextra-theme-docs package.  Surprisingly, the packages don't use `eslint-plugin-react` or `eslint-plugin-react-hooks` which would help catch most of these mistakes.  

I'm frequently seeing this warning in local dev mode:

Warning: React has detected a change in the order of Hooks called by ... This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://reactjs.org/link/rules-of-hooks

This is typically indicative of code that doesn't follow React's rules of hooks.  

More code smells:

1. This callback is being declared after it's used earlier in the function.
https://github.com/shuding/nextra/blob/main/packages/nextra-theme-docs/src/components/search.tsx#L179

3. This hook is being called conditionally, which is an anti-pattern: 
https://github.com/shuding/nextra/blob/main/packages/nextra-theme-docs/src/mdx-components.tsx#L86

4. The return result of this hook is being set to a variable outside of React's lifecycle.  Pretty sure this is also an anti-pattern.
https://github.com/shuding/nextra/blob/main/packages/nextra-theme-docs/src/mdx-components.tsx#L86

5. The definition for `HeaderLink` is a bit confusing.  I don't see a need for this to be curried.  Why not make it a simple function component?
https://github.com/shuding/nextra/blob/main/packages/nextra-theme-docs/src/mdx-components.tsx#L77

```
// potential fix:
function HeaderLink({
  children,
  id,
  context,
  Tag,
  ...props
}: ComponentProps<TagType> & {
  Tag: TagType
  context: { index: number }
}): ReactElement {
  // ...
}
...

h2: props => <HeaderLink context={context} Tag={'h2'} {...props} />,
h3: props => <HeaderLink context={context} Tag={'h3'} {...props} />,
...
```

There are more issues in other files as well.  

@shuding I'm happy to create a PR with some fixes.  Let me know, thanks!

B2o5T:
Enabling react, react-hooks, jsx-a11y, and next plugins in my todo list, recently eslint/typescript-recommended configs we’re already enabled

shuding:
Feel free to open PR for these, thanks!

[...]

PR #1434, opened 2023-01-30
fix: move intersection observe and slugs into ActiveAnchorProvider

promer94:
In this way,  React Hook "useSetActiveAnchor" would not be called conditionally.

fixes https://github.com/shuding/nextra/issues/1207

### DIFF
--- a/packages/nextra-theme-docs/src/contexts/active-anchor.tsx
+++ b/packages/nextra-theme-docs/src/contexts/active-anchor.tsx
@@ -3,7 +3,6 @@ import type {
   ReactElement,
   ReactNode,
   SetStateAction,
-  MutableRefObject
 } from 'react';
 import 'intersection-observer'
 import {
@@ -11,7 +10,6 @@ import {
   useContext,
   useState,
   useRef,
-  createRef
 } from 'react';
 import { IS_BROWSER } from '../constants';
 
@@ -30,7 +28,7 @@ const SetActiveAnchorContext = createContext<
   Dispatch<SetStateAction<ActiveAnchor>>
 >(v => v)
 
-const IntersectionObserverContext = createContext<MutableRefObject<IntersectionObserver | null>>(createRef())
+const IntersectionObserverContext = createContext<IntersectionObserver | null>(null)
 const slugs = new WeakMap()
 const SlugsContext = createContext<WeakMap<any, any>>(slugs)
 // Separate the state as 2 contexts here to avoid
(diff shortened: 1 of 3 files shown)


packages/nextra-theme-docs/src/contexts/active-anchor.tsx
```diff
@@ -24,22 +27,89 @@ const SetActiveAnchorContext = createContext<
...
+export const useIntersectionObserver = () => useContext(IntersectionObserverContext)
+export const useSlugs = () => useContext(SlugsContext)
+
 export const ActiveAnchorProvider = ({
   children
 }: {
   children: ReactNode
 }): ReactElement => {
   const [activeAnchor, setActiveAnchor] = useState<ActiveAnchor>({})
+  const [observer] = useState<IntersectionObserver | null>(() => {
```
B2o5T, line 47: we can move `observer` outside of provider definition to avoid confusing `const [observer] = useState` without `setState`

B2o5T (approved):

packages/nextra-theme-docs/src/contexts/active-anchor.tsx
promer94: It needs to use `setActiveAnchor `, so we still need to initialize it in provider. But i think i could use a `useRef` here.

packages/nextra-theme-docs/src/contexts/active-anchor.tsx
```diff
@@ -24,22 +30,85 @@ const SetActiveAnchorContext = createContext<
...
+        threshold: [0, 1]
+      }
+    )
+  }
   return (
     <ActiveAnchorContext.Provider value={activeAnchor}>
       <SetActiveAnchorContext.Provider value={setActiveAnchor}>
-        {children}
+        <SlugsContext.Provider value={slugs}>
+          <IntersectionObserverContext.Provider value={observerRef}>
```
B2o5T, line 108:
```suggestion
          <IntersectionObserverContext.Provider value={observerRef.current}>
```

packages/nextra-theme-docs/src/mdx-components.tsx
```diff
@@ -78,28 +19,28 @@ function HeadingLink({
   tag: `h${2 | 3 | 4 | 5 | 6}`
   context: { index: number }
 }): ReactElement {
-  // eslint-disable-next-line react-hooks/rules-of-hooks -- todo: fix React Hook "useSetActiveAnchor" is called conditionally
-  setActiveAnchor ??= useSetActiveAnchor()
+  const setActiveAnchor = useSetActiveAnchor()
+  const slugs = useSlugs()
+  const observer = useIntersectionObserver().current
```
B2o5T, line 24:
```suggestion
  const observer = useIntersectionObserver()
```

B2o5T (approved):

Issue #1432, opened 2023-01-30
focus outline is cut off


n1ru4l:
<img width="305" alt="image" src="https://github.com/shuding/nextra/assets/14338007/bee75b27-ad2e-4624-8cc5-ff3a390720b8">
<img width="409" alt="image" src="https://github.com/shuding/nextra/assets/14338007/01646b73-502c-4ccc-adfe-936aaec51a04">

dimaMachina:
will be fixed in next alpha

PR #1437, opened 2023-01-31
Fix link overflow

niikeec:
This PR resolves issue #1432, except if a different approach should be taken.


shuding:
That overflow hidden is intentional, the same bug will be fixed with #1433. Thanks!

Issue #1425, opened 2023-01-29
`trailingSlash` breaks navigation (prev, next links)


alex-flea:
It looks like somewhere in the latest minor versions 2.2.X navigation was broken (at least with i18n enabled). It always shows `next` position as if it was at the first page of the current folder.

B2o5T:
Please add a reproduction

PR #1443, opened 2023-02-01
fix: remove trailing slash

niikeec:
This PR resolves issue #1425.


B2o5T (changes requested):
Nice! Can you add some test to ensure this will be not broken in the future, example of tests for this function are here https://github.com/shuding/nextra/blob/main/packages/nextra-theme-docs/__test__/use-fs-route.spec.ts

niikeec:
Pushed a test

B2o5T (approved):
Thank you!

niikeec:
No problem

Issue #1445, opened 2023-02-01
[nextra-theme-docs] Support Localization for `ThemeSwitch`


SonMooSans:
The theme switch component currently supports `Light`, `Dark`, and `System`, I'd hope to change their name depending on the languages the user using

I suggest adding a `useThemeOptions` property to the theme config, I can open a pull request for that

PR #1460, opened 2023-02-02
Configuration support for `ThemeSwitch` component

SonMooSans:
Original issue: #1445


Gary50613:
can I also add the Traditional Chinese version?

SonMooSans:
Sure! It had been added to the theme-theme-configuration page

SonMooSans:
> can I also add the Traditional Chinese version?

`zh-CN` is currently enough as a default configuration, but translations for French and Russian are welcome

Gary50613:
> > can I also add the Traditional Chinese version?
> 
> `zh-CN` is currently enough as a default configuration, but translations for French and Russian are welcome

Agree, other languages can overwrite the default function as well

docs/pages/docs/docs-theme/theme-configuration.mdx
```diff
@@ -483,6 +483,25 @@ export default {
...
 ```
 
+## Theme Switch
+
+You are able to customize the option names:
+
+```jsx
+export default {
+  themeSwitch: {
+    useOptions() {
```
nihgwu, line 493: API looks weird to me, can we use 
```
type ThemeSwitchLabel = 'toggle' | 'system' | 'light' | 'dark'
themeSwitch: {
  ...
  labels: Record<ThemeSwitchLabel, string> | (...) => Record<ThemeSwitchLabel, string>
}
```

nihgwu: OK I saw you are using hook style API in other places, and only for `locale`, can we simply make it as an parameter of the function like `({ locale }) => ...` or `({ router }) => ...`, don't think it's a good idea to require hook as option getter

SonMooSans: Nice idea! I did agree with that, it's similar to other existing solutions such as the `search` property in the docs theme

nihgwu: Anyway another point in my comments is that you can simply use object instead of array for localised labels, and you'd better to provide a way to localise the toggle button's aria-label

docs/pages/docs/docs-theme/theme-configuration.mdx
```diff
@@ -483,6 +483,26 @@ export default {
...
+You are able to customize the option names:
+
+```jsx
+export default {
+  themeSwitch: {
+    options() {
+      //hooks
+      
+      return [
+        { key: 'light', name: 'Light' },
```
nihgwu, line 497: That's not the way, what if user returns only one item or two? makes no sense to use a select anymore

still, my initial suggestion

```
themeSwitch: {
  labels: ({ locale }) => ({
    toggle: "Toggle theme",
    system: "With system",
    dark: "Dark mode",
    light: "Light mode",
  })
}
```
and if we want to keep consistency with other API, then rename it to `useLabels`

nihgwu: Those are accessible labels, and internally we will use it as `const options = [{key: 'system', label: labels.system...`
I really don't think user need to provide the full options, unless you support more themes, from the codebase, Nextra supports 3 modes, and can't be changed (e.g. I don't need `system` but I can't opt it out)

nihgwu: And even we support customise modes, like I only need two modes(light/dark), then I don't need a select at all, a toggle icon button will do the job and that's how it looks in most websites, then what's the point of `options`?

nihgwu: @B2o5T please provide more constructive comments

SonMooSans: I don't disagree that using an object instead of array might be better

Support for multi-themes or disabling `system` theme is unrelated to this pull request, it's just basically a config of the component, please override the next-themes configuration or open a feature request for that

My suggestion is to add a `component` property for overriding the whole component, similar to others

nihgwu: A good API is hiding internal details, by adding `options`, you made *visible* modes having accessible labels but what about the toggle button? Exposing another prop? It's also important for A11y

SonMooSans: The theme switch is a Select component based on headless ui, What did you mean by the toggle button?

nihgwu: the select trigger...

SonMooSans: ok I got you, the localisation for aria-label isn’t supported by Nextra (ex: Sidebar toggle) 
It’s weird to only support this feature in theme switch

SonMooSans:
Now, you can override the component with `themeSwitch.component` or customize its option names

docs/pages/docs/docs-theme/theme-configuration.mdx
```diff
@@ -483,6 +483,26 @@ export default {
...
+    options() {
+      return {
+        light: 'Light',
+        dark: 'Dark',
+        system: 'System'
+      }
+    }
+  }
+}
+```
```
B2o5T, line 504: can you update docs to mention `component` also? and use `<OptionTable />`

packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -301,6 +306,25 @@ export const DEFAULT_THEME: DocsThemeConfig = {
     titleComponent: ({ title }) => <>{title}</>,
     toggleButton: false
   },
+  themeSwitch: {
+    component: ThemeSwitch,
+    options: function useOptions() {
```
B2o5T, line 311:
```suggestion
    useOptions: function useOptions() {
```

I remember @shuding prefer `use` prefix for config options where you can use react hooks

docs/pages/docs/docs-theme/theme-configuration.mdx
nihgwu: > the localisation for aria-label isn’t supported by Nextra (ex: Sidebar toggle)
It’s weird to only support this feature in theme switch

So apparently you agree it's wrong, so why not do it right here, and we can address the issue in other places

packages/nextra-theme-docs/src/constants.tsx
nihgwu: still I think it's better to call it `useLabels`, or if you want to keep `useOptions` then, return `{labelLight, labelDark, labelSystem}`

SonMooSans: > still I think it's better to call it `useLabels`, or if you want to keep `useOptions` then, return `{labelLight, labelDark, labelSystem}`

Bad idea, I don't think calling items in a select as "label" is suitable...

SonMooSans: The `options` property has been renamed to `useOptions`! Also improved the documentation as well

B2o5T (approved):
Thanks for your first contribution!

PR #1471, opened 2023-02-06
Fix search text in French

EricCote:
The "Search documentation..." text was originally translated in French to "Rechercher de la documentation...".  This is wrong (and is overly long).  

In this setting, it should be "Rechercher documents..." or simply "Rechercher...". 

Reference:  https://translate.google.com/?sl=fr&tl=en&text=rechercher%20documents&op=translate&hl=en

Thanks


B2o5T:
Why your link points to google translate with `search documents` result?

PR #1491, opened 2023-02-10
update: theme-configuration.mdx

SwimmingPolar:
should consider i18n locale when composing url just to be accurate


B2o5T (approved):
Thank you!

PR #1497, opened 2023-02-11
fix local theme on Windows

KiruyaMomochi:
Fix the following error by adding `slash(...)`:

```
Error:
  x Bad character escape sequence, expected 2 hex characters
   ,-[C:\Users\x...\Projects\kyaru-blog\pages\blog\index.mdx:1:1]
 1 | import { setupNextraPage } from 'nextra/setup-page'
 2 | import __nextra_layout from 'C:\Users\x...\Projects\kyaru-blog\theme.tsx'
   :                                      ^^
 3 |
 4 |
   `----

Caused by:
    Syntax Error
```

### DIFF
--- a/nervous-tigers-camp.md
+++ b/nervous-tigers-camp.md
@@ -1,0 +1,5 @@
+---
+'nextra': patch
+---
+
+wrap path for local theme with `slash` for windows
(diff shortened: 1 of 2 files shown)


B2o5T (approved):
Could you add a changeset, so users will know about this fix?

PR #1504, opened 2023-02-13
Optimize Printing Experience

### LINKED ISSUE #1502: Add print style support
Recently I would like to export *.PDF from my nextra site, however, a problem appears that it's still using the web style. 
Is there any way to add a print style change so that the PDF will be better for reading?
e.g.: remove the nav bar, change the footer position, etc.
Many thanks.

MrWillCom:
Hide some unnecessary parts of the page, like navbar, sidebar, when being printed. fixes #1502

### DIFF
--- a/packages/nextra-theme-docs/src/components/footer.tsx
+++ b/packages/nextra-theme-docs/src/components/footer.tsx
@@ -8,7 +8,7 @@ import { renderComponent } from '../utils'
 export function Footer({ menu }: { menu?: boolean }): ReactElement {
   const config = useConfig()
   return (
-    <footer className="nx-bg-gray-100 nx-pb-[env(safe-area-inset-bottom)] dark:nx-bg-neutral-900">
+    <footer className="nx-bg-gray-100 nx-pb-[env(safe-area-inset-bottom)] dark:nx-bg-neutral-900 print:nx-bg-transparent">
       <div
         className={cn(
           'nx-mx-auto nx-flex nx-max-w-[90rem] nx-gap-2 nx-py-2 nx-px-4',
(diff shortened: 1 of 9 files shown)


packages/nextra-theme-docs/src/components/footer.tsx
```diff
@@ -8,7 +8,7 @@ import { renderComponent } from '../utils'
 export function Footer({ menu }: { menu?: boolean }): ReactElement {
   const config = useConfig()
   return (
-    <footer className="nx-bg-gray-100 nx-pb-[env(safe-area-inset-bottom)] dark:nx-bg-neutral-900">
+    <footer className="nx-bg-gray-100 nx-pb-[env(safe-area-inset-bottom)] dark:nx-bg-neutral-900 print:nx-bg-transparent">
```
B2o5T, line 11: I guess we can hide the footer while printing, also maybe we should hide the breadcrumb also?

And I like the idea of using `print:*` classes!

MrWillCom: In my opinion, I prefer save the footer and the breadcrumb. The footer tells the author and license information of the page. And the breadcrumb can be seen as the page header showing the current chapter. Plus, they are useful when the page is exported to PDF. Maybe we can simply make the footer take less area when being printed?

B2o5T (approved):
Could you add a changeset, so users will know about this improvement?

MrWillCom:
👌 Sure! Added in 8d5b57d.

packages/nextra-theme-docs/src/components/footer.tsx
```diff
@@ -24,7 +24,7 @@ export function Footer({ menu }: { menu?: boolean }): ReactElement {
       <div
         className={cn(
           'nx-mx-auto nx-flex nx-max-w-[90rem] nx-justify-center nx-py-12 nx-text-gray-600 dark:nx-text-gray-400 md:nx-justify-start',
-          'nx-pl-[max(env(safe-area-inset-left),1.5rem)] nx-pr-[max(env(safe-area-inset-right),1.5rem)]'
+          'nx-pl-[max(env(safe-area-inset-left),1.5rem)] nx-pr-[max(env(safe-area-inset-right),1.5rem)] print:nx-py-2'
```
B2o5T, line 27:
```suggestion
          'nx-pl-[max(env(safe-area-inset-left),1.5rem)] nx-pr-[max(env(safe-area-inset-right),1.5rem)]'
```

B2o5T (approved):

PR #1544, opened 2023-02-22
Adds Chimera UI to showcase

williamlmao:


B2o5T:
Sorry, I don’t think we’ll include it, we can't include every project who uses nextra, only popular

PR #1559, opened 2023-02-25
Filenames and titles to syntax-highlighting

mrmaglet:
Added a section on how to render and use the `filename` attribute (thus showing the title bar for a code section).

The attribute is frequently used throughout the Nextra documentation - but information on how to use it was missing. 

image


B2o5T (approved):
Nice catch! Thanks

PR #1565, opened 2023-02-27
Adds Kitchen to showcase

YounesBessa:
I replaced hop with kitchen because hop no longer uses nextra for their project


docs/pages/showcase.mdx
```diff
@@ -6,7 +6,7 @@ import swr from 'public/showcase/swr.png'
 import cobe from 'public/showcase/cobe.png'
 import javaScriptPatterns from 'public/showcase/javascript-patterns.jpg'
 import codesandbox from 'public/showcase/codesandbox.jpg'
-import hop from 'public/showcase/hop.jpeg'
```
AntoineKM, line 9: Remove hop image from `public/showcase`

AntoineKM (changes requested):

AntoineKM (approved):
It looks great thank you so much!

jerembdn (approved):

B2o5T:
Sorry, I don’t think we’ll include it, we can't include every project who uses nextra, only popular

PR #1534, opened 2023-02-20
Support markdown file path links

ocavue:
Closes https://github.com/shuding/nextra/issues/1532

This PR adds a Remark plugin, to remove the `.md` and `.mdx` extension from links.

PR #1615, opened 2023-03-08
Support Markdown links with query or anchor

ocavue:
This is a following up to my previous PR https://github.com/shuding/nextra/pull/1534. 

After this PR, nextra will remove the `.md` extension in the link, but keep the anchor and query, if exist. 

```
// Input markdown
[link](../my_markdown_file.md#heading)
```

```
// Output HTML
<a href="../my_markdown_file#heading">
```

Other updates in this PR:

- Replace `as any` to typescript 4.9's `satisfies` keyword.
- Update `tsup` by removing it from `pnpm.overrides`, so that it can support `satisfies`.

### DIFF
--- a/packages/nextra-theme-blog/tsup.config.ts
+++ b/packages/nextra-theme-blog/tsup.config.ts
@@ -7,6 +7,6 @@ export default defineConfig({
   dts: true,
   name: 'nextra-theme-blog',
   outExtension: () => ({ js: '.js' }),
-  target: tsconfig.compilerOptions.target,
+  target: tsconfig.compilerOptions.target as 'es2016',
   external: ['nextra']
 })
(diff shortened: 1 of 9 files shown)


packages/nextra-theme-blog/tsup.config.ts
```diff
@@ -7,6 +7,6 @@ export default defineConfig({
   dts: true,
   name: 'nextra-theme-blog',
   outExtension: () => ({ js: '.js' }),
-  target: tsconfig.compilerOptions.target,
+  target: tsconfig.compilerOptions.target as 'es2016',
```
B2o5T, line 10: I guess there is no workaround to avoid using `as` and take `const` value of target? found active issue https://github.com/microsoft/TypeScript/issues/32063

B2o5T (approved):
Nice! I also forget about `#` links, just one question and we ready to go 🚀

packages/nextra-theme-blog/tsup.config.ts
ocavue: I'm not sure what's better way to do it

PR #1621, opened 2023-03-09
fix: cusdis display  when config cusdis in blog

BlackBerry009:
I noticed that the pre-existing cusdis component was directly imported here, but after the last change, it seems that the comments are now being imported with a configuration. 
This might be a mistake as it could result in the cusdis component not being used.

### DIFF
--- a/packages/nextra-theme-blog/src/article-layout.tsx
+++ b/packages/nextra-theme-blog/src/article-layout.tsx
@@ -3,6 +3,7 @@ import Meta from './meta'
 import { MDXTheme } from './mdx-theme'
 import { useBlogContext } from './blog-context'
 import { BasicLayout } from './basic-layout'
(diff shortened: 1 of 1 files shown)


B2o5T:
cusdis components are optional, for this reason they are marked in `peerDependencyMeta` by @promer94 to reduce bundle size for those who don't need them

BlackBerry009:
> cusdis components are optional, for this reason they are marked in `peerDependencyMeta` by @promer94 to reduce bundle size for those who don't need them

I am aware that Cusdis is optional, but currently it is being used to configure the appId through the Cusdis options without addressing the issue of importing the Cusdis component in the code. This means that the internally wrapped Cusdis component will never be utilized.

BlackBerry009:
I get the current usage is like this

Issue #1648, opened 2023-03-16
Prefix for dates on posts using Blog theme


AndaxH:
The top of a `.md` page looks like this - 
```
---
title: Version Control Systems – Git and Friends
date: 2023-03-14
description: Keep track of changes to your code
tag: software development, version control, git
author: You
---
```

I'd love to be able to add a `datePrefix` or similar, something like - 
```
---
title: Version Control Systems – Git and Friends
date: 2023-03-14
datePrefix: Last updated on
description: Keep track of changes to your code
tag: software development, version control, git
author: You
---
```

which would then render "Last updated on Tue Mar 14 2023" instead of just "Tue Mar 14 2023"

B2o5T:
will be available in next release

PR #1652, opened 2023-03-17
feat: add date formatter

hothoney:
add date formatter to the configuration 
related to https://github.com/shuding/nextra/issues/1648


B2o5T (approved):
Thanks, only I don't think we should include date formatter on posts page

PR #1657, opened 2023-03-18
Provide item props to `titleComponent` 

irangarcia:
Give flexibility and allow to customize the sidebar item options by providing all the `item` props to the component. 

Simple example for render custom icons for each item:

**_overview.mdx_**
```mdx
---
title: Overview
icon: home
---

# Overview 
```

**_theme.config_**
```jsx
const mappedIcons = {
  home: <HomeIcon />
}

export default {
  sidebar: {
    titleComponent({ title, frontMatter }) {
      return (
        <>
          {frontMatter.icon !== undefined && mappedIcons[frontMatter.icon]}
          {title}
        </>
      )
    }
  }
}
```


NafiurSiddiqui:
Hey guys,

I tried the example here.

But I get an error

```
Server Error
TypeError: Cannot read properties of undefined (reading 'icon')

This error happened while generating the page. Any console logs will be displayed in the terminal window.

```

 Console logged `frontMatter` but it is undefined.

Do we have to install a plugin like `remark`or`gray-matter`  etc?

## Issue 2

In my case, I tried a workaround since mine would not be coming from a DB.
I directly rendered a svg component inside the `sidebar`. Works.

with a little inline css it looks like [this](https://ibb.co/1TytPN8)

Is there any way I can skip the parents to inherit this checklist? Something like negating grandparents and parents but only allowing child items to have the icons? Maybe a new feature?

PR #1692, opened 2023-03-28
fix crash of dev server when `_meta.js` contains errors

B2o5T:
fix crash of dev server when `_meta.js` contains errors

disallow caching of `_meta.js` on dev env

set minimal `node` as `16`


packages/nextra/src/plugin.ts
```diff
@@ -102,48 +101,58 @@ export async function collectFiles(
...
-            // is imported/exported by another _meta.js
-            data: { ...meta }
+          return fileMap[fp]
+        }
+
+        if (fileName === DYNAMIC_META_FILENAME) {
+          // _meta.js file. Need to check if it's dynamic (a function) or not.
+
+          // querystring to disable caching of module
+          const importPath = `${filePath}?d=${Date.now()}`
```
B2o5T, line 120: @shuding `_meta.js` never refreshes, this trick disables caching :)

PR #1695, opened 2023-03-28
chore: Remove an unnecessary changeset file

torn4dom4n:
I only remove an unnecessary changeset file.


B2o5T:
Thanks for the catch!

Issue #1703, opened 2023-03-29
Optionally disable "/" (forward slash) search shortcut


james-bowers:
Hey,

Thanks for nextra, it's very nice, however we're using interactive code editors in our docs, but we cannot write a closing HTML/JSX element because the `/` key is a shortcut for searching. Is there a way of turning this off in the configuration?

danez:
Exactly the same issue here, you beat me to 8 minutes :P

This is my site: https://react-docgen.dev/playground

danez:
@james-bowers which code editor are you using?

B2o5T:
I am against removing `cmd-/` shortcut (it is a bit popular and in the end we can still find a workaround like @danez mentioned)

adhrinae:
We're integrating [Sandpack](https://sandpack.codesandbox.io/) inside our internal documentation project made with Nextra.

Even though I set the editor of Sandpack's `contenteditable` to `true` the search bar always gets focused when I press the `/` key.

[The Sandpack editor component itself doesn't set the `contenteditable` attribute](https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/components/CodeEditor/CodeMirror.tsx#L468-L485), so I had to manually add it using the `useEffect` hook.

```jsx
const CustomSandpack = () => {
  const editorRef = useRef(null);
  // ...
  useEffect(() => {
    if (showCode && editorRef.current) {
      editorRef.current.contentEditable = 'true';
    }
  }, []);
  // ...
}
```

After that when I focused on the editor `document.activeElement.isContentEditable` become `true`, however, the search bar still hijacking the `/` keypress.

Is there any workaround for this?

dimaMachina:
closing due to a lack of interest and stale date

PR #1704, opened 2023-03-29
fix: Do not focus search when currently focusing a HTMLElement with contenteditable active

danez:
Fixes: #1703 

I tested this localy on my site by patching node_modules directly and it works at least with `codemirror`.


B2o5T (approved):
Thank you!

Issue #1619, opened 2023-03-09
Hide headings from TOC Sidebar 


niteshsrivats:
Is there a way to hide a heading from the TOC sidebar?

For example, on a page I might have:
```
### Node

#### Properties

type: A | B | C

##### Additional properties for type C

Lorem ipsum dolor sit amet...
```
But I might not want the H5 heading in my sidebar

# Request

I add a /ignore flag that doesn't include my H5 from the TOC sidebar
```
### Node

#### Properties

type: A | B | C

##### Additional Properties for type C /ignore

Lorem ipsum dolor sit amet...
```
# Current Workaround

Copying the CSS classes being applied into a HTML tag.
```
### Node

#### Properties

type: A | B | C

<h5 className={'nx-font-semibold nx-tracking-tight nx-text-slate-900 dark:nx-text-slate-100 nx-mt-8 nx-text-lg'}>
Additional Properties for type C
</h5>

Lorem ipsum dolor sit amet...
```

B2o5T:
you can just hide them with css
```css
.nextra-toc a.ltr:nx-ml-12,
.nextra-toc a.ltr:nx-ml-16 {
  display: none;
}
```

B2o5T:
or for advanced usage use follow function https://github.com/nodejs/nodejs.org/blob/e7251994629c9fa0553216e74f03620753df4ea4/next.config.mjs#L11-L18
to filter `pageOpts.headings` to exclude headings with depth 5 and 6

PR #1780, opened 2023-04-20
Feature Request: add `depth` prop to TOC headingComponent

jvzaniolo:
# Add `depth` prop to TOC headingComponent

I would like to display only the headings from depth 1 to 4 in the TOC component.

```jsx
// ./theme.config.tsx

export default {
  toc: {
    headingComponent({ depth, children }) {
      //               ^? Property 'depth' does not exist on type '{ id: string; children: string; }'.
      return depth < 5 ? <>{children}</> : null
    },
  },
}
```

## Suggested solution

- Provide the `depth` prop to the `headingComponent`
- Renders the default `headingComponent` when `config.toc.headingComponent` is undefined
- Renders the custom `config.toc.headingComponent` when its return value is truthy
- Renders nothing when the `config.toc.headingComponent` return value is falsy

### DIFF
--- a/.changeset/eleven-kids-clap.md
+++ b/.changeset/eleven-kids-clap.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': minor
+---
+
+Add `depth` prop to TOC `headingComponent`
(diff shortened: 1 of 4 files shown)


B2o5T:
https://github.com/shuding/nextra/issues/1619#issuecomment-1464495932

PR #1790, opened 2023-04-24
Replace nx-overflow-x-hidden by nx-overflow-clip

### LINKED ISSUE #1481: Nextra v2.2.13 breaks position sticky of elements
In nextra v 2.2.13  a `overflow-x: hidden` was added (https://github.com/shuding/nextra/commit/fd020ca9d4834bb4ffa7ebdace2cb9d221b143c6#diff-8310a54b1ec9da38ed7c74e8c1328e33f07848d25bd9779811614642e3268747R51) which breaks `sticky` elements.

Ran into it with CodeHike's Scrollycoding (https://github.com/code-hike/codehike/issues/322). See CodeHike's issue for repro sandbox.

madhavKD:
Fix #1481

### DIFF
--- a/.changeset/violet-penguins-compare.md
+++ b/.changeset/violet-penguins-compare.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+remove `nx-overflow-x-hidden` from main content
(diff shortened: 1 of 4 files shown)


B2o5T (approved):
just tested that we can even remove `nx-overflow-x-hidden` and keep only `nx-break-words`

PR #1828, opened 2023-05-06
Improve installation instructions

bmuenzenmeyer:
Running nextra for the first time locally. 

The installation command is incorrect. Running `pnpm` itself presents the help text for pnpm.

image

----

I also removed `just simply` as not inclusive words. See https://bradfrost.com/blog/post/just/ for some more thoughts on this


B2o5T (approved):
Thanks!

Issue #1762, opened 2023-04-16
Upgrading from 2.3.0 → 2.4.0 breaks build due to updated path joining


terrymun:
I noticed that after upgrading from Nextra@2.3.0 → Nextra@2.4.0, attempts to build or export my Nextra docs site will fail. The way I use Nextra is not very well documented, as it actually lives as a package in a Nx-managed monorepo, in the folder `packages/docs`, for example.

Attempting to run the build / export command now will cause webpack to throw an error:

```text
./pages/my-page.mdx
Error: ENOENT: no such file or directory, open '[monorepo_root]/packages/docs/packages/docs/pages/my-page.mdx'
```

From the error message above, you can see that attempts to access `my-page.mdx` has failed because the subpath `packages/docs` (the package root in my monorepo) has been duplicated in the path:

```text
'[monorepo_root]/packages/docs/packages/docs/pages/my-page.mdx'
                 ^^^^^^^^^^^^^ ^^^^^^^^^^^^^
                               (duplication)
```

This error does not show up when I downgrade to Nextra@2.3.0.

For a little context, this is how my `next.config.mjs` looks like:

```mjs
import nextra from 'nextra';

const withNextra = nextra({
  theme: 'nextra-theme-docs',
  themeConfig: './theme.config.tsx',
});

/**
 * @type {import('@nrwl/next/plugins/with-nx').WithNxOptions}
 **/
const nextConfig = {
  images: {
    unoptimized: true,
  },
  experimental: {
    // NOTE: This includes files from the monorepo base
    externalDir: true,
  },
};

export default withNextra(nextConfig);
```

I am no sleuth, but I suspect this might be due to how Nextra reads relative file paths with logic that is slightly tweaked in 2.4.0: [changelog](https://github.com/shuding/nextra/compare/nextra-theme-docs%402.3.0...nextra%402.4.0#diff-b80cbbc7df15b07dd7dadbe8fdf2427e2741c1b9ba8fa65daf3841acb5b41632), namely in `packages/nextra/src/plugin.ts`

B2o5T:
Idk why it occurs only for you, pr with fix welcome

simoneldevig:
@B2o5T I have a similar issue and created a reproduction at https://github.com/simoneldevig/nextra-nx-repro 

Run `yarn` and `yarn serve:docs` to see the issue.

[...]

PR #1837, opened 2023-05-09
fix `Error: ENOENT: no such file or directory` in Nx-managed monorepos

B2o5T:
fixes https://github.com/shuding/nextra/issues/1762

### DIFF
--- a/packages/nextra/src/loader.ts
+++ b/packages/nextra/src/loader.ts
@@ -100,10 +100,11 @@ async function loader(
   const mdxPath = (
     context._module?.resourceResolveData
       ? // to make it work with symlinks, resolve the mdx path based on the relative path
-        path.join(
-          context.rootContext,
-          context._module.resourceResolveData.relativePath
-        )
+        /*
+         * `context._module.resourceResolveData.relativePath` could include path chunk of
+         * `context.rootContext` use `context._module.rawRequest` instead
+         */
+        path.join(context.rootContext, context._module.rawRequest)
       : context.resourcePath
   ) as MdxPath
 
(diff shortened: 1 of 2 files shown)


packages/nextra/src/loader.ts
```diff
@@ -100,10 +100,11 @@ async function loader(
...
       ? // to make it work with symlinks, resolve the mdx path based on the relative path
-        path.join(
-          context.rootContext,
-          context._module.resourceResolveData.relativePath
-        )
+        /*
+         * `context._module.resourceResolveData.relativePath` could include path chunk of
+         * `context.rootContext` use `context._module.rawRequest` instead
+         */
+        path.join(context.rootContext, context._module.rawRequest)
```
B2o5T, line 107: cc @lttb

lttb: hey! thanks for addressing it. I saw your latest change using `descriptionFileRoot`, and it makes a lot of sense but it seems that it might not be presented in some cases 🤔 I was investigating some similar issues recently and came up with this workaround, but it looks a bit clumsily, so I was hesitating to open a PR with these changes to be honest 😅

```ts
const mdxPath = context._module?.resourceResolveData
    ? // to make it work with symlinks, resolve the mdx path based on the relative path
      context._module.resourceResolveData.path ||
      context._module.resourceResolveData.descriptionFileRoot
      ? path.join(
          context._module.resourceResolveData.descriptionFileRoot,
          context._module.resourceResolveData.relativePath,
        )
      : path.join(
          context.rootContext,
          context._module.resourceResolveData.relativePath,
        )
    : context.resourcePath
```

PR #1845, opened 2023-05-11
fix(blog): Use tailwind prose for small screens

tristandubbeld:
I started using the blog theme and noticed that some styles were missing on smaller screens. Things like `blockquote`, headings and `a` elements weren't properly styled. After digging into it a bit I found that the `prose` class wasn't applied to all breakpoints, just `sm`. Adding the `nx-prose` class to the basic layout fixed things.

I walked through the blog example after the change and confirmed that the styles were applied and nothing else was affected. If necessary I can add some before and after screenshots.

### DIFF
--- a/packages/nextra-theme-blog/src/basic-layout.tsx
+++ b/packages/nextra-theme-blog/src/basic-layout.tsx
@@ -1,5 +1,5 @@
 import Head from 'next/head'
-import type { ReactNode} from 'react';
+import type { ReactNode } from 'react'
 import { useRef } from 'react'
 import { useBlogContext } from './blog-context'
 import { HeadingContext } from './mdx-theme'
@@ -10,7 +10,7 @@ export const BasicLayout = ({ children }: { children: ReactNode }) => {
   const ref = useRef<HTMLHeadingElement>(null)
   return (
     <article
-      className="nx-container nx-prose-sm dark:nx-prose-dark md:nx-prose"
+      className="nx-container nx-prose nx-prose-sm dark:nx-prose-dark md:nx-prose"
       dir="ltr"
     >
       <Head>
--- a/packages/nextra-theme-blog/src/basic-layout.tsx
+++ b/packages/nextra-theme-blog/src/basic-layout.tsx
@@ -9,7 +9,7 @@ export const BasicLayout = ({ children }: { children: ReactNode }) => {
   const title = `${opts.title}${config.titleSuffix || ''}`
   const ref = useRef<HTMLHeadingElement>(null)
   return (
-    <article className="nx-container nx-prose dark:nx-prose-dark" dir="ltr">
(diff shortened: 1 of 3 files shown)


packages/nextra-theme-blog/src/basic-layout.tsx
```diff
@@ -10,7 +10,7 @@ export const BasicLayout = ({ children }: { children: ReactNode }) => {
   const ref = useRef<HTMLHeadingElement>(null)
   return (
     <article
-      className="nx-container nx-prose-sm dark:nx-prose-dark md:nx-prose"
+      className="nx-container nx-prose nx-prose-sm dark:nx-prose-dark md:nx-prose"
```
B2o5T, line 13: Yeah mistake is that `prose` class should be always included, we can also simplify
```suggestion
      className="nx-container nx-prose max-md:nx-prose-sm dark:nx-prose-dark"
```
In this case, this line should be updated too https://github.com/shuding/nextra/blob/6f3d3f06657cb434b7514394e2a81d4d7045f4f7/packages/nextra-theme-blog/src/styles.css#L24

could you also include a changeset?

tristandubbeld: Thanks for the quick reply! I noticed some `max-w` changes to the page when altering `md:nx-prose` so I didn't want to touch it initially, but I'll see what I can do. I will include a changeset.

packages/nextra-theme-blog/src/basic-layout.tsx
```diff
@@ -9,10 +9,7 @@ export const BasicLayout = ({ children }: { children: ReactNode }) => {
   const title = `${opts.title}${config.titleSuffix || ''}`
   const ref = useRef<HTMLHeadingElement>(null)
   return (
-    <article
-      className="nx-container nx-prose-sm dark:nx-prose-dark md:nx-prose"
-      dir="ltr"
-    >
+    <article className="nx-container nx-prose dark:nx-prose-dark" dir="ltr">
```
tristandubbeld, line 12: @B2o5T I've removed the `nx-prose-sm` class completely, as it didn't do anything in particular concerning breakpoints. Instead I've added the styles to hide backticks from inline code to the `nx-prose` class. If they're supposed to be shown on bigger breakpoints I can update the class, but tbh I like the code without backticks.

B2o5T, line 12:
```suggestion
    <article className="nx-container nx-prose max-md:nx-prose-sm dark:nx-prose-dark" dir="ltr">
```

B2o5T: > as it didn't do anything in particular concerning breakpoints

it did text small on small breakpoints

tristandubbeld: You're right, had a brainfart and didn't realize `prose-sm` was a tailwind class too. Thanks for adding it.

PR #1858, opened 2023-05-14
fixed accessability issue in light mode for nav bar

filiphuhta:
The current blog theme nav bars contrast is insufficient between the background and foreground colors for the `<span>` element in light mode. It now passes WCAG standards according to AA and AAA.


filiphuhta:
The current blog theme nav bars contrast is insufficient between the background and foreground colors for the` <span> `element in **light mode**.  It now passes WCAG standards according to AA and AAA.

B2o5T (approved):
thank you!

PR #1894, opened 2023-05-23
fix: can't build package on windows

AkaraChen:
On windows, `fast-glob` return path use '/' as separator, but esbuild use '\\', causing the esbuild plugin `add-mjs` can't identify client import, so the `nextra` package build output client script as `.mjs`:

image

### DIFF
--- a/packages/nextra/tsup.config.ts
+++ b/packages/nextra/tsup.config.ts
@@ -14,6 +14,8 @@ const CLIENT_ENTRY = [
 const entries = fg.sync(CLIENT_ENTRY, { absolute: true })
 const entriesSet = new Set(entries)
 
+const winPath = (p: string) => p.replace(/\\/g, '/')
+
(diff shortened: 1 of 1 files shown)


packages/nextra/tsup.config.ts
```diff
@@ -14,6 +14,8 @@ const CLIENT_ENTRY = [
 const entries = fg.sync(CLIENT_ENTRY, { absolute: true })
 const entriesSet = new Set(entries)
 
+const winPath = (p: string) => p.replace(/\\/g, '/')
```
B2o5T, line 17: Could you use `slash` package for this case (he is already installed in nextra)

AkaraChen: > Could you use `slash` package for this case (he is already installed in nextra)

Edited

B2o5T (approved):

Issue #454, opened 2022-05-03
[Bug] NX / Monorepo : Missmatch using `rootDir`, NextJS can't find  `pages` folder


Thisisjuke:
Hello, 

I have seen this thread : Not working inside a monorepo https://github.com/shuding/nextra/issues/151
The fix is working well for dev purposes !

However, lot of monorepos intruce build enhancement functionalities like : 
- https://nx.dev/next/overview
- https://nx.dev/packages/next/executors/build

Nextra is working well with NX except for path resolution at build time : there is a miss-match / side-effect with relative paths. 

> tsconfig.json
```.json
{
  "extends": "../../tsconfig.base.json",
  "compilerOptions": {
    "rootDir": ".",
    //...
  },
  "include": [...],
  "exclude": [...]
}
```
```shell
info  - Checking validity of types  
(node:8515) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, scandir '/home/projects/workspace/pages'
```
Where the real path should be : ` "home/projects/workspace/apps/nextra-doc/pages")

So i add the right `rootDir`:
> tsconfig.json
```.json
{
  "extends": "../../tsconfig.base.json",
  "compilerOptions": {
    "rootDir": "./apps/nextra-doc",
    //...
  },
  "include": [...],
  "exclude": [...]
}
```
```shell
Failed to compile.
Type error: File '/home/projects/workspace/apps/nextra-doc/next.config.js' is not under 'rootDir' '/home/projects/workspace/apps/nextra-doc/apps/nextra-doc'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Root file specified for compilation
```
The path in the tsconfig.json is applied mandatory to build, but when used, its applied 2 times.

These relative paths and other things are handle by : 
```js
module.exports = withNx(nextConfig)
```

The best solution is that you let the possibility to use `withNx` with `withNextra` : 
You let NX generate the appropriate NextConfig, and then, you appy your own loader + mdx parser 

As this a chance to be fixed a day or not at all (outside the scope or else) ?

**I can provide a repository with all of this setup if you want to test or fix this issue !**

Thx for your time :v: !

Thisisjuke:
Hello there,

I have fixed some issues with `next-compose-plugins` and the following setup :
```
const withNx = require('@nrwl/next/plugins/with-nx')
const withPlugins = require('next-compose-plugins');

const withNextra = require('nextra')({
  theme: 'nextra-theme-docs',
[...]

PR #1900, opened 2023-05-24
fix: project root path error in nx

StringKe:
A simple attempt to fix the error only, without verifying that the rest of the functionality is correct.

new `next.config.js`
```ts
//@ts-check
// nextra need root dir
process.env.NEXT_ROOT = __dirname

// eslint-disable-next-line @typescript-eslint/no-var-requires
const { composePlugins, withNx } = require('@nx/next')
const withNextra = require('nextra')

/**
 * @type {import('@nx/next/plugins/with-nx').WithNxOptions}
 **/
const nextConfig = {
  nx: {
    // Set this to true if you would like to use SVGR
    // See: https://github.com/gregberge/svgr
    svgr: false,
  },
}

const plugins = [
  // Add more Next.js plugins to this list if needed.
  withNx,
  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
  // @ts-ignore
  withNextra({
    theme: 'nextra-theme-docs',
    themeConfig: './theme.config.tsx',
    rootDir: __dirname,
  }),
]

module.exports = composePlugins(...plugins)(nextConfig)

```

### DIFF
--- a/packages/nextra/src/file-system.ts
+++ b/packages/nextra/src/file-system.ts
@@ -7,7 +7,8 @@ const { findPagesDir, existsSync } = getDefault(findPagesDirImport)
 export { existsSync }
 
 export function findPagesDirectory(): string {
-  const res = findPagesDir(CWD, false)
+  const rootDir = process.env.NEXT_ROOT || CWD
+  const res = findPagesDir(rootDir, false)
   return (
     res.pagesDir || // next v13
     (res as any).pages // next v12


StringKe:
https://github.com/shuding/nextra/issues/454

StringKe:
https://github.com/shuding/nextra/issues/454 
This error can be resolved and has been verified locally.

B2o5T:
What's process.env.NEXT_ROOT?

seems like you are introduced this env variable and I don't like it

StringKe:
too many variables in nextra are defined by way of constants, and changing to config parameter passing would require too many changes, and I'm not sure I can finish it.

StringKe:
https://github.com/nrwl/nx/pull/17205/files#diff-6abc0d30cb71c3db586bfa9cb55cac2d6952705ce0d528b6e142a43d1e7dff88R36

It seems that in the new nx version a new environment variable has been added to similarly identify the root path.

B2o5T:
@StringKe could you try locally modify this line
https://github.com/shuding/nextra/blob/61487906f8f498df88e74edf269e382e33209221/packages/nextra/src/constants.ts#L27

By
` const CWD = process.env.NX_NEXT_DIR ?? process.cwd()`

StringKe:
does not work, require already requires process.env.NX_NEXT_DIR.

NX_NEXT_DIR = __dirname. nx needs to be set in dev server startup

marcosluizfp:
I made this workaround and it worked for me:

https://github.com/shuding/nextra/issues/454#issuecomment-1907150823

PR #2045, opened 2023-07-07
fix: replace last 'export default MDXContent;'

soata:
Fix to replace the last 'export default MDXContent;' instead of the first one.

If .mdx file contains a string that exactly matches 'export default MDXContent;', the certain export is not removed and it will result in an error.

### DIFF
--- a/packages/nextra/src/loader.ts
+++ b/packages/nextra/src/loader.ts
@@ -321,7 +321,7 @@ const __nextraPageOptions = {
         (themeConfigImport && 'themeConfig: __nextra_themeConfig')
   }
 }
-${finalResult.replace('export default MDXContent;', '')}
+${finalResult.replace(/(export default MDXContent;).*$/, '$1')}
 if (process.env.NODE_ENV !== 'production') {
   __nextraPageOptions.hot = module.hot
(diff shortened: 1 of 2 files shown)


packages/nextra/src/loader.ts
```diff
@@ -321,7 +321,7 @@ const __nextraPageOptions = {
         (themeConfigImport && 'themeConfig: __nextra_themeConfig')
   }
 }
-${finalResult.replace('export default MDXContent;', '')}
+${finalResult.replace(/(export default MDXContent;).*$/, '$1')}
```
B2o5T, line 324: Nice catch, just one suggestion that we need only replace a last appearance and all code that comes after
image
maybe some solution with `String#lastIndexOf` and `String#slice`

could You add also a helpful comment in code why we did it and could you add changeset with `patch`?

soata: Thank you for the review.

I apologize for pushing the wrong regex code.
What I intended was 
`${finalResult.replace(/export default MDXContent;(.*)$/, '$1')}`

I used regex to minimize the changes, but I'll think about using `String#lastIndexOf`. 

> could You add also a helpful comment in code why we did it and could you add changeset with patch?

I'll try it later!

PR #2046, opened 2023-07-07
WIP es5

denisoster:
WIP
Added es5 to support older devices
Removed redundant code

Iphone 6 SSG - correct worked


B2o5T (changes requested):
Please do not introduce such as changes without creating issue first

Issue #1878, opened 2023-05-17
[Nextra-Blog] A mdx page that is hidden is visible as a post


md-weber:
I created a _meta.json with the following content:

```json
{
    "impressum": {
        "hidden": true,
        "title": "Impressum",
        "type": "page"
    }
}
```

As expected it gets removed from the sidebar and I cannot directly navigate to it anymore. However we can see it now in the posts page.

<img src="https://github.com/shuding/nextra/assets/8026644/4fc6b5ff-90dc-46cd-aecd-b9da9529a053" height=200/>

I tried already the following: Put the `_meta.json` into the MDX header part then it gets removed from the post part but we it is again visible in the sidebar.

```md
---
hidden: true
title: Impressum
type: page
---
```

And also I tried to use instead of `hidden: true` -> `display: hidden` with the same result.

This could be related to https://github.com/shuding/nextra/issues/1729

md-weber:
I believe I found the issue in the `collect.ts`, here we have the part 
```
const isNav = (page: PageMapItem): page is MdxFile => {
  const type = 'frontMatter' in page && page.frontMatter?.type
  return type && ['page', 'posts'].includes(type)
}
```
Which does not take the `"display": "hidden"` into consideration. I will work further on this issue and hope I can provide next week a PR for it.

B2o5T:
Only docs theme support _meta.json, possible way to hide described here https://github.com/shuding/nextra/issues/351

Issue #351, opened 2022-02-01
[nextra-theme-blog] Allow for a draft post


rgoomar:
Currently, you can get around this by putting a `_` as part of the `.mdx` file although, I think it would be helpful to have a metadata field like:
```
---
type: post
draft: true
---
```
That way, it won't display in the posts list although still have a link that you can send out to people to review that will remain the same after publishing.

YikSanChan:
@rgoomar 

> you can get around this by putting a _ as part of the .mdx file although

Do you mind sharing more details on how to make it work? Thx!

rgoomar:
@YikSanChan Basically, if you have an `_`, it ignores it based on this line https://github.com/shuding/nextra/blob/core/packages/nextra-theme-blog/src/index.tsx#L126

So you can link to it as a post, but doesn't appear in the list.

PR #2049, opened 2023-07-07
Support draft/hidden posts

kevinwuhoo:
A tiny bit of code to support draft or hidden posts. Also added an example post to demonstrate the behavior.

The name of the key that triggers this behavior is up for debate. I thought `draft` was a nice way to signal this. `hidden` is a bit confusing because the post isn't completely hidden as you can access it if you go to the url directly.

Closes https://github.com/shuding/nextra/issues/351
Closes https://github.com/shuding/nextra/issues/1878

Edit: Thinking about this more I think the "right" way to do this is to create a branch with the draft post and use Vercel's preview deployment instead of polluting your production deployment with non-finalized posts. Though I think this is still a good feature to have around.


mptorz (approved):
This would be really useful

B2o5T (approved):

PR #2104, opened 2023-07-25
fix: footnote styles, backlink jump

### LINKED ISSUE #2097: Footnotes <h2 /> is unstyled in docs theme
To reproduce, put this in `index.mdx` in the docs theme

```
# Testing

Footnote. [^1]

[^1]: The footnote value.
```

Will look like this:
Screenshot 2023-07-21 at 6 35 54 AM

Inspected element:
Screenshot 2023-07-21 at 6 37 12 AM

I think the reason this isn't a problem in the blog theme is that there, the subheading styles come from global CSS, but the docs subheadings rely on classnames (why is this? from what I can tell it is always the same):

https://github.com/shuding/nextra/blob/aa69c7f5dd9b0029e8110ba8f3e311c089ea5dcf/packages/nextra-theme-docs/src/mdx-components.tsx#L51

I believe the footnotes component is processed as a different component from `<h2 />`, so it won't be captured here:
https://github.com/shuding/nextra/blob/aa69c7f5dd9b0029e8110ba8f3e311c089ea5dcf/packages/nextra-theme-docs/src/mdx-components.tsx#L191

I can contribute a PR if let me know which is preferred:
* Style headings in global CSS
* Add footer section to return type of `mdx-components.tsx` if possible (I don't know off the top of my head if it is exposed by the mdx-js API - EDIT: it is [possible but nontrivial](https://markmichon.com/advanced-custom-mdx-components)). (Might also worth it to style some things in there besides the header - such as shrinking and dimming the text, and [adding an `<hr />`](https://github.com/shuding/nextra/issues/802))

llllvvuu:
resolves #2097 

also removes manual scroll-margin-top hack that applied only to subheading anchors specifically

before:

https://github.com/shuding/nextra/assets/5601392/0a8e2896-6c1c-4750-9510-fcb01376c1a0

after:

https://github.com/shuding/nextra/assets/5601392/2d3b3f0c-44ee-4141-ace0-2b372a75b166


suica:
Same issue here. 👍🏿

B2o5T (approved):
thx for the tip with `scroll-padding` property, btw `## Footnotes` should be appeared  only for screen readers see at the bottom (for this reason he has `sr-only` class) https://github.com/gatsbyjs/gatsby/blob/master/examples/using-remark/src/pages/2016-04-15---hello-world-kitchen-sink/index.md#user-content-fn-1-6a1ee311dd33d198f817867c028b8293

llllvvuu:
> `## Footnotes` should be appeared only for screen readers see at the bottom (for this reason he has `sr-only` class)

Great catch 🔥

PR #2106, opened 2023-07-25
fix: default config to exclude `ERROR_ROUTES` from search and posts

### LINKED ISSUE #2099: Search: 404 Appears in Results
It appears that all pages are indexed, including built-in overrides in Next apps (404, 50x, etc). This means that we see 404 in search results, including in the main Nextra docs site:
image

Perhaps there's a way to either: 
* hook into known overridable Next templates and auto-ignore indexing these pages?
* provide an escape hatch for auto-indexing certain pages?

llllvvuu:
resolves #2099

* sets default `indexKey` for search to exclude `ERROR_ROUTES`
* adds `type: hidden-page` to blog theme
* adds `hiddenPages` globs to blog theme; if `type` is not specified, it will be `hidden-page` if matching `hiddenPages` else `post`
* sets default `hiddenPages` for blog to exclude `ERROR_ROUTES`

### DIFF
--- a/.changeset/happy-clocks-tease.md
+++ b/.changeset/happy-clocks-tease.md
@@ -1,0 +1,6 @@
+---
+"nextra-theme-blog": patch
+"nextra": patch
+---
+
+fix: default config to exclude /404 from search and posts
--- a/.changeset/happy-clocks-tease.md
+++ b/.changeset/happy-clocks-tease.md
@@ -3,4 +3,4 @@
 "nextra": patch
 ---
 
-fix: default config to exclude /404 from search and posts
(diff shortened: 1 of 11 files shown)


C-EO (approved):
Fine by me

B2o5T:
closes in favour https://github.com/shuding/nextra/pull/2181

Thanks for your attempt, but I will stay with an approach to exclude search indexing by default for 404/500 mdx pages, and about `nextra-theme-blog` - I am against adding an additional dependency for this case and also you can always use regular jsx/tsx page for 404/500 pages

PR #2151, opened 2023-08-11
use github-slugger for custom heading ids to prevent duplicated headings

guilherssousa:
As noticed by @uonr on his issue at [vercel/swr-site#530](https://github.com/vercel/swr-site/pull/530), same headings generate the same anchor ids. This PR adds checking of existent headings before pushing, and, if necessary, adds a prefix based on the closest lower depth heading.

Here's a comparision from the SWR's Mutation and Revalidation heading tree:

```
--- BEFORE ---

mutate #mutate
  Global Mutate #global-mutate
  Bound Mutate #bound-mutate
  Revalidation #revalidation
  API #api
    Parameters #parameters
    Return Values #return-values
  useSWRMutation #useswrmutation
    API #api <-- DUPLICATED ID
      Parameters #parameters <-- DUPLICATED ID
      Return Values #return-values <-- DUPLICATED ID
    Basic Usage #basic-usage
    Defer loading data until needed #defer-loading-data-until-needed

--- AFTER ---

mutate #mutate
  Global Mutate #global-mutate
  Bound Mutate #bound-mutate
  Revalidation #revalidation
  API #api
    Parameters #parameters
    Return Values #return-values
  useSWRMutation #useswrmutation
    API #useswrmutation-api
      Parameters #useswrmutation-parameters
      Return Values #useswrmutation-return-values
    Basic Usage #basic-usage
    Defer loading data until needed #defer-loading-data-until-needed
```

### DIFF
--- a/packages/nextra/src/mdx-plugins/remark-headings.ts
+++ b/packages/nextra/src/mdx-plugins/remark-headings.ts
@@ -46,6 +46,23 @@ export const remarkHeadings: Plugin<[], Root> = function (this: Processor) {
             id:
               (node.data?.hProperties as HProperties)?.id || slugger.slug(value)
           }
+
+          // check if heading.id already exists in headings list
+          const sameHeadingIdIndex = data.headingMeta.headings.findIndex(h => h.id == heading.id);
+
+          // no need to iterate on depth 1 headings
+          if(sameHeadingIdIndex !== -1 && heading.depth > 1) {
+            const oldHeading = data.headingMeta.headings[sameHeadingIdIndex]
+            // findLastIndex workaround: clone, rotate, findLastIndex (todo: refactor)
+            const lowerDepthHeadingIndex = data.headingMeta.headings
+              .slice(0)
+              .reverse()
+              .findIndex((heading) => heading.depth == oldHeading.depth - 1);
+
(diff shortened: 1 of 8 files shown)


packages/nextra/src/mdx-plugins/remark-headings.ts
```diff
@@ -46,7 +46,33 @@ export const remarkHeadings: Plugin<[], Root> = function (this: Processor) {
             id:
               (node.data?.hProperties as HProperties)?.id || slugger.slug(value)
```
B2o5T, line 47:
```suggestion
              slugger.slug((node.data?.hProperties as HProperties)?.id || value)
```
Thanks for your report and PR, but we already use github-slugger package which is responsible for adding suffixes for duplicated headings, I think it can be fixed by the above suggestions. Could you try it?

Also, it should be applicable to h1, since user can have multiple h1 in his page, even it’s not a best practice

guilherssousa: Oh, I didn't noticed the description of the slugger function, seems nice and convenient, I removed the code and made the suggested change, thanks for the help

B2o5T (approved):

Issue #2172, opened 2023-08-16
CSS style `font-weight: 500;` for code blocks probably not work


realfish:
The styles of `<pre>` and `<code>` elements in the Docs theme are specified below:

```css
font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, Liberation Mono, Courier New, monospace;
font-weight: 500;
```

Therefore, if `SFMono-Regular` is installed (for instance, through <https://developer.apple.com/fonts/>), the font of code blocks will be probably rendered in the regular weight (`400`) of SF Mono.

In other words, the `font-weight: 500;` may have no effect when `SFMono-Regular` is used to rendering the code block text. This issue can be reproduced by Chrome/Firefox on macOS.

PR #2184, opened 2023-08-19
remove `font-weight: 500;` from styles of code blocks since it gives no effect

B2o5T:
fixes https://github.com/shuding/nextra/issues/2172

### DIFF
--- a/.changeset/weak-poems-explode.md
+++ b/.changeset/weak-poems-explode.md
@@ -1,0 +1,7 @@
+---
+'nextra': patch
+'nextra-theme-blog': patch
+'nextra-theme-docs': patch
+---
+
+remove `font-weight: 500;` from styles of code blocks since it gives no effect
--- a/packages/nextra/src/components/pre.tsx
+++ b/packages/nextra/src/components/pre.tsx
@@ -36,7 +36,7 @@ export const Pre = ({
       )}
       <pre
         className={cn(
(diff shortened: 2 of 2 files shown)


.changeset/weak-poems-explode.md
```diff
@@ -0,0 +1,7 @@
+---
+'nextra': patch
+'nextra-theme-blog': patch
+'nextra-theme-docs': patch
+---
+
+remove `font-weight: 500;` from styles of code blocks since it gives no effect
```
realfish, line 7: Hope I didn’t confuse you guys with issue #2172.

The style `font-weight: 500;` *does take effect* in those browsers that support the generic family name `ui-monospace`, such as Safari (Mac/iOS). However, in browsers that don’t support the `ui-monospace` (Chrome, Firefox, etc.), the styles `font-weight: 500;` in combination with `font-family: …, SFMono-Regular, Menlo, …;` make no sense.

Hence, I’m not sure removing `font-weight: 500;` complies with the design/typographic spec of the Doc theme. The SF Mono family does include the 500 weight. (Obviously, fonts such as Menlo, Monaco, and Consolas also do not have a font weight of 500.)

Btw, a mistake I made in #2172 was that actually the PostScript name `SFMono-Regular` seems to have no effect for the Chrome/Firefox/Safari (at least on macOS 12.6+).

Issue #1719, opened 2023-04-02
Feature request: Let editLink prop to take `null` not to render editLink component.


enk0de:
I have a need not to render `editLink` component. Please change a specification of editLink prop to take `nullish` value.

tgrassl:
Right now you could set the editLink.component to null to hide it:
```
editLink: {
    component: null,
},
```
This will show a validation error, but still build the site without a problem.
<img width="559" alt="CleanShot 2023-04-15 at 22 37 28@2x" src="https://user-images.githubusercontent.com/34568407/232252223-e06fdd57-508b-4368-9e9d-6ef03d42fa8b.png">

PR #2221, opened 2023-08-28
Update theme-configuration.mdx

kgolubic:
Added the note about disabling the edit link.

fixes https://github.com/shuding/nextra/issues/1719
fixes https://github.com/shuding/nextra/issues/2289

### DIFF
--- a/.changeset/strong-terms-bake.md
+++ b/.changeset/strong-terms-bake.md
@@ -1,0 +1,6 @@
+---
+'nextra-theme-docs': patch
+'nextra': patch
+---
+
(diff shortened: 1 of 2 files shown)


arnabsen:
hi @shuding, can we add this to the docs?

Aruelius:
if we set `editLink.text` to null, the `a` element still render without text.

image

 however if set 'editLink.component' to null, it works, but raise an error: **Must be React.FC. Path: "editLink.component"**

set `feedback.content` to null works will.

PR #2271, opened 2023-09-08
update to actions/checkout@v4 instead of commit hash

ImBIOS:


B2o5T (approved):
ty, renovate did it for some reason... here https://github.com/shuding/nextra/pull/2243 can you update other workflows too? 💪

PR #2296, opened 2023-09-15
DocsGPT showcase example

dartpain:
Hi, I added an example on nextra showcase. We use it for our project [here](https://docs.docsgpt.co.uk/)

We recently migrated from github wiki to nextra


B2o5T (approved):
enjoy

Issue #2374, opened 2023-10-03
Display math does not work in `math` code brackets


siefkenj:
It seems that code brackets of the form

`````
```math
\int_1^2x^2
```
`````

will not render as display math, even though `remark-math` advertises them as [working](https://github.com/remarkjs/remark-math/tree/main/packages/remark-math#authoring). (Additionally, it seems there is a parse error unless `$$` for display math and the rest of the math are on the same line, which also contradicts `remark-math`'s documentation...)

I've temporarily worked around this issue by writing a small plugin:

```javascript
import { visit } from "unist-util-visit";

/**
 * Explicitly convert math code blocks to display-math nodes.
 */
function mathCodeBlocksToMath() {
    /**
     * @param {import('mdast').Root} tree
     *   Tree.
     * @returns {undefined}
     *   Nothing.
     */
    return function (tree) {
        console.log(tree);
        visit(tree, (node) => {
            if (node && node.type === "code" && node.lang === "math") {
                node.type = "displayMath";
                delete node.meta;
                delete node.lang;
                node.data = {
                    hName: "code",
                    hProperties: {
                        className: ["language-math", "math-display"],
                    },
                    hChildren: [{ type: "text", value: node.value }],
                }
            }
        });
    };
}

const withNextra = nextraConfig({
    theme: "nextra-theme-docs",
    themeConfig: "./theme.config.tsx",
    latex: true,
    mdxOptions: {
        remarkPlugins: [mathCodeBlocksToMath],
    },
});

export default withNextra();
```

It would be nice if this functionality was included by default :-).

B2o5T:
thx for repo, this was a bug, we need to use `rehypeKatex` before `rehypePrettyCode` otherwise `rehypePrettyCode` will do some changes for `math` code blocks

Issue #2368, opened 2023-10-01
TypeError: Cannot read properties of undefined (reading 'mathFlowInside')


albertilagan:
Broke on the latest release.

![image](https://github.com/shuding/nextra/assets/22500561/827d2772-886a-4a55-94eb-76452b7b54ba)

<img width="875" alt="image" src="https://github.com/shuding/nextra/assets/22500561/e063a6a8-e64a-4cc2-819a-d87a81b60039">

<img width="679" alt="image" src="https://github.com/shuding/nextra/assets/22500561/9e15f160-6f06-466c-b3ab-08ed6d6cd630">

B2o5T:
upstream issue for latex/katex

dsernst:
Hmm, seeing this too all of a sudden, and it's breaking and blocking builds.

Not seeing any references to `mathFlowInside` in https://github.com/KaTeX/KaTeX. Unclear how to proceed.

B2o5T:
Please share code snippet, I’ll debug

B2o5T:
@acdibble thx for repro

this issue was made after `remark-math` update to v6, seems like in the new version `$$` syntax isn't supported and `math` code block language should be used instead aka:

````md
```math
````

I downgraded `remark-math` to v5 to fix breaking changes, and will do an update of `remark-math`  only for Nextra v3.

so you'll need to do following refactor for Nextra v3 or right now
````diff
- $$
+ ```math
  x^2
- $$
+ ```
````

PR #2378, opened 2023-10-04
fix latex issues

B2o5T:
fixes https://github.com/shuding/nextra/issues/2374
fixes https://github.com/shuding/nextra/issues/2368


Josemoe (approved):
READ AND APPROVE

PR #2409, opened 2023-10-09
fix default locale issue

3ru:
There was a problem that if the `defaultLocale` was set to a value that did not exist in `PLACEHOLDER_LOCALES`(en-US, fr, ru, zh-CN), it would be undefined.

```js
export default withNextra({
  i18n: {
    locales: [
      'en',
    ],
    defaultLocale: 'en',
  },
})
```

image

I added a condition and modified it to fall back to `DEFAULT_LOCALE` if the user-specified `defaultLocale` is not present.


3ru:
ping @shuding

PR #2433, opened 2023-10-11
fix `*` theme settings for dynamic routes when route is not exist in page map

B2o5T:


packages/nextra/__test__/normalize-page.spec.ts
```diff
@@ -327,4 +327,55 @@ describe('normalize-page', () => {
       }
     `)
   })
+
+  it('should apply `*` settings even page is not exists', async () => {
```
B2o5T, line 331:
```suggestion
  it('should apply `*` settings even if page is not exist', async () => {
```

PR #2434, opened 2023-10-12
Add link to example source code for blog

benjie:
Took a moment to track this down, so figured it's worth adding to the docs?


benjie:
This is actually the wrong link; the content differs.

benjie:
Link corrected.

B2o5T (approved):
This example is a bit outdated, but anyway thank you @benjie ❤️

Issue #2424, opened 2023-10-11
Escape Sequence Bug when upgrading from Nextra v2 to v3 during SSG


NPJigaK:
Upon upgrading from Nextra v2 to Nextra v3, I encounter an error regarding legacy octal escape sequences not being permitted in strict mode while performing SSG.

```
./pages/test/aaa.mdx
Error:
  × Legacy octal escape is not permitted in strict mode
    ╭─[D:\testapp\pages\test\aaa.mdx:18:1]
 18 │
 19 │ const hoc = HOC_MDXWrapper(
 20 │   MDXLayout,
 21 │   '/test\aaa',
    ·       ──
 22 │   {"filePath":"pages/test/aaa.mdx",pageMap,frontMatter,title},
 23 │   typeof RemoteContent === 'undefined' ? useTOC : RemoteContent.useTOC
 24 │ )
```

NPJigaK:
https://github.com/shuding/nextra/pull/2435

Debugged locally and created a PR to resolve.
This problem can be solved with a fix here.

PR #2435, opened 2023-10-12
Uniform Path Separators using slash() Method

NPJigaK:
fix: https://github.com/shuding/nextra/issues/2424

This PR resolves a path separator inconsistency issue on different operating systems when creating the `route` constant. On Windows, `path.relative()` produces backslashes (`\`), while on Unix-based systems it produces forward slashes (`/`). This inconsistency can lead to unexpected behavior.

The solution is to use the `slash()` method to ensure all path separators are standardized to forward slashes (`/`). This change ensures consistent path formatting across all operating systems.

Changes proposed:
1. Use the `slash()` method when defining the `route` constant to convert any backslashes to forward slashes.

### DIFF
--- a/packages/nextra/src/server/loader.ts
+++ b/packages/nextra/src/server/loader.ts
@@ -131,12 +131,12 @@ ${themeConfigImport && '__nextra_internal__.themeConfig = __themeConfig'}`
   const locale =
     locales[0] === '' ? '' : mdxPath.replace(PAGES_DIR, '').split('/')[1]
 
-  const route = slash(
+  const route =
     '/' +
-    path
-      .relative(PAGES_DIR, mdxPath)
+    slash(path
+      .relative(PAGES_DIR, mdxPath))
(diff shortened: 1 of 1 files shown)


packages/nextra/src/server/loader.ts
```diff
@@ -131,12 +131,12 @@ ${themeConfigImport && '__nextra_internal__.themeConfig = __themeConfig'}`
   const locale =
     locales[0] === '' ? '' : mdxPath.replace(PAGES_DIR, '').split('/')[1]
 
-  const route =
+  const route = slash(
     '/' +
     path
       .relative(PAGES_DIR, mdxPath)
```
B2o5T, line 137:
```suggestion
    slash(path
      .relative(PAGES_DIR, mdxPath))
```
we don't need wrap `.replace(..)` with slash

NPJigaK:
@B2o5T 
Review comments addressed.

PR #2440, opened 2023-10-12
MathJax Support

siefkenj:
This PR adds an option for using MathJax as the math renderer. It allows the `latex` config option to be an object with a `renderer: "mathjax" | "katex"` object.

There are a couple of questions I had:
 1. Every page that contains math needs to be wrapped in a `<MathJaxContext>` object. I wasn't sure how to do that, so I just manually added a `<MathJaxContext>` to `latex.mdx`. That should only be a temporary workaround.
 2. I wasn't sure how to get `zod` to use an existing typescript type. It would be nice if `{renderer: "mathjax", options: MathJaxOptions}` was one of the types for `latex: ...`. If so, the user could configure mathjax paths manually if they want. The default is to load mathjax from a CDN.


B2o5T (commented):
Thanks, why you didn’t use `rehype-mathjax`? did you encounter some blocking problems with it?

siefkenj:
> Thanks, why you didn’t use `rehype-mathjax`? did you encounter some blocking problems with it?

`rehype-mathjax` is meant mainly for static rendering, and the accessibility menu isn't available on statically rendered mathjax. If `rehype-mathjax` is used with dynamic rendering, it just replaces each math element with `$...$` and expects the user to call mathjax on the whole page. That doesn't work well with dynamic content created via React.

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+      const result: Element[] = [
+        {
+          type: 'element',
+          tagName: 'MathJax',
+          children: [{ type: 'text', value: bracketedValue }],
+          properties: displayMode ? {} : { inline: true }
+        }
+      ]
+      if (!displayMode) {
+        result[0].properties.inline = true
```
B2o5T, line 80: Redundant, already covered by line 76

packages/nextra/src/client/mdx.ts
```diff
@@ -3,15 +3,19 @@ import {
...
+import type { ComponentProps } from 'react'
 
 const DEFAULT_COMPONENTS = {
   img: props =>
     createElement(
       typeof props.src === 'object' ? Image : 'img',
       props as ImageProps
-    )
+    ),
+  MathJax: (props: ComponentProps<typeof MathJax>) =>
```
B2o5T, line 17: I would prefer to add import ast node directly in your rehype plugin, if you don’t know how to do it, I’ll do it for you when I’ll review your PR

examples/docs/src/pages/features/latex.mdx
```diff
@@ -21,7 +25,14 @@ For example, the following code
 $\sqrt{a^2 + b^2}$
 ```
 
-will be rendered as: $\sqrt{a^2 + b^2}$
+will be rendered as: $\sqrt{a^2 + b^2}$.
+
+You can also do display math:
+```math
+\int_2^3x^3\,\mathrm{d}x
```
B2o5T, line 32: I would prefer to add an example of mathjax usage in root/docs directory (source code of Nextra’s docs). These examples folder is just a playground

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+ *   Transform.
+ */
+export function rehypeReactMathjax(options: Options) {
+  const _settings = options || emptyOptions
+
+  /**
+   * Transform.
+   */
+  return function (tree: Root, _file: VFile) {
+    visitParents(tree, 'element', function (element, parents) {
```
B2o5T, line 29: I think we can avoid installing `visitParents` dependency, and visit tree by matching `pre` tag, if you don't want to do such as refactoring I'll do it for you

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+      const classes = Array.isArray(element.properties.className)
```
B2o5T, line 30: We can use `Set` here for classes instead of array

packages/nextra/src/server/compile.ts
```diff
@@ -153,6 +154,22 @@ export async function compileMdx(
...
 
+  const mathPlugin: PluggableList = []
+  if (latex) {
+    if (typeof latex === 'object') {
+      switch (latex.renderer) {
+        case 'katex':
+          mathPlugin.push(rehypeKatex)
+          break
+        case 'mathjax':
+          mathPlugin.push([rehypeReactMathjax, latex.options])
```
B2o5T, line 165: Can you mention any example of which options we can pass? 🙏 so i could test it too

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+        ? element.properties.className
+        : emptyClasses
+      // This class can be generated from markdown with ` ```math `.
+      const languageMath = classes.includes('language-math')
```
B2o5T, line 34:
```suggestion
      const hasLanguageMath = classes.includes('language-math')
```

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
+import type { Element, Root } from 'hast'
+import { toText } from 'hast-util-to-text'
+import { SKIP, visitParents } from 'unist-util-visit-parents'
+import type { VFile } from 'vfile'
```
B2o5T, line 4: You can remove this dependency and take a look how type annotation is declared for other nextra’s remark/rehype plugin

B2o5T (commented):
Did quick review from mobile, hope to test it this week, thank you for your work 😍

examples/docs/src/pages/features/latex.mdx
```diff
@@ -1,3 +1,7 @@
+import {MathJaxContext} from "better-react-mathjax"
+
+<MathJaxContext>
```
B2o5T, line 3: Tip to wrap your mdx file with this component you can go to mdx2 playground put this code and see how ast nodes should looks like, if you’ll not succeed I’ll help with it

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      let parent = parents[parents.length - 1]
+      let scope = element
+
+      // If this was generated with ` ```math `, replace the `<pre>` and use
+      // display.
+      if (
+        element.tagName === 'code' &&
+        languageMath &&
```
B2o5T, line 53:
```suggestion
```
Redundant, this variable is always truthy, according to check on line 42

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+   * Transform.
+   */
+  return function (tree: Root, _file: VFile) {
+    visitParents(tree, 'element', function (element, parents) {
+      const classes = Array.isArray(element.properties.className)
+        ? element.properties.className
+        : emptyClasses
+      // This class can be generated from markdown with ` ```math `.
+      const languageMath = classes.includes('language-math')
+      // This class is used by `remark-math` for flow math (block, `$$\nmath\n$$`).
```
B2o5T, line 35: I think support of $$ was removed in remark-math v6? https://github.com/shuding/nextra/issues/2368#issuecomment-1747752925

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+ * @returns
+ *   Transform.
+ */
+export function rehypeReactMathjax(options: Options) {
+  const _settings = options || emptyOptions
+
+  /**
+   * Transform.
+   */
+  return function (tree: Root, _file: VFile) {
```
B2o5T, line 28:
```suggestion
  return function (tree: Root) {
```
Since not used in plugin

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Render elements with a `language-math` (or `math-display`, `math-inline`)
+ * class with KaTeX.
+ *
+ * @param {Readonly<Options> | null | undefined} [options]
+ *   Configuration (optional).
+ * @returns
+ *   Transform.
+ */
+export function rehypeReactMathjax(options: Options) {
+  const _settings = options || emptyOptions
```
B2o5T, line 23:
```suggestion
```
You don’t use settings (options) so why do we need to allow plugin to pass it? For the future ?

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      /* c8 ignore next -- verbose to test. */
+      if (!parent) {
+        return
+      }
+
+      const value = toText(scope, { whitespace: 'pre' })
+      const bracketedValue = displayMode ? `\\[${value}\\]` : `\\(${value}\\)`
+
+      const result: Element[] = [
```
B2o5T, line 71:
```suggestion
      const result: Element = {
```
no need to be an array

packages/nextra/src/server/compile.ts
```diff
@@ -46,6 +46,7 @@ import {
   remarkStructurize
 } from './mdx-plugins/index.js'
 import { rehypeExtractTocContent } from './mdx-plugins/rehype-extract-toc-content.js'
+import { rehypeReactMathjax } from './mdx-plugins/rehype-mathjax-react.js'
```
B2o5T, line 49:
```suggestion
import { rehypeBetterReactMathJax } from './mdx-plugins/rehype-better-react-mathjax.js'
```
Let’s rename to above to be consistent with the used dependency better-react-mathjax

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
siefkenj: I should have mentioned, this file is copied and modified from `rehype-katex`. I can refactor if you like.

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
siefkenj: I can change it, but it's probably overkill? I am not sure when sets become faster, but if most elements have 2 or 3 classes, I think arrays are actually faster.

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
siefkenj: I think it would be good to still support this since someone may modify the parser and insert `display-math` tags manually.

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
siefkenj: I don't think so... `langaugeMath = false`, `mathDisplay = true`, `mathInline = false` would pass the check on line 42, but not on line 53.

B2o5T: Yeah, here I did mistake

examples/docs/src/pages/features/latex.mdx
siefkenj: I assume I cannot just wrap the whole page since it may have `import` statements at the beginning? Do you have an example where just the body is wrapped in a component?

packages/nextra/src/client/mdx.ts
siefkenj: Is there a place in the code you could point me to where this is done. I looked at the AST for an import statement, and it is fairly long...surely you don't mean for me to hardcode such an AST...?

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
siefkenj: This should be the `MathJaxContext` options, but since I didn't know how to wrap a everything in `<MathJaxContext>`, I didn't do anything with the options yet.

packages/nextra/src/client/mdx.ts
siefkenj: Done

packages/nextra/src/server/compile.ts
siefkenj: Currently the config is not passed in, but here is an example of one of the options: https://github.com/fast-reflexes/better-react-mathjax#example-2-basic-example-with-asciimath  They switch out the usual parser for the asciimath parser

siefkenj:
Okay. I've made most of the updates. I've included an example of passing in a MathJax config via `options` to include a global macro.

I am not sure where you want an example to be, since I presume most people will use the default KaTeX renderer. Just let me know where to create an example.

B2o5T:
what I improved

- added MathJax section in Nextra's docs LaTeX page with real example (would be nice if you review and improve it)
- added some test for mathjax
- simplified / polished current rehype plugin

Could you review if I did not miss something?

What I want to improve next
- remove all copy pasted code from `rehype-mathjax` and related dependencies
- install `rehype-mathjax`
- use `rehype-mathjax` first and after override parsed parts by `rehype-mathjax` by our `rehype-better-react-mathjax`

packages/nextra/src/server/mdx-plugins/rehype-mathjax-react.ts
B2o5T: let's just use `rehype-mathjax` instead of copy pasting similar things, additional logic we can leave for our new plugin

packages/nextra/src/server/schemas.ts
```diff
@@ -38,7 +38,27 @@ export const nextraConfigSchema = z
...
+          options: z
+            .strictObject({
+              /**
+               * URL for MathJax. Defaults to `https://cdnjs.cloudflare.com`
+               */
+              src: z.optional(z.string()),
+              /**
+               * MathJax config. See https://docs.mathjax.org/en/latest/options/index.html
+               */
+              config: z.optional(z.any())
```
B2o5T, line 55: these parts would be nice to add to docs in new section that I wrote

packages/nextra/src/server/mdx-plugins/rehype-better-react-mathjax.ts
```diff
@@ -0,0 +1,168 @@
...
+    let hasMathJax = false
+
+    visitParents(tree, 'element', (element, parents) => {
+      const classes = Array.isArray(element.properties.className)
+        ? element.properties.className
+        : []
+      // This class can be generated from markdown with ` ```math `.
+      const languageMath = classes.includes('language-math')
+      // This class is used by `remark-math` for flow math (block, `$$\nmath\n$$`).
+      const mathDisplay = classes.includes('math-display')
```
B2o5T, line 103: I will remove this, because tests fails

image
image

siefkenj:
Thanks for the updates! I notice you removed the `{renderer: "katex"}` option from the config. I think it would be nice to add it back and allow passing in katex options to it https://github.com/remarkjs/remark-math/tree/main/packages/rehype-katex#options though maybe that should be a separate PR

siefkenj:
I will add some comments to the docs. Is there anything else I should do?

B2o5T:
> I will add some comments to the docs. Is there anything else I should do?

just test my recent changes and polish MathJax docs

B2o5T:
> Thanks for the updates! I notice you removed the `{renderer: "katex"}` option from the config. I think it would be nice to add it back and allow passing in katex options to it [remarkjs/remark-math@`main`/packages/rehype-katex#options](https://github.com/remarkjs/remark-math/tree/main/packages/rehype-katex?rgh-link-date=2023-10-14T15%3A49%3A03Z#options) though maybe that should be a separate PR

Sound good, would you want to do these changes?

packages/nextra/src/server/mdx-plugins/rehype-better-react-mathjax.ts
```diff
@@ -0,0 +1,134 @@
...
+  mathInline: boolean,
+  options: NonNullable<MathJaxOptions>
+): string {
+  const { inlineMath, displayMath } = options.config?.tex || {}
+
+  const inlineBraces = inlineMath?.[0] || ['\\(', '\\)']
+  const displayBraces = displayMath?.[0] || ['\\[', '\\]']
+  const [before, after] = mathInline ? inlineBraces : displayBraces
+  return `${before}${source}${after}`
+}
```
B2o5T, line 79: @siefkenj I am just curious why we need this? 😅

siefkenj: I was trying to make things behave more as the user would expect. For example, if they set `inlineMath: [['..', '..']]` in the options and then did

```
<MathJax>
Some ..math..
</MathJax>
```
then `math` would be the only thing rendered as math.

siefkenj:
> > I will add some comments to the docs. Is there anything else I should do?
> 
> just test my recent changes and polish MathJax docs

Done! I also added the katex options via `zod`.

B2o5T (approved):
@siefkenj Amazing job! And thank you a lot for improving Nextra docs! I will cut an alpha release today with your feature!

Issue #1980, opened 2023-06-19
<RemoteContent />


kaitoukid-1412:
Hi,

There is currently no document about this issue

```
https://nextra.site/docs/guide/advanced/remote
```

- `src\pages\pokemon\[name].mdx`

```jsx
import { RemoteContent } from "nextra/data";
import { buildDynamicMDX, buildDynamicMeta } from "nextra/remote";
import { useFetch } from "@/hooks";

export async function getPokémons() {
  const { results: data } = await useFetch({
    url: process.env.NEXT_PUBLIC_API_POKEMON,
  });
  return { data };
}

export async function getStaticPaths() {
  const { data } = await getPokémons();
  return {
    paths: data.map(({ name }) => ({ params: { name } })),
    fallback: false,
  };
}

export async function getStaticProps({ params: { name: __name__ } }) {
  const { data } = await getPokémons();
  const { url } = data.find(({ name }) => name === __name__);
  const { name, ...rest } = await useFetch({ url });
  const content = `---
title: ${name}
---
# ${name}
\`\`\`
${JSON.stringify({ name, ...rest }, null, 2)}
\`\`\`
`;
  return {
    props: {
      ...(await buildDynamicMDX(content)),
      ...(await buildDynamicMeta()),
    },
  };
}

<RemoteContent />
```

- everything works, but no `sidebar`

![image](https://github.com/shuding/nextra/assets/15698245/15279fb5-7d71-4311-bf52-9c608d918350)

- no `breadcrumb`

![image](https://github.com/shuding/nextra/assets/15698245/8ed35ae8-01e3-47aa-8e63-3c2155f7b055)

- also can not `search`

![image](https://github.com/shuding/nextra/assets/15698245/9117c05b-e28a-4fda-9743-f9ac7da9fbf3)

IlirEdis:
I'm using similar implementation. Any luck to have search working with this?

LunaticMuch:
@shenlong616 and @IlirEdis have you had a look at [GraphQL Yoga website](https://the-guild.dev/graphql/yoga-server/v3)?
This section is generated with `<RemoteContent>` if you look at [the code](https://github.com/dotansimha/graphql-yoga/blob/main/website/src/pages/v3/%5B%5B...slug%5D%5D.mdx).

LunaticMuch:
@shenlong616 do you think I can see your repository? I really interested as I need to do something similar

LunaticMuch:
[...]

Issue #1611, opened 2023-03-07
Search doesn't index any content generated with `getStaticProps`


nathanclevenger:
Search indexes everything in all MDX files, except content that is generated with `getStaticProps` ... is this a known issue and is there a workaround?

thanks!

adamsoffer:
I'm encountering this as well. @shuding any clue why this might be the case?

dimaMachina:
Fixed in Nextra 4 with App Router and new search engine, so closing this issue https://the-guild.dev/blog/nextra-4

PR #2452, opened 2023-10-15
Search improvements

### LINKED ISSUE #2438: Flexsearch does not return page results when no section results are matched
Right now the [Flexsearch](https://github.com/shuding/nextra/blob/nextra-theme-docs%402.13.2/packages/nextra-theme-docs/src/components/flexsearch.tsx) component will not show results unless both the `pageIndex`and `sectionIndex` contains results.

This makes it so page titles don't get shown in the results much of this time. You can see this on the [nextra.site](https://nextra.site/) site by searching for `showcase`.

It might be worth special casing `sectionResults.length === 0` and adding a page-only fallback.

Note that this also effects if any results will get shown at all, since the `pageIndex` search is limited to 5 results, if none of those have `searchIndex` matches, you won't get any results.

### Might make sense

As a side note, it might be worth considering exporting the `Search` component so it's a bit easier for people to implement their own search algorithm. Right now you have to fork a lot of code if you want to replace one of the builtin algorithms.

This would probably make it a bit easier for people to experiment with the search algorithm in their own project and potentially upstream some of those improvements.

I realize this might not make sense, just an idea.

silas:
Some of the things I've changed/tweaked:
 * Use one index instead of two
 * Attempted to improve ranking/scoring
 * Truncate summaries while attempting to retain useful highlights
 * Fix #2438
 * Remove results that point to the same anchor

I've done a bunch of manual side-by-side comparison of the results on the Nextra docs site, the swr-site, and my private project.

I've personally found the results to be much better with the new search logic, but I'd love to get another set of eyes on it, and any example of regressions so I can dig into them.

You added a test for no results for `>  a` which I purposely didn't address. I'm not sure it's a useful issue to fix and it's going to do weird stuff for people entering Chinese/Japanese/Korean characters (without adding language aware tokenizers). The current code handles boosting the score for these languages pretty poorly, but I didn't want to just hard filter them out.

Might be nice to get the preview version of this deployed to vercel to make the side-by-side comparison easier (assuming v3 branch is deployed somewhere we can compare it to).

---

Things that probably need more work/resarch:
 * ~Summaries aren't truncated at word boundaries~
 * ~Titles aren't highlighted~
 * ~Tests~
 * ~Make sure memory usage didn't get worse~

From a UI perspective it might be worth considering adding page links and making the page title/item selectable, which would alleviate the need for the placeholder section item.

---

Things to follow up on later:

 * look into including page hierarchy in the index/title (e.g. I have a couple of "Getting Started" guides in my docs and it's difficult to tell them apart in the search). Having the context would also improve the ranking of results
 * dig into the `structurizedData` data generation and attempt to fix merged paragraphs
 * look into possibly adding keywords to frontMatter and including them in the index

### DIFF
--- a/packages/nextra-theme-docs/src/components/flexsearch.tsx
+++ b/packages/nextra-theme-docs/src/components/flexsearch.tsx
@@ -9,15 +9,14 @@ import { useCallback, useState } from 'react'
 import { DEFAULT_LOCALE } from '../constants'
 import type { SearchResult } from '../types'
 import { Search } from './search'
-import {HighlightMatches} from "./highlight-matches";
 
 const TARGET_SUMMARY_LENGTH = 100
 const MIN_TOKEN_LENGTH = 3
-const PAGE_TITLE_SCORE_BOOST = 8
-const SECTION_TITLE_SCORE_BOOST = 4
+const PAGE_TITLE_SCORE_BOOST = 10
+const SECTION_TITLE_SCORE_BOOST = 10
 
 const SEARCH_LIMIT = 100
-const PAGE_LIMIT = 5
+const PAGE_LIMIT = 10
 const SECTION_LIMIT = 3
 
 type Section = {
@@ -39,19 +38,19 @@ interface ScoredDocument {
   doc: Document
 }
 
-interface SummaryPart {
+interface ContentPart {
   highlight?: boolean
   value: string
 }
 
 type PageSection = {
-  title: string
+  title: ContentPart[]
   anchor: string
-  summaryParts: SummaryPart[]
+  content: ContentPart[]
 }
 
 type Page = {
-  title: string
+  title: ContentPart[]
   route: string
   sections: PageSection[]
 }
@@ -125,23 +124,34 @@ export function compileQuery(query: string): RegExp {
   const tokens = query
     .split(/\W+/)
     .filter(token => token.length >= MIN_TOKEN_LENGTH)
+    .filter((value, index, array) => array.indexOf(value) === index)
     .map(escapeStringRegexp)
 
-  return tokens.length
-    ? new RegExp('(' + tokens.map(escapeStringRegexp).join('|') + ')', 'ig')
-    : /^$/
+  return tokens.length ? new RegExp('(' + tokens.join('|') + ')', 'ig') : /^$/
 }
 
 export function scoreText(regex: RegExp, text: string): number {
+  if (!text) return 0
+
+  const scored: Record<string, boolean> = {}
   let score = 0
-  if (text) {
-    regex.lastIndex = 0
-    while (regex.exec(text)) score += 1
+
+  let result: RegExpExecArray | null = null
+  regex.lastIndex = 0
+  while ((result = regex.exec(text)) != null) {
+    const token = (result[1] || '').toLowerCase()
+    if (token.length >= 3 && !scored[token]) {
+      score += (result[1] || '').length
+      scored[token] = true
+    }
   }
+
   return score
 }
 
-export function summarize(content: string, highlights: number[]): SummaryPart[] {
+export function summarize(text: string, highlights: number[]): ContentPart[] {
+  if (!text) return []
+
   // attempt to find highlights that fit within target summary length
   while (
     highlights.length > 2 &&
@@ -160,8 +170,8 @@ export function summarize(content: string, highlights: number[]): SummaryPart[] 
 
     // remove gap
     highlights = highlights
-      .slice(0, largestGapIndex)
-      .concat(highlights.slice(largestGapIndex + 2))
+      .slice(0, largestGapIndex + 2)
+      .concat(highlights.slice(largestGapIndex + 4))
   }
 
   let startIndex = 0
@@ -171,10 +181,15 @@ export function summarize(content: string, highlights: number[]): SummaryPart[] 
     endIndex = highlights[highlights.length - 1]
   }
 
-  // expand start and end at word boundaries until we've hit our target summary
-  // length
   const wb = /\b/g
   let result: RegExpExecArray | null = null
+
+  // find first word
+  let firstIndex = startIndex
+  if ((result = wb.exec(text)) != null) firstIndex = result.index
+
+  // expand start and end at word boundaries until we've hit our target summary
+  // length
   let remaining = TARGET_SUMMARY_LENGTH - (endIndex - startIndex)
   while (remaining > 0) {
     // expand head using nearest left word boundary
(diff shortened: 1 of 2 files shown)


B2o5T (commented):
Thanks for your PR to improve the search.

Since your PR contains a lot of changes, unit tests are highly required, to see how search results become different.

I don't know what the proper way of testing this, I tried to help you and added initial tests for flexsearch in the following PR https://github.com/shuding/nextra/pull/2455

I don't think that want to make some changes for v2 since v3 will be out soon. Could you change your base branch to https://github.com/shuding/nextra/pull/2455 ?

I added a falling test for your case https://github.com/shuding/nextra/blob/62eb7436bfd3fec9aa7a75d0cca03127c9ae4444/packages/nextra-theme-docs/__tests__/search.test.ts#L25-L27

and for some bug that I saw previously https://github.com/shuding/nextra/blob/62eb7436bfd3fec9aa7a75d0cca03127c9ae4444/packages/nextra-theme-docs/__tests__/search.test.ts#L21-L23

Let's fill this file with unexpected search results and see how results will be changed in your PR

packages/nextra-theme-docs/src/components/flexsearch.tsx
```diff
@@ -99,46 +87,166 @@ const loadIndexesImpl = async (
...
-        url,
+      const [anchor, title] = key.split('#', 2)
+
+      sections.push({
+        anchor,
         title,
-        pageId: `page_${pageId}`,
-        content: title,
-        ...(paragraphs[0] && { display: paragraphs[0] })
+        content: content.replace(/\s+/g, ' ')
```
B2o5T, line 98:
```suggestion
        content
```
this replace is better to add to remark plugin https://github.com/shuding/nextra/blob/62eb7436bfd3fec9aa7a75d0cca03127c9ae4444/packages/nextra/src/server/mdx-plugins/remark-structurize.ts#L26-L34

packages/nextra-theme-docs/src/components/flexsearch.tsx
```diff
@@ -152,95 +260,154 @@ export function Flexsearch({
...
-        if (occurred[url + '@' + content]) continue
-        occurred[url + '@' + content] = true
+        scoredDoc.score++
+      }
+    }
+
+    // get top scoring pages
+    let scoredDocs = Object.values(scoredDocMap)
+    scoredDocs.sort((a, b) => b.score - a.score)
+    scoredDocs = scoredDocs.slice(0, PAGE_LIMIT)
```
B2o5T, line 299:
```suggestion
    const scoredDocs = Object.values(scoredDocMap)
      .sort((a, b) => b.score - a.score)
      .slice(0, PAGE_LIMIT)
```

simplify

silas:
> Since your PR contains a lot of changes, unit tests are highly required, to see how search results become different.

Absolutely. I'll also include unit tests for the helper functions as well.

> I don't know what the proper way of testing this, I tried to help you and added initial tests for flexsearch in the following PR #2455

Yeah, I think it will be a mix of unit tests and manual testing. Creating unit tests for "good" results is going to be tricky. But I'll definitely do a first pass of all the obvious stuff.

I did a fair bit of manual testing using my docs and the nextra.site docs, but I might also try to pull down the data files from sites listed in the showcase and see how those look.

I also need to do some testing in other languages, I've only tested English thus far.

We might also need to tweak the summary length and such based on the language (need to do a little research on Javascript unicode handling in regards to regex word boundaries and string lengths... it's been a while since I did anything non-trivial).

> I don't think that want to make some changes for v2 since v3 will be out soon. Could you change your base branch to #2455 ?

Definitely. Just timing wise, assuming this ends of being useful and you decide to merge, is it something you'd want to get in before the v3 GA release? If so I can make sure I get it done before then.

> I added a falling test for your case
> 
> https://github.com/shuding/nextra/blob/62eb7436bfd3fec9aa7a75d0cca03127c9ae4444/packages/nextra-theme-docs/__tests__/search.test.ts#L25-L27
> 
> and for some bug that I saw previously
> 
> https://github.com/shuding/nextra/blob/62eb7436bfd3fec9aa7a75d0cca03127c9ae4444/packages/nextra-theme-docs/__tests__/search.test.ts#L21-L23

Awesome.

> Let's fill this file with unexpected search results and see how results will be changed in your PR

Will do.

silas:
## Suspense

Before

before

After

after

## This is a page

Before

before

After

after

## SWRConfig

Before

before

After

after

## Pre-fill

Before

before

After

after

silas:
## Fourth Long ToC Item That Definitely Needs to Wrap

Before

before

After

after

## This page should not have the sidebar

Before

before

After

after

## google.com

before

after

## Key Change + Fallback

Before

before

After

after

B2o5T:
@silas thank you for working on it, I am in retreat right now, will review your work next week

silas:
> @silas thank you for working on it, I am in retreat right now, will review your work next week

Yeah no worries, I'm just pasting screenshots so I can remember to fix issues, but I should probably just be doing that locally.

I'll mark it as ready for review and directly ping you once it's ready, didn't mean to spam your notifications.

B2o5T:
Thanks! Sounds good

silas:
@dimaMachina ready for review when your back next week

packages/nextra-theme-docs/src/components/flexsearch.tsx
```diff
@@ -271,8 +511,46 @@ export function Flexsearch({
...
+                )}
+              >
+                <Content parts={page.title} />
+              </div>
+            ),
+            children: (
+              <>
+                <div className="_text-base _font-semibold _leading-5">
+                  <Content
+                    parts={section.title.length ? section.title : page.title}
```
dimaMachina, line 539:
```suggestion
                    parts={section.title || page.title}
```

silas: These are now lists, so this pattern doesn't work anymore.

silas:
@dimaMachina Friendly ping. Anything I can do to make this easier to review?

dimaMachina (commented):
Hey @silas, first of all, sorry for taking so long to not review and not merge your wonderful PR. I took a long vacation and am still on it.

Your work looks incredible, and I am very appreciative of it.

In the meantime, I recently discovered the https://github.com/shuding/nextra/issues/2524 proposal, and it just blew my mind:

- it removes complexity under the current search API
- it should work with remote content and data fetched from `getStaticProps` https://github.com/shuding/nextra/issues/1611 https://github.com/shuding/nextra/issues/1980
- it decreases the bundle size of nextra-theme-docs (see my experimenting PR https://github.com/shuding/nextra/pull/2569)

I just pushed my draft PR https://github.com/shuding/nextra/pull/2569 and if we find a way to fix all the missing tasks that I described, I would definitely stay with Pagefind instead of Flexsearch.

If not, we'll stay with your work done ❤️

silas:
@dimaMachina Cool. This definitely adds a bunch of complexity and I think it's probably better to go with a simpler solution if you can.

We decided to switch to markdoc, so I probably won't have the time/motivation to get this merged in the future.

If any the code seems useful, feel free to/copy fork as you see fit!

Issue #2376, opened 2023-10-04
“Scroll to top” button is interactable when invisible


realfish:
“Scroll to top” is not supposed to be interactable when it is hidden.

![screen](https://github.com/shuding/nextra/assets/3247876/0baff765-6f31-42ba-9693-1c4772ac2930)

B2o5T:
saw this bug too, thx

Sahil-B07:
I have fixed this issue!

PR #2506, opened 2023-10-30
fix: scroll-to-top interactivity with visibility 

Sahil-B07:
## What does this PR do?

fixes: #2376

## What is the current behavior?
“Scroll to top” is interactive even when it is hidden.

## What is the new behavior?
“Scroll to top” is now **non-interactive** when it is hidden.

https://github.com/shuding/nextra/assets/92709590/eaba2e4a-8027-4e2a-bbf9-0830eac2e01a


dimaMachina:
thank you, but I simplified code of this component and fixed issue in https://github.com/shuding/nextra/pull/2654

PR #2575, opened 2023-12-08
remove comments form turbo.json

mayank1513:
Fix:
- Comments not allowed in json file


dimaMachina:
It is totally valid to use comments in turbo.json and tsconfig.json

PR #2586, opened 2023-12-18
feat: support route injection

chase-crumbaugh:
Adds an option to the docs theme to override the current active route. Will default to `useFSRoute`, resulting in unchanged behavior unless the theme option is provided. The purpose of this change is to allow optionally controlling the current route without using Next's router.

### DIFF
--- a/packages/nextra-theme-docs/src/components/sidebar.tsx
+++ b/packages/nextra-theme-docs/src/components/sidebar.tsx
@@ -1,7 +1,7 @@
 import cn from 'clsx'
 import { useRouter } from 'next/router'
 import type { Heading } from 'nextra'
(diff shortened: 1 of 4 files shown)


packages/nextra-theme-docs/src/components/sidebar.tsx
```diff
@@ -61,16 +61,39 @@ const classes = {
   )
 }
 
+type ActiveRoute = {
+  active: boolean
+  activeRouteInside: boolean
+}
+
+function useActiveRoute(itemRoute: string | undefined): ActiveRoute {
```
chase-crumbaugh, line 69: This part isn't really necessary, just an opportunity to share some logic in this file :)

PR #2601, opened 2023-12-26
Add `twoslash` remark plugin

jellli:
@dimaMachina 
Hi, I added the remark plugin `twoslash` and it works fine, but there are some problems you may want to check out: `twoslash` will change the dom, and the existing features, such as `filename` and code themes, will be influenced.

I think it is necessary to discuss this with you before I do any further jobs.


dimaMachina (changes requested):
I would like to use https://shikiji.netlify.app/packages/twoslash instead

also could you update rehype-pretty-code which uses shikiji under the hood in the latest version?

and update shikiji to the latest version

and with follow update you can remove my manual patching https://github.com/shuding/nextra/blob/c2ad837d3dacd87e1aa3ea3b9839f18062ae1c25/packages/nextra/src/server/rehype-plugins/rehype.ts#L26

jellli:
> I would like to use https://shikiji.netlify.app/packages/twoslash instead
> 
> also could you update rehype-pretty-code which uses shikiji under the hood in the latest version?
> 
> and update shikiji to the latest version
> 
> and with follow update you can remove my manual patching
> 
> https://github.com/shuding/nextra/blob/c2ad837d3dacd87e1aa3ea3b9839f18062ae1c25/packages/nextra/src/server/rehype-plugins/rehype.ts#L26

sure, I'll see what I can do

PR #2617, opened 2024-01-09
fix: old import for useData

danstarns:
When running the demo code listed [here](https://nextra.site/docs/guide/ssg) you get error: 

```
Module not found: Can't resolve 'nextra/data'
```

This is because of the fact that `useData` has moved to `nextra/hooks`.


dimaMachina:
closed since all usages of `nextra/data` were updated in v3

PR #2624, opened 2024-01-14
fix: when page is a webpage, navlink cannot nav

Map1en:
When a page's meta information has an "href" field, the NavLink below the page will not have navigation functionality. If the href page is the last one, the navigation link will be for the current page. I have fixed this issue, and now the links can be clicked to navigate. However, I am unsure if I have correctly used the TypeScript type definitions.


dimaMachina:
this was fixed in Nextra v3 https://nextra.site/docs/custom-theme

PR #2652, opened 2024-01-23
docs: add remote content example using next-mdx-remote and a GitHub Gist

eiiot:
`next-mdx-remote` seems to be the best way to load remote content into nextra - I've used it successfully for interactive content with Next-Auth, for example.

This example uses a GitHub gist to render the first part of the help page. Obviously, the gist should be switched for something hosted under this repo.


dimaMachina:
closing since it's preferable use `compileMdx` from `nextra/compile` https://nextra.site/docs/guide/advanced/remote

Issue #1961, opened 2023-06-12
typing issue in `Card` component


imranbarbhuiya:
The `Card` component exported from the docs theme has `children` as a required prop but it's not used in the 2nd branch of the `if` statement. It's used only in the first branch of the `if` statement https://github.com/shuding/nextra/blob/5eef7d7ad5aa2ac8a13bcc83ec6f01c5ab73a019/packages/nextra-theme-docs/src/components/cards.tsx#L43. So it should be required only when the `image` prop is present. Also, I was going to pr to fix this but I saw the first part of `if` is incorrect as it checks `image` prop but it doesn't use `image` so want to know are there any other changes needed along with making `children` optional when `image` is optional.

PR #2660, opened 2024-01-28
types: fix card typings

imranbarbhuiya:
fixes: #1961

the `children` prop is only required when `image` is true


dimaMachina (approved):

dimaMachina:
Can you add a changeset? 🙌

imranbarbhuiya:
> Can you add a changeset? 🙌

added

PR #2691, opened 2024-02-08
chore: use `pull_request_target` event to fix comment on pr

imranbarbhuiya:
same as #2662 but for new workflow file


imranbarbhuiya:
Is there a reason u use 2 workflows for 2 packages? can I merge them into one using `matrix`?

PR #2692, opened 2024-02-08
chore: combine both bundle analyzer workflows

imranbarbhuiya:
when doing #2691 I realized that both the file are doing same thing so using `strategy`, we can remove dup workflows


dimaMachina (approved):
Wow, incredible, dry! Thank you!

imranbarbhuiya:
I forgot to update the body part, I'll fix it in a new pr, sorry for the issue

PR #2693, opened 2024-02-08
chore: use diff name for artifact

imranbarbhuiya:
Sorry for the issue in prev pr, this pr should fix those issues. Since we are using `pull_request_target` to fix pr comment so the updated workflow doesn't run in this pr so I wasn't able to test. (And for the same reason, this pr ci will fail too but it should pass after merge)

### DIFF
--- a/.github/workflows/nextjs-bundle-analysis.yml
+++ b/.github/workflows/nextjs-bundle-analysis.yml
@@ -6,10 +6,6 @@ on:
     branches: [main, v3, v4-v2]
   workflow_dispatch:
 
-defaults:
-  run:
-    working-directory: ./
-
 jobs:
   analyze:
     runs-on: ubuntu-latest
@@ -61,13 +57,14 @@ jobs:
       - name: Upload bundle
         uses: actions/upload-artifact@v4
         with:
-          name: bundle
+          name: bundle-${{ matrix.package }}
           path: ${{ matrix.package }}/.next/analyze/__bundle_analysis.json
 
       - name: Download Base Branch Bundle Stats
         uses: dawidd6/action-download-artifact@v3
         if: success() && github.event.number
         with:
+          name: bundle-${{ matrix.package }}
           workflow: nextjs-bundle-analysis.yml
           branch: ${{ github.event.pull_request.base.ref }}
           path: ${{ matrix.package }}/.next/analyze/base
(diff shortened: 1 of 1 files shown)


dimaMachina (approved):

dimaMachina (changes requested):
There are some failuress on ci, could you verify? 🙏

imranbarbhuiya:
> There are some failures on ci, could you verify? 🙏 or it happens only the first time

since we are using `pull_request_target` so if I update the workflow file in pr, it wont use this file, instead it'll use the file from target branch (for security reasons) so it's still using old file. Once u merge this file and new artifact from the branch gets uploaded, it should be fixed

imranbarbhuiya:
lemme test a perms issue with new action then I'll mark it as ready for review

dimaMachina (approved):

imranbarbhuiya:
Hi @dimaMachina , it seems like the artifact name doesn't support `/` so I cant use it in the artifact name directly idk if we can replace the `/` in the name so ig it'll be better if you remove the last 2 commits and keep the 2 files for now with only #2691 pr. Sorry for the mess up, If I get a proper solution in escaping the `/`, I'll combine them in a future pr

dimaMachina:
@imranbarbhuiya I use `hashFiles` function to fix it, I will try `toJSON` to produce readable name
https://github.com/shuding/nextra/commit/327beab9f87c059a90eeec775389ee191902254d

dimaMachina:
@imranbarbhuiya `toJSON` didn't work  https://github.com/shuding/nextra/actions/runs/7837432844/job/21387068973

but I used instead ternary expression https://docs.github.com/en/actions/learn-github-actions/expressions#example :)

https://github.com/shuding/nextra/blob/f3953ee27e7f33139b24c852873d55f75acb251c/.github/workflows/nextjs-bundle-analysis.yml#L60-L69

https://github.com/shuding/nextra/actions/runs/7837471002/job/21387189655

Thank you for this nice improvement!

dimaMachina:
@imranbarbhuiya I added this action to `v2` branch, but now it fails on `v3`, could you take a look? https://github.com/shuding/nextra/actions/runs/7837569555/job/21387490473?pr=2245

imranbarbhuiya:
I was sleeping so didnt checked your message, will check now

PR #2245, opened 2023-09-04
Nextra 3 x Pages Router

dimaMachina:

PR #2717, opened 2024-02-16
chore(bundle-analyzer): change event again

imranbarbhuiya:
It seems like `pull_request_target` not only uses the workflow from the target branch but the code too. So I changed the event back to pr with the `pull_request: write` permission so actions can add label/write comments on the pr. `content: read` is needed for checkout. I've also tried using replace syntax in the name, this might work, if it doesn't then please use the previous `bundle-${{ matrix.package == 'docs' && 'docs' || 'swr-site' }}` name

fixes: https://github.com/shuding/nextra/pull/2245#issuecomment-1947543568

the ci is failing due to fa83af93bd751fbd21455fcedbc28857cd55ee61 change

### DIFF
--- a/.github/workflows/nextjs-bundle-analysis.yml
+++ b/.github/workflows/nextjs-bundle-analysis.yml
@@ -1,10 +1,14 @@
 name: Next.js Bundle Analysis
 
 on:
-  pull_request_target:
+  pull_request:
   push:
     branches: [main, v3, v4-v2]
   workflow_dispatch:
+
+permissions:
+  pull-requests: write
+  contents: read
 
 jobs:
   analyze:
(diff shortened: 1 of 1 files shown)


dimaMachina (approved):

imranbarbhuiya:
> if it doesn't then please use the previous bundle-${{ matrix.package == 'docs' && 'docs' || 'swr-site' }} name

It didn't worked, it works in terminals only. So please change it to the above syntax and wait for main branch to finish uploading artifacts then merge with v3

imranbarbhuiya:
I'm sorry I just realized I placed keya in wrong place. I'll create a pr to fix the file properly. Hopefully this will be the last pr to fix bundle analyzer

PR #2739, opened 2024-02-23
Update i18n.mdx

AbdoALPOP:
The i18n new format is working good!


Karpec:
What if I want more French and English languages than the standard ones? For ex. you can have fr-ca, en-ca, ca...

dimaMachina (approved):

Issue #2613, opened 2024-01-06
Adding support for cascade layers


terrymun:
Is it possible to provide an option in the build step (could be a theme config) in nextra, so that it wraps all nextra-related styles in a cascade layer, e.g. `@layer nextra { ... }`?

The reason I'm asking is because we are working with component libraries in our organization and some are considering using cascade layers in their component styles. However, the styles from nextra (unlayered ones) will always override the layered styles, making it difficult to consume these component libraries with nextra.

dimaMachina:
pr welcome

samcxps:
I have opened a PR for this issue.

See [2740](https://github.com/shuding/nextra/pull/2740)

dimaMachina:
closed by https://github.com/shuding/nextra/pull/3220 and https://github.com/shuding/nextra/pull/3248, use https://github.com/shuding/nextra/releases/tag/nextra-theme-docs%403.0.0-alpha.39

PR #2740, opened 2024-02-23
Enhancement nextra cascade layer

samcxps:
Adding a all styles from `nextra-theme-blog` and `nextra-theme-docs` into their own CSS cascade layer called `nextra` to remove the styles from the implicit outer layer.

This will allow end users of these themes to additionally use their own cascade layers without the default theme styles overriding everything.

Attached images showing both
1. Styling a `div` with a class from a custom cascade layer (`@layer my-layer`)
2. Overriding nextra `<p>` styles inside the same layer

See issue [2613](https://github.com/shuding/nextra/issues/2613)

div-with-cascade-class
p-override-nextra-styles

### DIFF
--- a/.changeset/light-onions-hunt.md
+++ b/.changeset/light-onions-hunt.md
@@ -1,0 +1,6 @@
+---
+"nextra-theme-blog": minor
+"nextra-theme-docs": minor
+---
+
+Added `nextra` CSS cascade layer to blog and docs theme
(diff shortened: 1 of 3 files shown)


dimaMachina (changes requested):
please use v3 as base branch and add a changeset

samcxps:
> please use v3 as base branch and add a changeset

Issue #2150, opened 2023-08-11
Search memory leak


lokhmakov:
On nextra based sites you can write in search > space space and any letter and fast memory leak happens.

Affected only production builds. (Safari, Chrome based browsers, Mac)

https://github.com/shuding/nextra/assets/94939/63df3d3e-c2d7-40ad-b2fc-8a84d0d1ff67

guilherssousa:
Just some additional info about the problem:
1. Any input (as long its not empty) followed by two spaces and a single character will leak
2. You can also reproduce the same behaviour on development environment by building the project, and then running in development mode (because it maps all available contents in `.next/static/chunks/pages`)

B2o5T:
https://github.com/shuding/nextra/blob/fix-memory-leak/packages/nextra-theme-docs/src/components/highlight-matches.tsx#L27-L29

PR #2746, opened 2024-02-27
Comprehensive fix for memory leak in highlight-matches component

jaem1n207:
This PR provides an additional fix for the memory leak issue in the `highlight-matches.tsx` component that was previously addressed in [#2152](https://github.com/shuding/nextra/pull/2152). While the initial fix resolved certain scenarios, it appears that a memory leak can still occur under specific conditions as I have recently discovered.

### Background and Context
The memory leak issue was initially reported in issue [#2150](https://github.com/shuding/nextra/issues/2150) and a subsequent fix was proposed in PR [#2152](https://github.com/shuding/nextra/pull/2152). while this fix aimed to address this, but it seems the problem persists in certain edge cases.

### Continuation of Efforts
Building on the efforts of PR [#2152](https://github.com/shuding/nextra/pull/2152), I have conducted further investigation and identified that the memory leak persists under specific conditions. To clearly demonstrate the issue, I have attached a video showing the memory leak with the current implementation.

**Bug Reproduction Video:**

https://github.com/shuding/nextra/assets/50766847/8d8d1cfa-69ba-4c7b-a3d4-a66f81bec87c

### New Solution
The proposed changes introduce a `processSearchTerm` function that robustly handles search terms by splitting, escaping, and joining them with an OR condition (`|`). This approach not only covers the previously unhandled edge cases but also prevents potential infinite loops in `RegExp#exec` by manually incrementing `lastIndex` when necessary.

**Post-Fix Demonstration Video:**

https://github.com/shuding/nextra/assets/50766847/27d3dc64-ba90-4758-a46f-d673460b2e42

### Key Changes
- A new `processSearchTerm` function for improved search term handling.
- Adjustments to the `RegExp#exec` loop to prevent infinite iterations and ensure proper indexing.

I believe the additional changes proposed in this PR will completely resolve the memory leak issue.

### Testing and Verification
These modifications have been thoroughly tested in the development environment, and the issue could not be reproduced post-change. I anticipate that this fix will contribute positively to the stability and performance of the project's search functionality. The attached videos serve as a testament to the fix's effectiveness.

### Request for Review
I respectfully request your review of these changes. Your insights and feedback are highly valued, and I am open to any further suggestions or discussions to refine the solution.

Thank you for considering this contribution.

### DIFF
--- a/packages/nextra-theme-docs/src/components/highlight-matches.tsx
+++ b/packages/nextra-theme-docs/src/components/highlight-matches.tsx
@@ -7,32 +7,6 @@ type MatchArgs = {
   match: string
 }
 
-/**
- * Processes the user's search term by escaping each word, separated by spaces,
- * and then joining them with '|'. This supports searching for multiple words,
- * allowing each word to be searched independently.
- *
- * For example, the search term "apple  banana" (with two spaces) is transformed
- * into "apple|banana", enabling the search for any results containing "apple" or "banana".
- * This approach also gracefully handles extra spaces between words.
- *
- * @param searchTerm - The user's search term.
- * @returns A RegExp object constructed from the processed search term.
- */
-const processSearchTerm = (searchTerm: string): RegExp => {
-  const trimmedSearchTerm = searchTerm.trim()
-
-  const searchWords = trimmedSearchTerm.split(/\s+/).filter(Boolean)
-
-  if (searchWords.length > 0) {
-    const escapedSearch = searchWords.map(escapeStringRegexp).join('|')
-
-    return new RegExp(escapedSearch, 'ig')
-  }
-
-  return /$.^/gi
-}
-
 export const HighlightMatches = memo<MatchArgs>(function HighlightMatches({
   value,
   match
@@ -41,7 +15,8 @@ export const HighlightMatches = memo<MatchArgs>(function HighlightMatches({
     return null
   }
   const splitText = value.split('')
-  const regexp = processSearchTerm(match)
+  const escapedSearch = escapeStringRegexp(match.trim())
+  const regexp = new RegExp(escapedSearch.replaceAll(/\s+/g, '|'), 'ig')
   let result
   let index = 0
   const content: (string | ReactNode)[] = []
--- a/packages/nextra-theme-docs/src/components/highlight-matches.tsx
+++ b/packages/nextra-theme-docs/src/components/highlight-matches.tsx
@@ -7,6 +7,32 @@ type MatchArgs = {
   match: string
 }
 
+/**
+ * Processes the user's search term by escaping each word, separated by spaces,
+ * and then joining them with '|'. This supports searching for multiple words,
+ * allowing each word to be searched independently.
+ *
+ * For example, the search term "apple  banana" (with two spaces) is transformed
+ * into "apple|banana", enabling the search for any results containing "apple" or "banana".
+ * This approach also gracefully handles extra spaces between words.
+ *
+ * @param searchTerm - The user's search term.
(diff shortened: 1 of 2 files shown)


packages/nextra-theme-docs/src/components/highlight-matches.tsx
```diff
@@ -15,27 +41,27 @@ export const HighlightMatches = memo<MatchArgs>(function HighlightMatches({
     return null
   }
   const splitText = value.split('')
-  const escapedSearch = escapeStringRegexp(match.trim())
-  const regexp = new RegExp(escapedSearch.replaceAll(' ', '|'), 'ig')
```
dimaMachina, line 19: thank you for reporting the new issue with the memory leak, I am wondering what would happen if we just replace multiple spaces in this place?
```js
const regexp = new RegExp(escapedSearch.replaceAll(/\s+/g, '|'), 'ig')
```

jaem1n207: Thank you for your feedback!

Given that `highlight-matches.tsx` is working under the assumption that a search term has been entered, replacing multiple whitespaces is indeed sufficient to address the memory leak issue.

I've updated the PR to reflect this change.

packages/nextra-theme-docs/src/components/highlight-matches.tsx
```diff
@@ -16,26 +16,27 @@ export const HighlightMatches = memo<MatchArgs>(function HighlightMatches({
...
-      before,
-      <span key={result.index} className="nx-text-primary-600">
-        {after}
-      </span>
-    )
-    index = regexp.lastIndex
+  while ((result = regexp.exec(value))) {
+    if (result.index === regexp.lastIndex) {
+      regexp.lastIndex++
+    } else {
```
dimaMachina, line 27: just curious, behaviour should be the same as previously?
`result.index === regexp.lastIndex` only when `index = 0`?
So you increment `lastIndex` to exit loop?

jaem1n207: The `result.index === regexp.lastIndex` condition in the `highlight-matches` component serves as a safeguard against infinite loops that can occur when the regular expression matches an empty string. Here's a simple example to illustrate this:

```js
const value = 'Docs Theme';
const match = '  ';
const escapedSearch = escapeStringRegexp(match.trim());
const regexp = new RegExp(escapedSearch.replaceAll(/\s+/g, '|'), 'ig'); // -> /(?:)/gi
let result;

while ((result = regexp.exec(value))) {
  if (result.index === regexp.lastIndex) {
    // lastIndex: 0, 1, 2...10.  (value.length === 10)
    // This condition is true when the regex matches an empty string.
    // Without incrementing lastIndex, the loop would be infinite.
    regexp.lastIndex++;
  } else {
    // Process the match...
  }
}
```

In this example, the regex `/(?:)/gi` matches an empty string, and without incrementing `regexp.lastIndex`, the `exec` method would keep matching at the same position, causing an infinite loop.
However, the condition to increment `lastIndex` is not currently necessary in the`highlight-matches` component. This is because, in the `flexsearch` component that utilizes `highlight-matches`, if `search` is an empty or whitespace-only string, `highlight-matches` is not invoked. Therefore, the `match` in `highlight-matches` will never be assigned a value that could trigger an infinite loop, and the condition can be safely removed.
Nonetheless, the `result.index === regexp.lastIndex` condition was included as a precaution to ensure that `highlight-matches` can handle all inputs without falling into an infinite loop, even in scenarios where it is not used within `flexsearch` component.

dimaMachina (approved):
Huge thanks for your fix 🙏

PR #2750, opened 2024-02-28
fix: Invalid webkit style

caffeine-addictt:
# The issue
I noticed that `-webkit-background-clip` was being used in the JSX style prop in the OpenGraph image generation to the nextra site.

`-webkit-background-clip` is an invalid JSX style prop, and `WebkitBackgroundClip` should be used instead.

Although the webkit clip would not matter most of the time as `BackgroundClip` is present, I believe it is still best to use `WebkitBackgroundClip`.

## Example
Here is an example where I copied the image JSX and removed the `backgroundClip` style prop to compare the difference when `-webkit-background-clip` or `WebkitBackgroundClip` is used.

image
image

### DIFF
--- a/docs/pages/api/og.jsx
+++ b/docs/pages/api/og.jsx
@@ -78,7 +78,7 @@ export default async function (req) {
             letterSpacing: -4,
             backgroundImage: 'linear-gradient(90deg, #fff 40%, #aaa)',
             backgroundClip: 'text',
-            '-webkit-background-clip': 'text',
+            WebkitBackgroundClip: 'text',
             color: 'transparent'
           }}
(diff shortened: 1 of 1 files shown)


dimaMachina (commented):
I don't see any difference between your preview https://nextra-v2-q52clspby-shuding1.vercel.app/api/og?title=Syntax%20Highlighting%20Hello%20world and https://nextra.site/api/og?title=Syntax%20Highlighting%20Hello%20world

Maybe we can completely remove this prop?

caffeine-addictt:
Makes sense, I'll update the PR.

PR #2757, opened 2024-03-01
Tailwind imports using path instead of package

Boettner-eric:
The current tailwind config for the Docs theme generates tailwind classes for nextra components that cross the package boundary. This leads to weird behavior if a user wants to modify either nextra or the docs theme without effecting the other as the css is coupled. 

This PR uses the Nextra build assets instead of the source files from the Nextra package. This should make it easier for developers to build in the monorepo without unexpected css issues.

### DIFF
--- a/.changeset/shy-hotels-pay.md
+++ b/.changeset/shy-hotels-pay.md
@@ -1,0 +1,7 @@
+---
+"nextra-theme-docs": patch
+"nextra": patch
+"docs": patch
+---
+
+- use Package instead of Path for taildwind imports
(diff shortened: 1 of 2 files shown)


dimaMachina (changes requested):
I don't understand your described problem, can you share more details about `weird behavior`

Boettner-eric:
If you try building the `nextra-theme-docs` package from the monorepo while using a `nextra` version that isn't in the monorepo the tailwind classes for `nextra`  are pulled from the nextra folder instead of the package itself. If someone modifies the nextra folder or in my case uses a different nextra version the classes aren't read by the tailwind compiler. 

The easiest way to replicate this is to swap the nextra version to a non-workspace one and temporarily remove or rename the nextra folder. The page renders with lots of css issues and its clear that tailwind is missing some of the required classes.

tailwind-issue-example

dimaMachina:
I don't see the valuable reasons to include these changes

Issue #2755, opened 2024-03-01
theme.config.jsx doesn't react on changes


pavliukpetro:
Hello,
first of all, Nextra is really cool. I like it so much.

Just have one question.
I am adding configs to theme.config.jsx file but I don't see on my site.
The only thing that works is "logo" and "project". Other like "useNextSeoProps", "head" and "footer" are not working for me.

Could you please advice what could be wrong?

I am using
"next": "14.1.0",
"nextra": "^3.0.0-alpha.19",
"nextra-theme-docs": "^3.0.0-alpha.19"

Thanks

dimaMachina:
Hi, do you have similar console message while updating theme.config file?
```sh
 ⚠ Fast Refresh had to perform a full reload when ./theme.config.tsx changed. Read more: https://nextjs.org/docs/messages/fast-refresh-reload
```

pavliukpetro:
Hi @dimaMachina haven't seen that. Also tried to restart the project but it had no effect.

pavliukpetro:
Hey @dimaMachina . After a small research I found that some keys in theme.config for Nextra 3 have another names. Like here https://github.com/shuding/nextra/issues/826. Also check the themeSchema to understand what could be used.
Looks like we can close this.

Just a small question. Do you plan to create the documentation for ver 3 and describe all the configs?
Thanks

dimaMachina:
Nextra 4 with App Router has been released, and support for the theme.config was removed, so closing this issue https://the-guild.dev/blog/nextra-4

PR #2765, opened 2024-03-05
Fix leaky styles

### LINKED ISSUE #2764: Global styles are leaky
In [styles.css](https://github.com/shuding/nextra/blob/main/packages/nextra-theme-docs/css/styles.css#L33), global elements like `a`, `button`, and `input` have styling that leak into any custom components supplied:

```css
a,
summary,
button,
input,
[tabindex]:not([tabindex='-1']) {
  @apply nx-outline-none;
  &:focus-visible {
    @apply nx-ring-2 nx-ring-primary-200 nx-ring-offset-1 nx-ring-offset-primary-300 dark:nx-ring-primary-800 dark:nx-ring-offset-primary-700;
  }
}
```

This imposes 2 problems:

1. I have to provide Nextra-specific style overrides for my components (in my case I’m showcasing a design system, and don’t want to have Nextra-specific code shipped to all consumers)
2. For the styles I’m declaring anyway, the Nextra docs have a `0 2 0` specificity that wins out over my `0 1 0` styles (note this only happens in dark mode with `:is(html[class~="dark"] input:focus-visible)`, but our docs/site are all in dark mode)

If we fix this, it will fix issues for anyone writing docs in Nextra for their design system.

drwpow:
## Changes

Fixes #2764 by scoping conflicting global styles to a classname. Now this is scoped to Nextra internal elements, while letting user styles operate as-expected.

CleanShot 2024-03-05 at 15 47 32

### DIFF
--- a/packages/nextra-theme-docs/css/styles.css
+++ b/packages/nextra-theme-docs/css/styles.css
@@ -23,20 +23,12 @@ body {
   @apply nx-w-full nx-bg-white dark:nx-bg-dark dark:nx-text-gray-100;
 }
 
-a,
-summary,
-button,
-input,
-[tabindex]:not([tabindex='-1']) {
+.nextra-focusable {
   @apply nx-outline-none;
+
   &:focus-visible {
     @apply nx-ring-2 nx-ring-primary-200 nx-ring-offset-1 nx-ring-offset-primary-300 dark:nx-ring-primary-800 dark:nx-ring-offset-primary-700;
   }
-}
-
-a,
-summary {
-  @apply nx-rounded;
 }
 
 .nextra-content {
@@ -83,7 +75,7 @@ summary {
 }
 
 @media (prefers-reduced-motion: reduce) and (max-width: 767px) {
-  article:before,
+  article::before,
   .nextra-sidebar-container,
   .nextra-sidebar-container.open,
   body.resizing .nextra-sidebar-container {
@@ -92,7 +84,7 @@ summary {
 }
 
 /* Content Typography */
-article details > summary {
+.nx-summary {
   &::-webkit-details-marker {
     @apply nx-hidden;
   }
--- a/packages/nextra-theme-docs/css/styles.css
+++ b/packages/nextra-theme-docs/css/styles.css
@@ -84,7 +84,7 @@ body {
 }
 
 /* Content Typography */
-.nx-summary {
+.nextra-summary {
   &::-webkit-details-marker {
     @apply nx-hidden;
   }
--- a/.changeset/tame-geckos-rhyme.md
+++ b/.changeset/tame-geckos-rhyme.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+Fix leaky global styles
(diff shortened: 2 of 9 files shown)


packages/nextra-theme-docs/css/styles.css
```diff
@@ -23,22 +23,14 @@ body {
...
   @apply nx-outline-none;
+
   &:focus-visible {
     @apply nx-ring-2 nx-ring-primary-200 nx-ring-offset-1 nx-ring-offset-primary-300 dark:nx-ring-primary-800 dark:nx-ring-offset-primary-700;
   }
 }
 
-a,
-summary {
-  @apply nx-rounded;
```
drwpow, line 39: Removed this because this seemed to be already applied in other areas.

packages/nextra-theme-docs/css/styles.css
```diff
@@ -23,22 +23,14 @@ body {
   @apply nx-w-full nx-bg-white dark:nx-bg-dark dark:nx-text-gray-100;
 }
 
-a,
-summary,
-button,
-input,
-[tabindex]:not([tabindex='-1']) {
```
drwpow, line 30: I couldn’t find any `input` elements that needed this, or `[tabindex]` elements in the codebase. I’m guessing these were just anticipatory styles?

All the `<a>`, `<summary>`, and `<button>` tags generated by Nextra got the `.next-focusable` class.

dimaMachina (changes requested):
but now focus state in header/sidebar/toc is broken

Also, what do you think about approach in this PR? https://github.com/shuding/nextra/pull/2740

https://github.com/shuding/nextra/assets/7361780/169f7088-4adc-4aa0-bf39-23f882851427

drwpow:
> Also, what do you think about approach in this PR? https://github.com/shuding/nextra/pull/2740

I wouldn’t mind that PR landing as well; that’s actually unrelated to this fix. That PR as I understand it is more resolving the cascade among internal styles. My issue is I’m using Nextra to document a design system, so I want to avoid _any_ style overrides whatsoever, because then I’d have to ship Nextra-specific style overrides to everywhere the DS is used.

Scoping, and just having no global element styles, would be my preferred fix. And using `@layers` or not wouldn’t affect that. But good catch on the missing elements! Will update those if you are OK with this approach to the solution (and open to alternate methods too!)

dimaMachina:
@drwpow thank you, to be honest, I don't have time to maintain nextra v2 and v3 versions, I update v2 only for critical bugs to avoid fixing huge git conflicts

Could you point your PR to v3 branch and only fix git conflicts

I will take care of rest and will push my changes directly on your PR

drwpow:
> Could you point your PR to v3 branch and only fix git conflicts

Will do! Will just open up another PR to make it easy.

> @drwpow thank you, to be honest, I don't have time to maintain nextra v2 and v3 versions, I update v2 only for critical bugs to avoid fixing huge git conflicts

Totally understand! ❤️ Thanks for all your amazing work on nextra. Setting boundaries on open source maintenance is necessary, and I respect and understand completely. It’s a smart thing to do.

dimaMachina:
Thank you for your compliments and for your contributions to make it better

More excited to ship Nextra 4 in 2024 with app router support and server components 🙌

drwpow:
Superceded by #2755

PR #2766, opened 2024-03-06
fix(toc): `backToTop` can't work alone

Aruelius:


dimaMachina (approved):
will be fixed only in v3

.changeset/popular-dancers-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+`backToTop` is hidden when enabled when `toc.extraContent`, `feedback.content` and `editLink.component` are falsy
```
dimaMachina, line 5:
```suggestion
`backToTop` is hidden when enabled but when `toc.extraContent`, `feedback.content` and `editLink.component` are falsy 
```

PR #2775, opened 2024-03-08
Fix leaky styles

### LINKED ISSUE #2764: Global styles are leaky

drwpow:
## Changes

Fixes #2764. Supercedes #2765.

Also fixes user-generated content in landmark elements (`nav`, `footer`, etc.)

CleanShot 2024-03-08 at 10 19 31@2x

CleanShot 2024-03-08 at 10 19 42@2x

### DIFF
--- a/packages/nextra-theme-docs/css/styles.css
+++ b/packages/nextra-theme-docs/css/styles.css
@@ -22,19 +22,18 @@ body {
   @apply _w-full dark:_text-gray-100;
 }
 
-a,
-summary,
-button,
-input,
-[tabindex]:not([tabindex='-1']) {
+.nextra-focusable,
+.nextra-landmark a,
+.nextra-landmark button,
+.nextra-landmark input {
+  @apply _outline-none;
   &:focus-visible {
     @apply _outline-none;
     @apply _ring-2 _ring-primary-200 _ring-offset-1 _ring-offset-primary-300 dark:_ring-primary-800 dark:_ring-offset-primary-700;
   }
 }
 
-a,
-summary {
+.nextra-landmark a {
   @apply _rounded;
 }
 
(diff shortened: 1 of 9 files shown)


packages/nextra-theme-docs/css/styles.css
```diff
@@ -22,19 +22,18 @@ body {
   @apply _w-full dark:_text-gray-100;
 }
 
-a,
-summary,
-button,
-input,
-[tabindex]:not([tabindex='-1']) {
+.nextra-focusable,
+.nextra-landmark a,
```
drwpow, line 26: These classnames I am not particular about; rename to anything else if they are more consistent with existing styles

dimaMachina (commented):
Hey! Thank you! Sorry for delay, I will include your PR with my additional fixes in next v3 alpha release 🙏

PR #2783, opened 2024-03-13
feat(theme-docs): add ability to add custom links to navbar like project & chat links

itsmaty:
Previously you could add only two (project & chat link), you could change the icon & link to use these for completely other links but the internal names always where chat & project link and as said you where hardcoded limited to two links which in my opinion was a bad design decision, maybe there was a reason when it was developed.
image

Now you can add as many links with icons as you want while still supporting "chat" & "project" in the config for backwards compatibility.

You can now use "links" in the config which is an array of link components with and "link" and an "icon" property.
Internally this array just gets mapped over and renders all link components. To have said backwards compatibility the icon & chat link (if set), just get pushed to the links array.

Maybe there is a better solution than using useEffect to pus project & chat link to the links array as you can see i did in the code.

I tested the functionality, works great.
image

Would love to see this feature in v3.

Best regards,
Matthias


dimaMachina (commented):
why not to use `navbar.extraContent` to add customizable content?

itsmaty:
Technically you can use `navbar.extraContent` to add more "icon buttons", but its not a clear way when you intend to add more buttons and you have to add the linking functionality yourself.

I just think its easier that if you want to add more buttons you have an config where you can add a button object and it just works.

It's your decision whether you think it improves nextra and is useful or not. For me it is.

dimaMachina (commented):
Thank you for your PR, I would like to stay with the current approach for now

PR #2784, opened 2024-03-13
Add langfuse to showcase

marcklingen:
Thanks for building nextra! I've recommended it to so many friends and often shared our docs repo which includes all of marketing/blog/docs built with Nextra: https://github.com/langfuse/langfuse-docs

Hope you like our little page as a showcase!


dimaMachina (approved):
Thank you! Your customization of docs theme looks nice!

Issue #2785, opened 2024-03-14
Inconsistent Navbar breakpoint


arno-fukuda:
The navbar breakpoint to hamburger menu kicks it too late under certain conditions (too many categories), pushing the logo out of viewport.

**Good**:
<img width="1362" alt="Screenshot 2024-03-14 at 19 55 52" src="https://github.com/shuding/nextra/assets/133849024/034b0fa3-276b-42e4-abe6-66dc628eeb42">
---

**Bad**:
<img width="889" alt="Screenshot 2024-03-14 at 19 56 24" src="https://github.com/shuding/nextra/assets/133849024/b4e3c83b-a5d0-4136-a1a4-96e2b6d23502">
---

**Worse**:
<img width="773" alt="Screenshot 2024-03-14 at 19 56 39" src="https://github.com/shuding/nextra/assets/133849024/0ad54f66-1bcc-43b5-ae30-766e9e436488">
---

**Reproduction:**
`_meta.json` in `pages/`
```
{
  "index": "Introduction",
  "another": "Another Page",
  "advanced": "Advanced (A Folder)",
  "about": {
    "title": "About",
    "type": "page"
  },
  "contact": {
    "title": "Contact ↗",
    "type": "page",
    "href": "https://twitter.com/shuding_",
    "newWindow": true
  },
  "test": {
    "title": "Contact ↗",
    "type": "page",
    "href": "https://twitter.com/shuding_",
    "newWindow": true
  },
  "test2": {
    "title": "Contact ↗",
    "type": "page",
    "href": "https://twitter.com/shuding_",
    "newWindow": true
  },
  "test3": {
    "title": "Contact ↗",
    "type": "page",
    "href": "https://twitter.com/shuding_",
    "newWindow": true
  },
  "test4": {
    "title": "Contact ↗",
    "type": "page",
    "href": "https://twitter.com/shuding_",
    "newWindow": true
  },
  "test5": {
    "title": "Contact ↗",
    "type": "page",
    "href": "https://twitter.com/shuding_",
    "newWindow": true
  },
  "test6": {
    "title": "Contact ↗",
    "type": "page",
    "href": "https://twitter.com/shuding_",
    "newWindow": true
  }
}

```

`theme.config.tsx`
```
const logo = (
  <svg
    height='20'
    viewBox='0 0 361 70'
    fill='none'
    xmlns='http://www.w3.org/2000/svg'
  >
    <path
[...]

PR #2787, opened 2024-03-14
fix: Cut the navbar items when it's too long (#2785)

RiverTwilight:
This pull request addresses issue #2785, ensuring that when the navbar has many items or items with long names, the overflow is managed elegantly without pushing the logo out of view.

Previously, an overflow of menu items caused layout issues, leading to a cluttered and inaccessible navbar.

## Before

image

## After

Screenshot 2024-03-14 at 21 02 41

It's important to note that for most configurations, the visual appearance of the navbar remains unchanged:

Screenshot 2024-03-14 at 21 04 40

### DIFF
--- a/packages/nextra-theme-docs/src/components/navbar.tsx
+++ b/packages/nextra-theme-docs/src/components/navbar.tsx
@@ -93,7 +93,7 @@ export function Navbar({ flatDirectories, items }: NavBarProps): ReactElement {
           'contrast-more:nx-shadow-[0_0_0_1px_#000] contrast-more:dark:nx-shadow-[0_0_0_1px_#fff]'
         )}
       />
-      <nav className="nx-mx-auto nx-flex nx-h-[var(--nextra-navbar-height)] nx-max-w-[90rem] nx-items-center nx-justify-end nx-gap-2 nx-pl-[max(env(safe-area-inset-left),1.5rem)] nx-pr-[max(env(safe-area-inset-right),1.5rem)]">
+      <nav className="nx-mx-auto nx-flex nx-h-[var(--nextra-navbar-height)] nx-max-w-[90rem] nx-items-center nx-justify-between nx-gap-2 nx-pl-[max(env(safe-area-inset-left),1.5rem)] nx-pr-[max(env(safe-area-inset-right),1.5rem)]">
         {config.logoLink ? (
           <Anchor
             href={typeof config.logoLink === 'string' ? config.logoLink : '/'}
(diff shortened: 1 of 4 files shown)


dimaMachina (changes requested):
now dropdown with `type: 'menu'` is hidden while opening

https://github.com/shuding/nextra/assets/7361780/c892a367-0c8c-4d50-a91e-dc1c9a82f1f5

packages/nextra-theme-docs/src/components/navbar.tsx
```diff
@@ -106,61 +106,71 @@ export function Navbar({ flatDirectories, items }: NavBarProps): ReactElement {
...
-              <span className="nx-absolute nx-inset-x-0 nx-text-center">
-                {page.title}
-              </span>
-              <span className="nx-invisible nx-font-medium">{page.title}</span>
-            </Anchor>
-          )
-        })}
+
+        <div className="nx-flex-auto nx-overflow-auto">
+          <div className="nx-flex nx-flex-row-reverse">
```
dimaMachina, line 111: seems like now all links are in reversed order

RiverTwilight:
@dimaMachina The problems you mentioned are addressed. Currently if the menu is too long the search bar will be pushed out of the viewport, and the whole page will be scrollable on x-axis, which it's the best workaround, I think.

Screenshot 2024-03-15 at 16 35 18

arno-fukuda:
> @dimaMachina The problems you mentioned are addressed. Currently if the menu is too long the search bar will be pushed out of the viewport, and the whole page will be scrollable on x-axis, which it's the best workaround, I think.
> 
> Screenshot 2024-03-15 at 16 35 18

I might be overcomplicating things but ideally I'd like to see the hamburger kick in as soon as the sum of the total navbar items width starts exceeding the current viewport width.

dimaMachina:
> > @dimaMachina The problems you mentioned are addressed. Currently if the menu is too long the search bar will be pushed out of the viewport, and the whole page will be scrollable on x-axis, which it's the best workaround, I think.
> > Screenshot 2024-03-15 at 16 35 18
> 
> I might be overcomplicating things but ideally I'd like to see the hamburger kick in as soon as the sum of the total navbar items width starts exceeding the current viewport width.
> 
> So if someone created 100 categories, it would immediately start off as hamburger menu, even on fullscreen.

this would be a perfect fix, not sure how to properly implement it, without overcomplicating things, any ideas are welcome

Also, I would accept the solution with `overflow: scroll` for nav links in navbar

RiverTwilight:
> > > @dimaMachina The problems you mentioned are addressed. Currently if the menu is too long the search bar will be pushed out of the viewport, and the whole page will be scrollable on x-axis, which it's the best workaround, I think.
> 
> > > Screenshot 2024-03-15 at 16 35 18
> 
> > 
> 
> > I might be overcomplicating things but ideally I'd like to see the hamburger kick in as soon as the sum of the total navbar items width starts exceeding the current viewport width.
> 
> > 
> 
> > So if someone created 100 categories, it would immediately start off as hamburger menu, even on fullscreen.
> 
> 
> 
> this would be a perfect fix, not sure how to properly implement it, without overcomplicating things, any ideas are welcome
> 
> 
> 
> Also, I would accept the solution with `overflow: scroll` for nav links in navbar

Yes. I've thought about this idea but this might need JS to intercept, which makes the Navbar heavier, since we need dynamically measure the width of the available size and the horizontal stack width. Whatever I'll give another shot soon. There is no native CSS workaround to do this as far as I know.

The `overflow` solution seems not working as this cut dropdown menu (even only `overflow-x` is set to "scroll").

RiverTwilight:
@dimaMachina Now it should be the expected behavior you want. If the items are too long it will be moved to sidebar, which's similar to mobile screen.

Screenshot 2024-03-19 at 13 55 50

Since the sidebar & hamburger share the same content, it's oevrcomplicated to let the hamburger contain the nav items and sidebar remains unchanged under overflow context.

Feel free to drop any comment!

arno-fukuda:
> If the items are too long it will be moved to sidebar

I believe this won't work for pages that use raw or full layout theme which is often used for landing pages.
```
{
  "index": {
    "title": "Home",
    "theme": {
      "layout": "raw"
    }
  }
}
```

dimaMachina:
superseded by https://github.com/shuding/nextra/pull/3376

PR #2854, opened 2024-04-27
feat: add more social icon

ikhsangama:
add more some common social media icon


Netail:
Facebook & Twitter in that case maybe too?

glmn:
Telegram 🙏

Netail:
> > Facebook, Twitter, Whatsapp, Threads in that case maybe too?
> 
> @Netail isn't Twitter icon already reserved here as 'X'? Do we need legacy logo?

Woops, you're right hahaha. Still didn't got used to it being called X now

Netail:
Terima kasih 🙏

dimaMachina:
thanks, but I don't think I would like to merge it, I would like to keep icons minimal and would suggest you to use icons library or copy-paste needed icons directly for your project

PR #2884, opened 2024-05-12
Add topLevelLink option

vordgi:
When type is set as page - the page is displayed in the navbar. I can add `display: hidden` and it will be hidden from all navigations. However, sometimes you want to hide this page only in the navbar itself, and instead, for example, add it to the menu.

As an example - my website https://nimpl.tech/. I want to split the documentation, but keep them in dropdown lists.


dimaMachina (commented):
Hi, sorry, I don't think I want to include this

PR #2906, opened 2024-05-28
fix(Card): change TW from justify-start to justify-center

arno-fukuda:
The Card components <a> tag is using justify-start. This is not necessary since the nested span already has it.
When used with the `Cards` grid component, the layout becomes inconsistent across the rows.

Changing to justify-center to ensure the span content remains centered across a row.

**Before**
before

**After**
after


dimaMachina:
this is intentional behaviour to align to top, otherwise cards become like 
image

PR #2921, opened 2024-06-07
Updated Theme color properties on theme-configuration.mdx

chrisroselli:


dimaMachina (commented):
thanks, but this is intentional, in TypeScript we use `;` for types

image

PR #2922, opened 2024-06-07
Omit `...{:type}` annotations from search index

### LINKED ISSUE #2913: [v3] Extra {:foo} annotations not stripped from search text
When writing something like 
```
`<script></script>{:html}`
```
the `{:html}` appears as part of the search text when using the search box. It should be stripped out before the text is added to the search box.

siefkenj:
Fixes #2913 by omitting the `{:type}` annotations from the search index when they appear in inline code.


packages/nextra/src/server/remark-plugins/remark-structurize.ts
```diff
@@ -56,10 +59,18 @@ export const remarkStructurize: Plugin<[Search], Root> = options => {
...
+      type === 'inlineCode' ||
+      type === 'tableCell'
     ) {
-      result += node.value
-      if (!skip) content += node.value
+      // Because of markdown extensions, inline code may have a `{:type}` suffix.
+      // Trim this off if it exists.
+      const value =
+        type === 'inlineCode'
+          ? node.value.replace(/\s*\{:\w+\}$/, '')
```
dimaMachina, line 70:
```suggestion
      // Inline code may have a `{:language}` suffix, trim this off if it exists.
      const value =
        type === 'inlineCode'
          ? node.value.replace(/\{:\w+\}$/, '')
```

i think `\s*` is useless here

siefkenj:
@dimaMachina Are there any other changes needed on this PR?

dimaMachina:
I am currently busy with other tasks, no worries I will include this change in next release when I will back to clean up Nextra for contributions

dimaMachina (approved):
LGTM, sorry for taking the time to merge this!

PR #2940, opened 2024-06-19
Add ability to dynamically hide menu / nav items

chriswiggins:
Hi there!

Loving what you're doing with the v3 branch. We have a use-case where we want to dynamically hide / show certain pages (we have some docs behind auth).

It isn't possible to dynamically do checks as to whether these pages are shown or not, so I've added a config option that takes a function, that allows you to check if the menu item should be rendered or not (both for the Navbar and Sidebar)

It is used like so in your `theme.config.tsx`:

```typescript
export default {
  menuItems: {
    visibilityCheck: (item: PageItem | Item) => {
      if (item.route === '/hidden-page') {
        return false;
      }

      return true;
    }
  }
}
```

I thought a better route would be supporting `_meta.tsx` with a dynamic visibility type, but that would require the meta file supporting items as functions and I'm not sure how to do that 😄 

Feedback appreciated!

### DIFF
--- a/packages/nextra-theme-docs/src/components/navbar.tsx
+++ b/packages/nextra-theme-docs/src/components/navbar.tsx
@@ -88,6 +88,8 @@ export function Navbar({ items }: NavBarProps): ReactElement {
   const activeRoute = useFSRoute()
   const { menu, setMenu } = useMenu()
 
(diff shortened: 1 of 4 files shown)


dimaMachina:
please create an issue first

PR #2945, opened 2024-06-25
Provide themes without any wrappers

mayank1513:
Advantage
- It makes it easy to switch to next.js app router and utilize server components
- Wrapper would force everything to be client component.
- Has more granular configuration for theme and colorScheme
- Possibility of adding theme sets - i.e., dark and light variants of each theme.


dimaMachina:
Sorry I am not interested in this change

PR #2949, opened 2024-06-26
feat: TOC and Sidebar override [v3]

andrii-bodnar:
This PR includes improvements to the TOC and Sidebar components that can be overridden through the theme config.

**1. TOC component export to allow the following:**

```ts
toc: {
  title: "Custom TOC title",
  backToTop: false,
  float: true,
  component(props: ComponentProps<typeof TOC>) {
    return (
      <div className="flex items-end justify-between">
        <div>Custom element</div>
        <TOC {...props} />
      </div>
    );
  }
}
```

This will allow us to add a wrapper around the built-in TOC component.

**2. A possibility to set a custom component for the Sidebar component + its export:**

```ts
sidebar: {
  component(props: ComponentProps<typeof Sidebar>) {
    return <div className="flex items-end justify-between">
      <div>Custom element</div>
      <Sidebar {...props}/>
    </div>
  },
  toggleButton: false
},
```

The same approach already exists for a number of other components, so I think it would be quite useful and fair to have the same DX for the TOC and Sidebar components.

### DIFF
--- a/.changeset/silver-chicken-bathe.md
+++ b/.changeset/silver-chicken-bathe.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': minor
+---
+
+TOC and Sidebar override
--- a/docs/pages/docs/docs-theme/theme-configuration.mdx
+++ b/docs/pages/docs/docs-theme/theme-configuration.mdx
@@ -418,6 +418,11 @@ Customize the entire navbar component.
 <OptionTable
   options={[
     [
+      'sidebar.component',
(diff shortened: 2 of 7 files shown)


andrii-bodnar:
Hi @dimaMachina, could you please take a look? Would it be possible to include this in the `3.0.0-alpha.25` release? Thanks!

dimaMachina:
Sorry I am not interested in overriding the sidebar and TOC, you probably should use a custom theme

Issue #2917, opened 2024-06-06
v3 back-to-top button does not scroll all the way to top


ZYJLiu:
Clicking "scroll to top" will scroll up until the button toggles to hidden, which causes scrolling to stop. 
Solution may be to keep "scroll to top" button visible until scrolling is complete.

For example, you can test on the graphql docs pages: https://graphql.org/learn/queries/

![CleanShot 2024-06-05 at 21 43 41](https://github.com/shuding/nextra/assets/75003086/49f77d2b-8142-406b-abf5-ec464ad9f39f)

dimaMachina:
upstream https://github.com/facebook/react/issues/20770

PR #2979, opened 2024-07-17
[v3]: Fixed scroll to top button bug

Shubham-EV:
Fixed scroll-to-top button bug discussed here [#2917](https://github.com/shuding/nextra/issues/2917)


dimaMachina:
thanks for the attempt, but this solution is too overcomplicated, fixed by https://github.com/shuding/nextra/pull/2981

PR #3143, opened 2024-08-27
[v3] remove explicit `ZodError` type assertion

87xie:
## Description

This change allows TypeScript to infer the `ZodError` type, so we don't need to import it for explicit assertion. Feel free to close this if the approach is not suitable.


dimaMachina (approved):
Nice 👍

PR #3149, opened 2024-08-28
[v3] avoid focus-visible style being cut off by overflow hidden

87xie:
fixes https://github.com/shuding/nextra/issues/1432

## Description

I noticed some adjustments to the focus-visible style in [Auth.js](https://authjs.dev/) related to Nextra, such as:

- [tab button](https://github.com/nextauthjs/next-auth/blob/538333d86e93e0e4019d117d07741e02f9581c2b/docs/pages/global.css#L177-L180)
- [sidebar](https://github.com/nextauthjs/next-auth/blob/538333d86e93e0e4019d117d07741e02f9581c2b/docs/pages/global.css#L207-L210)

I think these adjustments could be applied more comprehensively within Nextra. For instance, Auth.js only targets the focus-visible style of the first-level sidebar items and the focus-visible style is not working well in the TOC either.

image

image

image

## Screenshots

**Sidebar**

image

image

**Tabs**

image

**TOC**

image

image

**Direction: rtl**

image

image

PR #3150, opened 2024-08-29
[v3 ] adjust the sidebar auto-collapse animation

87xie:
## Description

I noticed that enabling sidebar auto-collapse causes menus above level 2 to unexpectedly collapse during keyboard navigation, as shown in the video below.

I hope we can resolve #3149 first, and then I will make adjustments to this PR 🙏

##  How

- The `<FolderImpl>` item has a focusable element, but it does not record `item.route` in `FocusedItemContext` when focused, so related collapses are hidden.
- When the current item is blurred, it updates the `focused` state in `FocusedItemContext` to `null`. When the next item gains focus, it updates the corresponding `item.route`. This causes `focusedRouteInside` to experience unnecessary changes, resulting in unexpected animations

I set a debounced callback in the sidebar focusout event to check if `setFocused(null)` is needed.
## Screenshots

**Before**

https://github.com/user-attachments/assets/a9fd29bb-116d-4d0d-a93f-d88db77f66af

**After**

https://github.com/user-attachments/assets/7e5329be-dcfd-4f55-82f5-c6ef49153a83

### DIFF
--- a/packages/nextra-theme-docs/src/components/sidebar.tsx
+++ b/packages/nextra-theme-docs/src/components/sidebar.tsx
@@ -3,7 +3,7 @@ import type { Heading } from 'nextra'
 import { useFSRoute, useMounted } from 'nextra/hooks'
 import { ArrowRightIcon, ExpandIcon } from 'nextra/icons'
 import type { Item, MenuItem, PageItem } from 'nextra/normalize-pages'
-import type { ReactElement } from 'react'
+import type { ReactElement, ReactNode } from 'react'
 import {
   createContext,
   memo,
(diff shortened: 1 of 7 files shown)


packages/nextra-theme-docs/src/components/sidebar.tsx
```diff
@@ -173,6 +237,12 @@ function FolderImpl({ item, anchors }: FolderProps): ReactElement {
           TreeState[item.route] = !open
           rerender({})
         }}
+        onFocus={() => {
+          focusControl?.setFocused(item.route)
+        }}
+        onBlur={() => {
+          focusControl?.setBlurred(item.route)
+        }}
```
dimaMachina, line 245: Nice catch and reactivity to fix this issue (I also discovered it yday 😀). I think simple omitting `onBlur` handler should fix this confusing behaviour when `autoColapse` is set to `true`

87xie: My thought is that if we omit `onBlur`, the collapse may not automatically close when keyboard navigation shifts focus away from a sidebar item, such as switching to the locale switch.

dimaMachina: I think it's fine to have this case, instead of introducing complexity with `setTimeout`s IMO

87xie: Yes, I agree the setTimeout approach isn't ideal. I will fix it in the next commit.

87xie: I think we can remove the `onBlur` prop (regardless of whether `sidebar.autoCollapse` is enabled).

By using the sidebar's `focusout` event (currently the DOM referenced in `sidebarRef`), the use of `setTimeout` becomes meaningful, as we only need to check if `document.activeElement` is still within the sidebar during the most recent `focusout` in sidebar.

This logic is consolidated within a single `useEffect`, which also reduces the re-renders triggered by the `onBlur` when calling `setFocused(null)`. Does this approach look good to you?

```ts
useEffect(() => {
  const sidebarEl = sidebarRef.current;
  if (!sidebarEl) return;
  let timer: ReturnType<typeof setTimeout>;

  const handleFocusout = () => {
    clearTimeout(timer);
    timer = setTimeout(() => {
      if (sidebarEl?.contains(document.activeElement)) return;
      setFocused(null);
    }, 50);
  };

  sidebarEl.addEventListener('focusout', handleFocusout);
  return () => {
    clearTimeout(timer);
    sidebarEl.removeEventListener('focusout', handleFocusout);
  };
}, []);
```

dimaMachina:
closed by https://github.com/shuding/nextra/pull/3157, thanks for initial attempt as well 👍

Issue #3099, opened 2024-08-04
Style Conflict between Nextra and TailwindCSS


vineetver:
#### Description
We are experiencing a style conflict when using Nextra with TailwindCSS. Specifically, Nextra's focus-visible styles are causing unexpected styling issues.

#### Steps to Reproduce
1. Create a new project using Nextra.
2. Add TailwindCSS to the project.
3. Add code for Menu.

#### Example

The dropdown menu items styled with TailwindCSS are being overridden/merged by Nextra's styles, resulting in an unwanted outline, ring, or border:

<img width="467" alt="Screenshot 2024-08-04 at 12 47 54 AM" src="https://github.com/user-attachments/assets/053cbb23-db41-4466-b74a-0735a962c7fc">

For instance, this is Nextra's focus style:

<img width="315" alt="Screenshot 2024-08-04 at 12 54 37 AM" src="https://github.com/user-attachments/assets/03ecad1f-47b9-430d-8231-09ecba91b064">

#### Code

`menu.tsx`

```
import * as React from "react"
import * as DropdownMenuPrimitive from "@radix-ui/react-dropdown-menu"
import {Check, Circle} from "lucide-react"
import {cn} from "~/lib/util";

const Menu = DropdownMenuPrimitive.Root

const MenuTrigger = DropdownMenuPrimitive.Trigger

const MenuPortal = DropdownMenuPrimitive.Portal

// Left/right padding 12px
const MenuContainer = React.forwardRef<
    React.ElementRef<typeof DropdownMenuPrimitive.Content>,
    React.ComponentPropsWithoutRef<typeof DropdownMenuPrimitive.Content>
>(({className, sideOffset = 4, ...props}, ref) => (
    <DropdownMenuPrimitive.Portal>
        <DropdownMenuPrimitive.Content
            ref={ref}
            sideOffset={sideOffset}
            className={cn(
                "z-50 min-w-[112px] max-w-[280px] overflow-hidden rounded-[4px] shadow-md border bg-surface-container-low py-[4px] text-on-surface data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
                className
            )}
[...]

PR #3154, opened 2024-08-29
[v3] reduce global style conflicts with external components

87xie:
## Description

Hi teams,

I've referenced the following issues: #2613, #2764, and #3099. This change aims to reduce conflicts with outer-land styles and components.

I introduce cascade layers that unlayered styles can take precedence over Nextra's predefined styles. This approach seems to resolve issues like https://github.com/shuding/nextra/issues/747#issuecomment-1341362631. At the same time, it prevents Nextra's unlayered styles from inadvertently overriding user-defined CSS layers.

The current priority order is as follows:

- `nextra-utilities` (highest)
- `nextra-components`
- `nextra-base` (lowest)

To avoid conflicts with the focus-visible style in external components, I added a `nextra-focus-ring` class to replace the original element selector. Users can choose in the theme config's `focusRing` option whether to apply it to all elements or selectively opt out using the `not-nextra-focus-ring` class name.

As shown in the image below, when the `focusRing` option is `false`, the focus-visible style comes from the user agent settings.

I believe this modification is more aligned with the design system docs. As seen with [kuma-ui](https://www.kuma-ui.com/docs/Components/Input), some components are currently affected by Nextra's focus ring, which might confuse some users.

Feel free to close this if the approach is not suitable.

image

### DIFF
--- a/.changeset/wild-llamas-prove.md
+++ b/.changeset/wild-llamas-prove.md
@@ -1,0 +1,7 @@
+---
+'nextra-theme-blog': patch
+'nextra-theme-docs': patch
+'nextra': patch
+---
+
+reduce focus-visible style conflicts with external components.
--- a/.changeset/wild-llamas-prove.md
+++ b/.changeset/wild-llamas-prove.md
@@ -4,4 +4,4 @@
 'nextra': patch
 ---
 
-reduce focus-visible style conflicts with external components.
+reduce global style conflicts with external components.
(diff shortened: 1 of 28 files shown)


87xie:
I will make some modifications for fewer changes and better consistency, so I have turned it back to a draft. 🙏

dimaMachina:
@87xie I am OOO currently and will go to the graphql conf next week, will review once will have some free time

87xie:
Thanks for letting me know! No rush. I’ll go ahead and close this PR for now and split the content into separate PRs. This should make it easier to review and help avoid conflicts.

PR #3156, opened 2024-08-30
[v3] enhance search input to better support CJK language users

87xie:
## Description

Resolves #2833 

This pull request aims to provide better support for CJK language users while ensuring that the existing search behavior for non-Pinyin characters remains unaffected.

I referred to the [foxact](https://foxact.skk.moe/use-composition-input) implementation and changed the search input to be uncontrolled to ensure that the state in `<Flexsearch>` only updates the composition process is complete, allowing the input state in the browser to be reflected to the user.

To better understand the changes in this PR, here's an overview of the sequence of events when composing the character "你" using Pinyin:

Chrome

```
compositionstart
input (value: n, isComposing: true)
input (value: i, isComposing: true)
input (value: 1, isComposing: true)
input (value:你, isComposing: true)
composition end
```

Firefox

```
compositionstart
input (value: n, isComposing: true)
input (value: i, isComposing: true)
input (value: 1, isComposing: true)
compositionend
input (value: 你, isComposing: false)
```

## Screenshots

<details>
<summary>Pinyin characters</summary>

https://github.com/user-attachments/assets/f2f602d4-aff4-4944-a06f-18f5104d78c6
</details>

<details>
<summary>Non-Pinyin characters</summary>

https://github.com/user-attachments/assets/f4f2319d-0a32-4f22-819c-90932dc7e219
</details>

PR #3157, opened 2024-08-31
[v3] Avoid the sidebar collapse having unintended animations when `sidebar.autoCollapse` is set to `true`

dimaMachina:
another approach to fix https://github.com/shuding/nextra/pull/3150 because I think adding `sidebarEl.addEventListener` / `sidebarEl.removeEventListener` makes things complex – we can ignore the case when the folder will not be closed when switching to the locale switch for example

cc @87xie wdyt

https://github.com/user-attachments/assets/71730838-90d8-4de7-aeff-78f654aa4b04

### DIFF
--- a/packages/nextra-theme-docs/src/components/search.tsx
+++ b/packages/nextra-theme-docs/src/components/search.tsx
@@ -216,7 +216,7 @@ export function Search({
   const handleChange = useCallback(
     (e: ChangeEvent<HTMLInputElement> | CompositionEvent<HTMLInputElement>) => {
       if (!compositionStateRef.current.compositioning) {
-        const { value } = e.target as HTMLInputElement
+        const { value } = e.currentTarget
         onChangeProp(value)
         setShow(Boolean(value))
         compositionStateRef.current.emitted = true
--- a/.changeset/fair-hounds-watch.md
+++ b/.changeset/fair-hounds-watch.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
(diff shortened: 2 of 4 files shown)


packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -216,7 +216,7 @@ export function Search({
   const handleChange = useCallback(
     (e: ChangeEvent<HTMLInputElement> | CompositionEvent<HTMLInputElement>) => {
       if (!compositionStateRef.current.compositioning) {
-        const { value } = e.target as HTMLInputElement
+        const { value } = e.currentTarget
```
dimaMachina, line 219: just to avoid type assertion from https://github.com/shuding/nextra/pull/3156

87xie:
I like the simplicity of the approach, especially if we can ignore clearing the focused state in `FocusedItemContext` when the item loses focus.

Here are the scenarios I can think of that differ from the current version:

- When opening a folder via keyboard navigation, the user clicks on another element on the screen.
- When the last item is an item within a folder

dimaMachina:
> When a user opens a folder via keyboard navigation and then clicks on another element on the screen.
yeah, exactly, I don't think that https://github.com/shuding/nextra/pull/3150 we need to close folder for this case

https://github.com/user-attachments/assets/22cc2bf2-7dbe-49b2-8efb-27d0bd37b5c2

dimaMachina:
> When the last item is an item within a folder

yes 👍, the below case from your pr we can leave to be open too

https://github.com/user-attachments/assets/459a4704-06f4-450a-9c44-b0129ce6997c

PR #3160, opened 2024-08-31
[v3] avoid focus-visible style being cut off by overflow-hidden

dimaMachina:
fixes https://github.com/shuding/nextra/issues/1432

different approach to fix https://github.com/shuding/nextra/pull/3149 cc @87xie

## before 

https://github.com/user-attachments/assets/598ce6d8-0263-41a9-8bde-4c2531cd1edc

## after

https://github.com/user-attachments/assets/eec147b4-fd12-4579-bf32-06c9bf996be8

### DIFF
--- a/.changeset/short-icons-impress.md
+++ b/.changeset/short-icons-impress.md
@@ -1,0 +1,6 @@
+---
+'nextra-theme-docs': patch
+'nextra': patch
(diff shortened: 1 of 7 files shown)


87xie:
This is 👍

PR #3177, opened 2024-08-31
[v3] remove intersection-observer polyfill

87xie:
## Description

Nextra v3 requires Next.js >= 13. Based on the [Supported Browsers](https://github.com/vercel/next.js/pull/41529/files#diff-7e079a6684a14ec2884b491ffdb0ae0f1a61e68e8127b4e00e71acf214f4a98d) for Next.js 13, it seems that most users don't need `intersection-observer` polyfill.

Is it a good opportunity to remove this dependency in v3? If users require support for older browsers, we can suggest they import it in `_app.tsx`.

### DIFF
--- a/.changeset/clever-dots-unite.md
+++ b/.changeset/clever-dots-unite.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+Remove intersection-observer polyfill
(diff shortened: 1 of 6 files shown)


dimaMachina (approved):
Great 🎉 minus almost 3kb in bundle size 🙌

### 🎉  Global Bundle Size Decreased
  
Page | Size (compressed) | 
|---|---|
| `global` | `149.72 KB` _(🟢 -2.83 KB)_ |

Issue #2833, opened 2024-04-09
On Firefox, search box can't input and match Chinese files


CaryiWong:
Windows Firefox 121.0.1 (64 x)
Input Method: Sogou or Microsoft Input Method

When entering Chinese, the search box will display the corresponding pinyin. When I press "space" or "number" to select words, the input box does not respond and matches with pinyin, resulting in matching failure。
I’ve already used the latest nextjs and nextra. Please help fix it out ~~

![error1](https://github.com/shuding/nextra/assets/10609679/05aeedf7-bddb-4a28-a366-c984c5393111)

CaryiWong:
It also appeared at https://nextra.site/ while visiting by Firefox

yyrc:
我测试过自己的nextra，估计是firefox的问题。而且我测试过在firefox使用vitepress用中文搜索，也有问题，估计不会修复了。毕竟现在用chrome的人多太多。

gitforziio:
I have encountered this situation as well.

Perhaps modifying [this function](https://github.com/shuding/nextra/blob/e1574f06b5d8c0377c2e6f74bd30deb4d5198521/packages/nextra-theme-docs/src/components/flexsearch.tsx#L261) can resolve the issue.

dimaMachina:
I don't know Chinese, PR is welcome, some work with the Chinese language was done at https://github.com/shuding/nextra/pull/2124

dimaMachina:
fixed in https://github.com/shuding/nextra/releases/tag/nextra-theme-docs%403.0.0-alpha.34

PR #3186, opened 2024-09-01
[v3] migrate search to `@headlessui/react` Combobox

dimaMachina:


packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -241,85 +117,77 @@ export function Search({
...
+            <Fragment key={id}>
+              {prefix}
+              <ComboboxOption
+                as={NextLink}
+                value={route}
+                href={route}
+                className={cn(
+                  '_mx-2.5 _break-words _rounded-md',
+                  'contrast-more:_border',
+                  'data-[focus]:_ring data-[focus]:_bg-primary-500/10 data-[focus]:_text-primary-600 data-[focus]:contrast-more:_border-primary-500',
```
dimaMachina, line 178:
```suggestion
                  'data-[focus]:_ring data-[focus]:_bg-primary-500/10 data-[focus]:!_text-primary-600 data-[focus]:contrast-more:_border-primary-500',
```

to change text primary color in dark mode too

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -241,85 +117,77 @@ export function Search({
...
-      )}
+  const handleSelect = useCallback(
+    (route: string) => {
+      if (route) {
+        router.push(route)
+        setMenu(false)
+      }
+    },
+    [router, setMenu]
+  )
```
87xie, line 128: Should the combobox value be cleared after navigating to a new page? Currently, the implementation retains the value.

87xie:
I found a small difference, but I'm not sure if it's intended.

1. Typing something causes the 'No results found.' message to appear.
1. Clearing the input by pressing delete, the message still appears.

dimaMachina:
> 2. Clearing the input by pressing delete, the "No result Found." message still appears.

saw it too, but I haven't found a way to fix it 😅

87xie:
Adding a condition to render the empty result. Does it make sense?


https://github.com/user-attachments/assets/935da68f-3270-44c5-986b-14424410b1af

87xie:
It appears that headlessui ComboBox might not fully support IME (Input Method Editor) interactions. When selecting Pinyin characters using IME, also affects the ComboBox keyboard interactions.😭

https://github.com/user-attachments/assets/5c50e707-a2ed-49d7-84e3-f76c0c56e98d

dimaMachina:
> I adding a condition to render the empty result. Does it make sense?
> 
> ```ts
> value.length > 0 && renderComponent(themeConfig.search.emptyResult)
> ```
> 
>  _screencast.mov

oh yeah, it's exactly what's needed, I was struggled why `empty:_invisible` didn't work (because we still had a child no results 😂)
image

dimaMachina:
> It appears that headlessui ComboBox might not fully support IME (Input Method Editor). When selecting Pinyin characters using IME, also affects the ComboBox keyboard interactions.😭
> 
>  p.mov

maybe we need to add `onKeyDown` prop in Input ? because I removed him

It's a shame we don't have unit tests for chinese languages 😞

packages/nextra-theme-docs/src/components/search.tsx
dimaMachina: Added!

dimaMachina:
@87xie I did a nice refactoring here https://github.com/shuding/nextra/pull/3186/commits/6f3d35a1ac005ad184bcb194e42c5db6f29a8484

87xie:
>maybe we need to add onKeyDown prop in Input ? because I removed him

I referenced the Combobox [implementation](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/combobox/combobox.tsx#L1221-L1239), and it seems that currently, we can't prevent the combobox interactions with the up and down arrow keys. I will create a discussion to seek some suggestions.

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -85,121 +79,35 @@ export function Search({
...
+      ) : (
+        'CTRL K'
+      )}
+    </kbd>
   )
 
+  const handleKeyDown = useCallback(<T,>(event: KeyboardEvent<T>) => {
+    // skip the character selection process in the IME for CJK language users.
+    if (event.nativeEvent.isComposing) return
+  }, [])
```
dimaMachina, line 106: @87xie I added this, lmk maybe it fixes your issue

also I pushed another commit https://github.com/shuding/nextra/pull/3186/commits/3f93edfb14475540cd12381c25321b7cf44bdc97 where I fix cmdk+k shortcut (accidentally removed) and I removed `ESC` icon in input because it's doesn't work since headlessui adds `inert` attribute to esc icon once we type something

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -241,85 +148,95 @@ export function Search({
...
+      onChange={handleSelect}
+      className={cn('nextra-search _relative md:_w-64', className)}
+    >
+      <ComboboxInput
+        ref={inputRef}
+        as={Input}
+        autoComplete="off"
+        value={value}
         onCompositionStart={handleCompositionStart}
         onCompositionEnd={handleCompositionEnd}
```
87xie, line 184: need to remove `onCompositionStart` and `onCompositionEnd` because the previous usage was for uncontrolled input. (I've tested it and the issue in #2833 does not reoccur.)

dimaMachina: removed, anything else to improve on this PR? or ready to merge?

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -213,113 +121,105 @@ export function Search({
...
+        const { value } = event.currentTarget
+        onChange(value)
         compositionStateRef.current.emitted = true
         return
       }
       compositionStateRef.current.emitted = false
     },
-    [onChangeProp]
+    [onChange]
   )
```
87xie, line 134: is also related to composition but is no longer needed at the moment.

remove `compositionStateRef` and change to 

```ts
const handleChange = useCallback(
  (event: SyntheticEvent<HTMLInputElement>) => {
      const { value } = event.currentTarget
      onChange(value)
  },
  [onChange]
)
```

packages/nextra-theme-docs/src/components/search.tsx
87xie: It cannot prevent headlessui's `onKeyDown` callback from continuing to execute, so we can remove it for now.

packages/nextra-theme-docs/src/components/search.tsx
dimaMachina: removed here https://github.com/shuding/nextra/pull/3186/commits/5809fe9b355682b506cd2cf93ff84a270f00ee41 and did cleanup of a select component to avoid using render props

also, forget to mention that on mobile search results now take full height, i think is better

https://github.com/user-attachments/assets/4bdc3914-c3de-4c29-86a2-8ae49d7cb199

packages/nextra-theme-docs/src/components/search.tsx
```diff
@@ -85,241 +79,136 @@ export function Search({
...
+    >
+      <ComboboxInput
+        ref={inputRef}
+        as={Input}
+        autoComplete="off"
+        value={value}
         onChange={handleChange}
         onFocus={handleFocus}
         onBlur={handleFocus}
+        onKeyDown={handleKeyDown}
```
87xie, line 160: Currently not needed `onKeyDown`

dimaMachina: did merge and removed onKeyDown

87xie:
No other issues at the moment. Thank you!

PR #3192, opened 2024-09-02
Add introduction text and usage explanation to cards.mdx

SBobbe:
Added introductory sentences.

### DIFF
--- a/docs/pages/docs/guide/built-ins/cards.mdx
+++ b/docs/pages/docs/guide/built-ins/cards.mdx
@@ -20,7 +20,7 @@ A built-in component to embed a link in a card with an icon.
 ## Usage
 
 First, import the `Card` and `Cards` components. 
-Then, import the icons that you want to use from `../../icons`.
+Then, import the icons that you want to use.
 To create a set of cards, follow the example below.
 
 {/* prettier-ignore */}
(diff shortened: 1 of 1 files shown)


SBobbe:
Hello! I noticed that in the icons directory, there are many more cards icons listed than in this example. Are all of these icons included by default with all Nextra sites? If so, I would be happy to document them here.

dimaMachina:
> Are all of these icons included by default with all Nextra sites?

Do you mean https://github.com/shuding/nextra/tree/main/docs/components/icons in this folder? If yes, nope these icons can only be accessible for the Nextra documentation website (www.nextra.site)

Shared icons for websites built with `nextra` can be accessible from `nextra/icons` path https://github.com/shuding/nextra/tree/main/packages/nextra/src/icons

docs/pages/docs/guide/built-ins/cards.mdx
```diff
@@ -17,6 +19,10 @@ import { Card, Cards } from 'nextra/components'
 
 ## Usage
 
+First, import the `Card` and `Cards` components. 
+Then, import the icons that you want to use from `../../icons`.
```
dimaMachina, line 23:
```suggestion
Then, import the icons that you want to use.
```

I think we don't need to describe `from '../../icons'` since it's just an example maybe we need to change example code to

```mdx
import { CardsIcon, OneIcon, WarningIcon } from '../path-with-your-icons'
```

@SBobbe what do you think?

dimaMachina (approved):

PR #3212, opened 2024-09-08
[v3] fix first item in `FileTree` not within permitted parent element

87xie:
## Description

Currently, `<FileTree>` wraps its children with a `<div>`, while `<FileTree.Folder>` is represented as a list item `<li>`. This results in the list item not being placed within permitted parent elements. This change ensures that `<FileTree.Folder>` is appropriately nested within an `<ul>` element.

**after**

image

image

Additionally, does the border color of `<FileTree>` in dark mode need better contrast?

### DIFF
--- a/packages/nextra/src/client/components/file-tree.tsx
+++ b/packages/nextra/src/client/components/file-tree.tsx
@@ -31,9 +31,9 @@ function Tree({ children }: { children: ReactNode }): ReactElement {
         '_not-prose' // for nextra-theme-blog
       )}
     >
-      <div className="_inline-block _rounded-lg _border _px-4 _py-2 dark:_border-neutral-800">
+      <ul className="_inline-block _rounded-lg _border _px-4 _py-2 dark:_border-neutral-800">
         {children}
-      </div>
+      </ul>
     </div>
   )
 }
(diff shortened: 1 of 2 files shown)


packages/nextra/src/client/components/file-tree.tsx
```diff
@@ -31,9 +31,9 @@ function Tree({ children }: { children: ReactNode }): ReactElement {
         '_not-prose' // for nextra-theme-blog
       )}
     >
-      <div className="_inline-block _rounded-lg _border _px-4 _py-2 dark:_border-neutral-800">
+      <ul className="_inline-block _rounded-lg _border _px-4 _py-2 dark:_border-neutral-800">
```
dimaMachina, line 34:
```suggestion
      <ul className="_inline-block _rounded-lg _border _px-4 _py-2 _border-neutral-200/70 contrast-more:_border-neutral-400 dark:_border-primary-100/10 contrast-more:dark:_border-neutral-400">
```

dimaMachina: I changed border color to be same as `<hr />` color

dimaMachina (approved):

PR #3214, opened 2024-09-09
docs: fix meta tag theme color

SID12g:
When connecting nextra docs, theme color is maintained in #fff even though it is in dark mode, so I modified it so that the theme color can be determined according to the presence or absence of the user's dark mode.

<div>

<p>Existing docs appearance</p>
</div>
<div>

<p>After changing meta information</p>
</div>


dimaMachina (approved):
Hi, thanks, sorry for the time merging this, CSS variables don't work in `<meta name="theme-color"`, I pushed my commit which fix it for Nextra 4

PR #2569, opened 2023-12-05
[v4] migrate search from Flexsearch to Pagefind

dimaMachina:
closes https://github.com/shuding/nextra/issues/2524

todos:
- [x] include subresults https://pagefind.app/docs/sub-results/#retrieving-sub-results-using-the-javascript-api
	- [x] <s>fix subresult active state</s>
	- [x] fix `subheading-anchor` styles https://github.com/shuding/nextra/blob/v3/packages/nextra-theme-docs/css/typesetting-article.css#L25-L27 and search other places by `subheading-anchor` keyword https://github.com/shuding/nextra/pull/3132
	- [x] <s>fix if there is only 1 subresult like currently</s>
- [x] <s>include indexing directly in webpack loader instead of `"postbuild": "pagefind --site .next/server/pages --output-path .next/static/chunks/pages/pagefind"`</s>
- [x] figure out how to setup for i18n websites
- [x] support `frontmatter.searchable` property
- [x] support `search.codeblocks`

PR #3220, opened 2024-09-10
[v3] add auto-import-theme-style option in nextra configuration

87xie:
## Description

This pull request includes part of the content from https://github.com/shuding/nextra/pull/3154 and aims to allow Nextra's predefined styles to support cascade layers while avoiding inadvertently overriding user-defined CSS layers.

Considering browser compatibility, I think providing a flexible way that allows the outer-land to decide would be more appropriate at this stage, while also ensuring that users who do not need this feature are not affected.

With this change, cascade layers support is implemented as follows:

installing [postcss-import](https://www.npmjs.com/package/postcss-import), and adding it to `postcss.config.js`

```ts
export default {
  plugins: {
    'postcss-import': {},
    // ...
  }
}
```

In `next.config.js`, set `autoImportThemeStyle: false`:

```ts
const withNextra = nextra({
  autoImportThemeStyle: false,
  theme: 'nextra-theme-docs',
  // ...
})
```

Import the nextra-theme CSS and specify layers:

```css
@import "nextra-theme-docs" layer(any-name-you-like);

/* user defined CSS... */
```

Import CSS file in `page/_app.tsx`

```tsx
import '../styles.css'
 
export default function MyApp({ Component, pageProps }) {
  return <Component {...pageProps} />
}
```

I have tested this implementation in `examples/swr-site`

image
---

Additionally, in my opinion, allowing users to import the official theme CSS file in `app/layout.tsx` in  v4 app router is preferable to having it imported automatically in the [entry file](https://github.com/shuding/nextra/blob/v4-v2/packages/nextra-theme-docs/src/index.ts#L1). This reduces the CSS side effect imports and gives users more control.

### DIFF
--- a/.changeset/red-rockets-perform.md
+++ b/.changeset/red-rockets-perform.md
@@ -1,5 +1,4 @@
 ---
-'nextra-theme-blog': patch
 'nextra-theme-docs': patch
 'nextra': patch
 ---
--- a/.changeset/red-rockets-perform.md
+++ b/.changeset/red-rockets-perform.md
@@ -1,0 +1,7 @@
+---
+'nextra-theme-blog': patch
+'nextra-theme-docs': patch
+'nextra': patch
+---
+
+Add the `autoImportThemeStyle` option to the Nextra configuration. This allows users to import the official Nextra theme CSS into a specific cascade layer.
--- a/examples/swr-site/styles.css
+++ b/examples/swr-site/styles.css
@@ -1,4 +1,4 @@
-@import 'nextra-theme-docs' layer(nextra-base);
+@import 'nextra-theme-docs/dist/style.css' layer(nextra-base);
 @tailwind utilities;
 
 .dark .invert-on-dark {
--- a/examples/swr-site/styles.css
+++ b/examples/swr-site/styles.css
@@ -1,4 +1,5 @@
 @import 'nextra-theme-docs/dist/style.css' layer(nextra-base);
+
 @tailwind utilities;
 
 .dark .invert-on-dark {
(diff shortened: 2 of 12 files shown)


examples/swr-site/styles.css
```diff
@@ -1,3 +1,4 @@
+@import 'nextra-theme-docs' layer(nextra-base);
```
dimaMachina, line 1:
```suggestion
@import 'nextra-theme-docs/style.css' layer(nextra-base);
```

let’s be more explicit, so `style` field in package.json can be removed too

also, can you give an example (in swr-site example) how cascade layers do css priority before/after?

packages/nextra/src/server/schemas.ts
```diff
@@ -106,7 +106,8 @@ export const nextraConfigSchema = z
       recmaPlugins: z.custom<ProcessorOptions['recmaPlugins']>(),
       format: z.enum(['detect', 'mdx', 'md']),
       rehypePrettyCodeOptions: z.custom<RehypePrettyCodeOptions>()
-    })
+    }),
+    autoImportThemeStyle: z.boolean()
```
dimaMachina, line 110: Maybe we should use Zod's `.default(true)`? And for other properties declared in `DEFAULT_CONFIG` object, so this object can be removed?

.changeset/red-rockets-perform.md
```diff
@@ -0,0 +1,7 @@
+---
+'nextra-theme-blog': patch
```
dimaMachina, line 2:
```suggestion
```

examples/swr-site/styles.css
87xie: `postcss-import` cannot recognize conditional exports in `package.json`, so I changed it to:

```css
@import 'nextra-theme-docs/dist/style.css';
```

87xie: Both `swr-site/styles.css` and `nextra-theme-docs/style.css` define styles for the body, with the same specificity of `0 0 1`.

- **Before using cascade layers**: The browser applies the styles of the latter selector based on the final compiled CSS order.
- **After using cascade layers**: Unlayered styles will be applied with higher priority, without considering the order of the styles.

**nextra-theme-docs**

```css
body, .nextra-nav-container-blur, .nextra-toc-footer, .nextra-sidebar-footer {
    background-color: rgb(var(--nextra-bg));
}
```

**swr-site**

```css
body {
    background: linear-gradient(to bottom, rgba(255, 255, 255, 0) 0%, rgb(var(--nextra-bg)) 300px), 
    fixed 0 0 / 20px 20px radial-gradient(#d1d1d1 1px, transparent 0), 
    fixed 10px 10px / 20px 20px radial-gradient(#d1d1d1 1px, transparent 0);
}
```

packages/nextra/src/server/schemas.ts
87xie: It seems that we need to remove .deepPartial() for the expected default values to work. This might lead to many unintended changes beyond this PR, and I'm worried about potentially breaking something 😰

dimaMachina: you are right, I did this refactor in https://github.com/shuding/nextra/pull/2569 PR

87xie: The `nextraConfigSchema` in #2569 looks way better!

examples/swr-site/styles.css
dimaMachina: amazing improvement! It would be nice to mention this in docs as well

dimaMachina (approved):

dimaMachina (approved):

Issue #2020, opened 2023-06-28
[BUG] TOC highlights the wrong section when current section is exiting the viewport


lakshya0246:
### Description
There's an issue in the Table of contents (TOC) on the right. If you click on a link in the toc, and scroll beyond the section, as you scroll, the TOC jumps to and highlights a link that is not in view, then comes back to the correctly highlighted link.

### Reproduction
The issue can be observed on this page:
https://nextra.site/docs/docs-theme/theme-configuration

1. Click on the link `MDX Components` in the TOC under `Content`
2. The page scrolls to `MDX Components`. ✅
3. But `Global` is still highlighted in the TOC. ❌
5. Now scroll the page slowly up until `MDX Components` is about to exit the viewport from bottom. TOC jumps to `Global`.  ❌

Tested on: 
- Chrome 114.0.5735.198
- MacBook Pro 16
- Viewport sizes : 976 x 1920 and 1000 x 1666

B2o5T:
please attach video, can't reproduce

lakshya0246:
Sure, here's the video

B2o5T:
Thank you! Will debug why this occurs

dimaMachina:
should be fixed by https://github.com/shuding/nextra/pull/3249

PR #2224, opened 2023-08-28
fixes

dimaMachina:
- do not scroll to top while calling cmd+k in search input

- fix wrong toc highlight / simplify active heading implementation

fixes https://github.com/shuding/nextra/issues/2219 https://github.com/shuding/nextra/issues/2020

PR #3249, opened 2024-09-15
[v3] consider navbar height for the anchor intersection observer

87xie:
## Description

This change aims to resolve the issue shown below the image:
1. When I click the `href=“#two”` link in the table of contents (TOC),
1. The URL updates, and the page scrolls to the corresponding anchor,
1. However, the active style in the TOC stays on the link before `href=“#two”`.

image

## How

I think the intersection area should be restricted to below the navbar height, aligning it more closely with the user's visible area.

The current implementation in this PR uses `document.querySelector('.nextra-content')`, which may not be the best approach. If we could move the `<ActiveAnchorProvider>` into the `<NextraWrapper>` in the MDX components and obtain the corresponding DOM reference for `.nextra-content` through a ref, it might provide better consistency and more closely align with its actual scope of effect.

Feel free to close this if I missed something obvious.

**After this change**

https://github.com/user-attachments/assets/5aaa6e5e-7af6-4c4a-a241-8f227c36022e

### DIFF
--- a/packages/nextra-theme-docs/src/contexts/active-anchor.tsx
+++ b/packages/nextra-theme-docs/src/contexts/active-anchor.tsx
@@ -11,12 +11,12 @@ type ActiveAnchor = Record<
   }
 >
 
-const ActiveAnchorContext = createContext<ActiveAnchor>({})
+const ActiveAnchorContext = createContext<ActiveAnchor>(null!)
 ActiveAnchorContext.displayName = 'ActiveAnchor'
 
 const SetActiveAnchorContext = createContext<
   Dispatch<SetStateAction<ActiveAnchor>>
->(v => v)
+>(null!)
 SetActiveAnchorContext.displayName = 'SetActiveAnchor'
 
 const IntersectionObserverContext = createContext<IntersectionObserver | null>(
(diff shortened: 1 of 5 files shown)


dimaMachina (commented):
nice, seems it fixes https://github.com/shuding/nextra/issues/2020 and https://github.com/shuding/nextra/pull/2224

packages/nextra-theme-docs/src/contexts/active-anchor.tsx
```diff
@@ -42,8 +41,12 @@ export const ActiveAnchorProvider = ({
...
 }): ReactElement => {
   const [activeAnchor, setActiveAnchor] = useState<ActiveAnchor>({})
-  const observerRef = useRef<IntersectionObserver | null>(null)
-  if (IS_BROWSER && !observerRef.current) {
+  const observerRef = useRef<IntersectionObserver>(null!)
+
+  useEffect(() => {
+    const navbarHeight = getComputedStyle(document.body).getPropertyValue(
+      '--nextra-navbar-height'
+    )
```
dimaMachina, line 49: let's stay with navbar height value, because `nextraContentEl.offsetTop` can be 104px (navbar + banner) for this situation

image

87xie: Thank you for your suggestion. I hadn't considered this situation. The reason I used `offsetTop` is that the navbar could be configured by the user through `themeConfig`.

dimaMachina: oh yeah, in situation when user override navbar it could be not consistent with `--nextra-navbar-height` value

PR #3251, opened 2024-09-15
[v3] remove aria-hidden attr when scroll-to-top button is visible

87xie:
## Description

A small fix to correctly use the `aria-hidden` attribute, as referenced by [this guideline](https://dequeuniversity.com/rules/axe/4.9/aria-hidden-focus): elements with `aria-hidden: true` must not be focusable.

### DIFF
--- a/packages/nextra-theme-docs/src/components/back-to-top.tsx
+++ b/packages/nextra-theme-docs/src/components/back-to-top.tsx
@@ -26,6 +26,7 @@ export function BackToTop({
 }): ReactElement {
   return (
     <Button
+      // elements with `aria-hidden: true` must not be focusable or contain focusable elements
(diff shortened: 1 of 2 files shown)


packages/nextra-theme-docs/src/components/back-to-top.tsx
```diff
@@ -26,7 +26,7 @@ export function BackToTop({
 }): ReactElement {
   return (
     <Button
-      aria-hidden="true"
+      aria-hidden={hidden ? 'true' : undefined}
```
dimaMachina, line 29:
```suggestion
      // elements with `aria-hidden: true` must not be focusable or contain focusable elements
      aria-hidden={hidden ? 'true' : undefined}
```

dimaMachina (approved):

PR #2895, opened 2024-05-21
support custom label for back-to-top link

xijinping0:
This PR resolves https://github.com/shuding/nextra/issues/2308 by adding a `backToTopLabel` field to the `toc` config.

I have confirmed in `examples/docs` that this change is working locally. The label is set to `Go back to top please` in the screenshot:

Screen Shot 2024-05-21 at 00 16 22

Thank you in advance for your review.

Issue #2308, opened 2023-09-19
i18n support for `Scroll to top`


zwaittt:
expecting i18n config for this line

https://github.com/shuding/nextra/blob/093336858a23245fa47f8fe07f33f2d425a80416/packages/nextra-theme-docs/src/components/back-to-top.tsx#L34

xijinping0:
I created #2895 to address this issue.

dimaMachina:
will be available in next v3 alpha https://github.com/shuding/nextra/pull/3258

PR #3258, opened 2024-09-17
[v3] update the back-to-top option to support i18n

87xie:
## Description

[#2308](https://github.com/shuding/nextra/issues/2308) seems to be planned for v3, but it hasn't been addressed yet (currently, there is  a PR #2895  for v2), so I would like to try to help.

I referenced the implementation of the `feedback` option and decided not to expose the `backToTop.component` option, as the `className` and `hidden` props might not be necessary for users.

Here's how the implementation allows users to control the text through `backToTop.text`, for example:

```ts
const themeConfig = {
  backToTop: {
    text: (locale: string) => {
      switch (locale) {
        case 'ru': return 'Перейти наверх'
        case 'es': return 'Desplazarse hacia arriba'
        default: return 'Scroll to top'
      }
    }
  }
}
```

Feel free to close this if it is not suitable.

## Screenshots

`locale: es`

image

`locale: en`

image

### DIFF
--- a/packages/nextra-theme-docs/src/constants.tsx
+++ b/packages/nextra-theme-docs/src/constants.tsx
@@ -25,28 +25,13 @@ export const DEFAULT_LOCALE = 'en-US'
 export type DocsThemeConfig = z.infer<typeof themeSchema>
 export type PartialDocsThemeConfig = z.infer<typeof publicThemeSchema>
 
-const LOADING_LOCALES: Record<string, string> = {
-  'en-US': 'Loading',
-  fr: 'Сhargement',
-  ru: 'Загрузка',
-  'zh-CN': '正在加载'
-}
-
-const PLACEHOLDER_LOCALES: Record<string, string> = {
-  'en-US': 'Search documentation',
-  fr: 'Rechercher documents',
-  ru: 'Поиск документации',
-  'zh-CN': '搜索文档'
-}
-
 export const DEFAULT_THEME: DocsThemeConfig = {
   backToTop: {
     content: function BackToTopContent() {
       const themeConfig = useThemeConfig()
-      const { locale = '' } = useRouter()
       return (
         <>
-          {renderString(themeConfig.backToTop.text, locale)}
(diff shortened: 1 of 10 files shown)


packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint sort-keys: error */
-import { useRouter } from 'next/router'
```
87xie, line 2: should use `useRouter` from `nextra/hooks`, am I correct?

packages/nextra-theme-docs/src/constants.tsx
```diff
@@ -167,14 +191,17 @@ export const DEFAULT_THEME: DocsThemeConfig = {
...
+      return `${text}…`
     },
     placeholder: function usePlaceholder() {
       const { locale, defaultLocale = DEFAULT_LOCALE } = useRouter()
       const text =
         (locale && PLACEHOLDER_LOCALES[locale]) ||
-        PLACEHOLDER_LOCALES[defaultLocale]
+        PLACEHOLDER_LOCALES[defaultLocale] ||
+        PLACEHOLDER_LOCALES[DEFAULT_LOCALE]
       return `${text}…`
```
87xie, line 205: Avoid displaying `undefined` as text. For example, if  `defaultLocale: ja-JP`, it might display as `undefined...` by default.

packages/nextra-theme-docs/src/constants.tsx
dimaMachina: I think yes, good catch, btw I removed this hook in v4

dimaMachina (approved):
thanks, I did a bit of refactor in your PR to keep it simpler

packages/nextra-theme-docs/src/constants.tsx
dimaMachina: I think it's better to completely remove default values for other languages different from `en` for i18n properties, to keep it simple and not increase bundle size

PR #3260, opened 2024-09-18
[v3] make react 18 as minimal requirement

87xie:
## Description

In v3, React 18 needs to be considered the minimal requirement because of the following:

- The steps component currently uses `useId`
- Headless UI v2

Also, remove `@popperjs/core` (currently unused)


dimaMachina (approved):
Great!

PR #3248, opened 2024-09-14
[v3] fix global style conflicts for `<a>`/`<button>`/`<summary>`/`<input>`/`[tabindex]:not([tabindex='-1']`

dimaMachina:
superseded https://github.com/shuding/nextra/pull/3209 https://github.com/shuding/nextra/pull/3154
closes #2613, #2764, and #3099

PR #3286, opened 2024-09-24
[v3] enhance focus ring style consistency

87xie:
## Description

Adjustments made in this PR:

- Apply focus ring style to `<SkipNav>`.
- Apply a `_rounded` style to `a` and `summary` tags when they are focus-visible, making them more consistent with the navbar's menu button. A similar style was present in the previous version [here](https://github.com/shuding/nextra/pull/3248/files#diff-9b96b3afc3b3cf2aa0292d40310cf7848a9f2fe45eeb026eabb34c7ec3a0b82aL41-L44).
- Adjusted the `<Tab>` class name to `nextra-focus focus:_ring` because the `focus` parameter in the `className` callback wasn't updating when `<Tab>` was clicked. This ensures compatibility with both keyboard navigation and click activation.

Feel free to close this If I missed something obvious.

## Screenshots

**Before**

image

image

image

https://github.com/user-attachments/assets/3ee62056-0fc1-4118-9b6a-bf24838c641e

**After**

https://github.com/user-attachments/assets/c91e9187-d483-4471-94c9-702e53e20112

### DIFF
--- a/packages/nextra-theme-docs/css/styles.css
+++ b/packages/nextra-theme-docs/css/styles.css
@@ -30,6 +30,9 @@ body {
 .nextra-focusable,
 .nextra-focus:focus-visible {
   @apply _ring-2 _ring-primary-200 _ring-offset-1 _ring-offset-primary-300 dark:_ring-primary-800 dark:_ring-offset-primary-700;
+  &:is(a, summary) {
+    @apply _rounded;
+  }
 }
 
 .nextra-focusable,
--- a/packages/nextra/src/client/components/tabs/tabs.tsx
+++ b/packages/nextra/src/client/components/tabs/tabs.tsx
@@ -101,9 +101,9 @@ export function Tabs({
           <HeadlessTab
             key={index}
             disabled={isTabObjectItem(item) && item.disabled}
-            className={({ selected, disabled, hover, focus }) =>
+            className={({ selected, disabled, hover }) =>
               cn(
-                focus && 'nextra-focusable _ring-inset',
+                'nextra-focus focus:_ring-inset',
                 '_whitespace-nowrap',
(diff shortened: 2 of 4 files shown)


packages/nextra/src/client/components/tabs/tabs.tsx
```diff
@@ -101,9 +101,9 @@ export function Tabs({
           <HeadlessTab
             key={index}
             disabled={isTabObjectItem(item) && item.disabled}
-            className={({ selected, disabled, hover, focus }) =>
+            className={({ selected, disabled, hover }) =>
               cn(
-                focus && 'nextra-focusable _ring-inset',
+                'nextra-focus focus:_ring-inset',
```
dimaMachina, line 106:
```suggestion
focus && 'nextra-focusable _ring-inset',
selected && '_outline-none',
```

the fix is simpler we should hide outline for `selected` state

packages/nextra-theme-docs/css/styles.css
```diff
@@ -30,6 +30,9 @@ body {
 .nextra-focusable,
 .nextra-focus:focus-visible {
   @apply _ring-2 _ring-primary-200 _ring-offset-1 _ring-offset-primary-300 dark:_ring-primary-800 dark:_ring-offset-primary-700;
+  &:is(a, summary) {
+    @apply _rounded;
+  }
```
dimaMachina, line 35:
```suggestion
```

I would prefer to not apply rounded class, because element already can have radius (previous approach was a mistake)

87xie: Got it! Thanks for pointing that out.

dimaMachina (approved):

PR #3354, opened 2024-10-03
[v4] enable page reload of catch-all routes on content change

dimaMachina:
cc @87xie hot reloading now works for catch-all routes

https://github.com/user-attachments/assets/552381c8-b9d1-47a9-9016-8799755ad19e


.changeset/proud-ears-think.md
```diff
@@ -0,0 +1,7 @@
+---
+'nextra-theme-blog': patch
+'nextra-theme-docs': patch
+'nextra': patch
+---
+
+add refresh page when markdown content changed
```
dimaMachina, line 7:
```suggestion
enable page reload of catch-all routes `app/[[...slug]].jsx` on content change
```

PR #3362, opened 2024-10-04
ensure jump link behavior is consistent in markdown

87xie:
### Why:

Closes: N/A

In theme-docs, when using markdown to create a jump link like `[My content](#my-content)`, clicking the link to jump to specific content causes that content to be covered by the navbar, It can be reproduced by clicking on the `GFM syntax table:` [here](https://nextra.site/docs/guide/advanced/table#unexpected-result), and this behavior is inconsistent with clicking on TOC links and sub-heading anchors.

The jump link created via markdown syntax is converted into a next-link. When clicking it, the `activeElement` remains on the link, which causes the following CSS not to be applied:

```css
html:not(:has(*:focus)) {
  @apply _scroll-pt-[--nextra-navbar-height];
}
```

### What's being changed (if available, include any code snippets, screenshots, or gifs):

To resolve this, I added a check on the MDX anchor component to see if the href starts with `#`. If true, it uses `<a>`

https://github.com/user-attachments/assets/24e8abc3-ceee-49c2-bfee-259cfe5d2393

### Check off the following:

      opening the PR).

  - For content changes, you will also see an automatically generated comment
    with links directly to pages you've modified. The comment won't appear if
    your PR only edits files in the `data` directory.

- [ ] For content changes, I have completed the
      [self-review checklist](https://docs.github.com/en/contributing/collaborating-on-github-docs/self-review-checklist).

### DIFF
--- a/packages/nextra-theme-blog/src/mdx-theme.tsx
+++ b/packages/nextra-theme-blog/src/mdx-theme.tsx
@@ -56,7 +56,6 @@ function HeadingLink({
 }
 
 const EXTERNAL_HREF_REGEX = /https?:\/\//
-const ID_HREF_REGEX = /^#/
 
 const A = ({ children, href = '', ...props }: ComponentProps<'a'>) => {
   if (EXTERNAL_HREF_REGEX.test(href)) {
@@ -66,7 +65,7 @@ const A = ({ children, href = '', ...props }: ComponentProps<'a'>) => {
       </a>
     )
   }
-  const ComponentToUse = ID_HREF_REGEX.test(href) ? 'a' : Link
+  const ComponentToUse = href.startsWith('#') ? 'a' : Link
   return (
     <ComponentToUse href={href} {...props}>
       {children}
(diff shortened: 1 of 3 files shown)


dimaMachina (commented):
Wow 😍 this should be a very nice fix, I will test it before merging, once I will have free time (and your another docs PR too ❤️)

packages/nextra-theme-blog/src/mdx-theme.tsx
```diff
@@ -65,10 +66,11 @@ const A = ({ children, href = '', ...props }: ComponentProps<'a'>) => {
       </a>
     )
   }
+  const ComponentToUse = ID_HREF_REGEX.test(href) ? 'a' : Link
```
dimaMachina, line 69: Since regex contains only 1 character and start with `^` character, we can simplify it to, wdyt?
```suggestion
  const ComponentToUse = href.startsWith('#') ? 'a' : Link
```

87xie: Keeping it simple is a great idea! I referred to the approach used in `EXTERNAL_HREF_REGEX` to avoid receiving props other than strings, which could lead to type errors. Perhaps I was overthinking it.

dimaMachina: Yeah, I got it, you stayed with consistentsy with the current code 😊👍

dimaMachina (approved):
Nice catch and fix 🚀

PR #3376, opened 2024-10-05
add a scrollbar for nav links in the header when they exceed available free space

dimaMachina:
fixes https://github.com/shuding/nextra/issues/2785

inspired by [mdx website](https://mdxjs.com)

### DIFF
--- a/packages/nextra-theme-docs/src/components/navbar.tsx
+++ b/packages/nextra-theme-docs/src/components/navbar.tsx
@@ -127,47 +127,51 @@ export function Navbar({ items }: NavBarProps): ReactElement {
             {renderComponent(themeConfig.logo)}
           </div>
         )}
-        {items.map(pageOrMenu => {
-          if (pageOrMenu.display === 'hidden') return null
+        <div className="_flex _gap-4 _overflow-x-auto nextra-scrollbar _py-1.5">
+          {items.map(pageOrMenu => {
+            if (pageOrMenu.display === 'hidden') return null
 
-          if (pageOrMenu.type === 'menu') {
-            const menu = pageOrMenu as MenuItem
+            if (pageOrMenu.type === 'menu') {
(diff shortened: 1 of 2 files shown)


dimaMachina:
cc @87xie lmk wdyt 🙏

87xie:
I think each approach is quite subjective, and it might be more appropriate to leave these trade-offs to the outer-land.

If Nextra could provide lower-level components that allow users to compose their components (similar to the approach in v4):

```jsx
<Navbar>
  <MyLogo />
  <MyLinks />
  <Searchbar />
  <ProjectLink />
</Navbar>
```

This would give them the flexibility to implement the solutions that might be more suitable for their use case, for example, prioritizing the hiding of less important elements, using overflow: scroll, or dynamically calculating with JavaScript.

My use case with theme-docs doesn't involve a complex header, so I’m unable to provide useful suggestions 😅

marcklingen:
thanks for adding this (just noticed it and was looking for the pr)!

we use theme-docs for our main website and docs and would love to customize the main nav more easily as described by @87xie

packages/nextra-theme-docs/src/components/navbar.tsx
```diff
@@ -127,47 +127,51 @@ export function Navbar({ items }: NavBarProps): ReactElement {
             {renderComponent(themeConfig.logo)}
           </div>
         )}
-        {items.map(pageOrMenu => {
-          if (pageOrMenu.display === 'hidden') return null
+        <div className="_flex _gap-4 _overflow-x-auto nextra-scrollbar _py-1.5">
```
87xie, line 130: I noticed that the focus ring styles of the first and last links are being cut off.

image

dimaMachina: yeah, noticed it too, little regression

Issue #3476, opened 2024-10-17
CSS Selector Targeting Inline Code Blocks Too Aggressively


arno-fukuda:
In v3, there is a new css selector that targets all external Code elements.

```
:not(pre)>code:not([class*=twoslash-]) {
    overflow-wrap: break-word;
    border-width: 1px;
    border-color: rgb(0 0 0/var(--tw-border-opacity));
    --tw-border-opacity: .04;
    background-color: rgb(0 0 0/var(--tw-bg-opacity));
    --tw-bg-opacity: .03;
    border-radius: .375rem;
    padding: .125rem .25em;
    font-size: .9em;
}
```

Imo, it's too broad and can spill/overwrite custom components that use `<code>` with their own styles

dimaMachina:
Lgtm feel free to send PR

PR #3480, opened 2024-10-17
[ISSUE-3476] style: narrow css specificity for code elements

arno-fukuda:
## Why:
All code elements written as markdown in mdx files via backticks will automatically have the `nextra-code` class.

Any other code elements that could come from custom components, might not want the default nextra style applied to it.

In fact, if you have tailwind applied to your custom `<code>` elements, it will be overwritten by the global `code:not([class*='twoslash-'])` specifier since it has a higher specificity.

If someone prefers the global css for custom `<code>` elements, they can add the class `nextra-code` to get it applied still.

Closes:
https://github.com/shuding/nextra/issues/3476

css specifier

## Check off the following:

      opening the PR).

### DIFF
--- a/.changeset/fifty-birds-invite.md
+++ b/.changeset/fifty-birds-invite.md
@@ -1,0 +1,5 @@
+---
+"nextra": patch
+---
+
+update css selectors for `code`, use `code.nextra-code`
--- a/packages/nextra/styles/code-block.css
+++ b/packages/nextra/styles/code-block.css
@@ -3,7 +3,7 @@
   @apply dark:_bg-[--shiki-dark-bg] dark:_text-[--shiki-dark];
 }
 
-code:not([class*='twoslash-']) {
+code.nextra-code:not([class*='twoslash-']) {
   box-decoration-break: slice;
   font-feature-settings:
     'rlig' 1,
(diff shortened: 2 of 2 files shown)


dimaMachina (changes requested):
I think we should do the same for line
https://github.com/arno-fukuda/nextra/blob/e6e5bb3482f7736d6410dc7f6b770e8e109e269c/packages/nextra/styles/code-block.css#L19

wdyt?

> [!note]
> we need add new class for `<pre>` since it doesn't have for now 😅

arno-fukuda:
> I think we should do the same for line [arno-fukuda/nextra@`e6e5bb3`/packages/nextra/styles/code-block.css#L19](https://github.com/arno-fukuda/nextra/blob/e6e5bb3482f7736d6410dc7f6b770e8e109e269c/packages/nextra/styles/code-block.css#L19)
> 
> wdyt?
> 
> Note
> 
> we need add new class for `\<pre\>` since it doesn't have for now 😅 ignore it, we select `code` anyway

Good catch. 
It's safer indeed if someone wanted to to a custom <pre><code>.

dimaMachina (approved):

PR #3499, opened 2024-10-19
fix: banner and footer codeblock example

arno-fukuda:
## Why:
* replace text with content property in code-block example

## Check off the following:

      opening the PR).


dimaMachina (approved):
Thank you!

PR #3518, opened 2024-10-20
[v3] add tests for should respect order for `type: "separator"`, `type: "menu"` and item with `href`

dimaMachina:


.changeset/few-cups-smoke.md
```diff
@@ -0,0 +1,7 @@
+---
+'nextra-theme-docs': patch
+'nextra': patch
+---
+
+- fix empty dropdown menu when type: "menu" contains items with local pages
```
dimaMachina, line 6:
```suggestion
```

PR #3535, opened 2024-10-25
[v4] parse and transform `_meta` files with zod

dimaMachina:


.changeset/lovely-beds-argue.md
```diff
@@ -0,0 +1,6 @@
+---
+'nextra-theme-docs': patch
+'nextra': patch
+---
+
+check `_meta` files with zod
```
dimaMachina, line 6:
```suggestion
- check `_meta` files with zod
- remove `_meta` `newWindow` field
```

PR #3537, opened 2024-10-25
[v4] add required `Layout.navbar`, `Layout.footer` props and optional `Layout.banner` prop, you don't need to pass them as `children` now

dimaMachina:


docs/app/layout.tsx
```diff
@@ -75,36 +107,19 @@ const RootLayout: FC<{
...
+          navbar={navbar}
           pageMap={await getPageMap()}
           docsRepositoryBase="https://github.com/shuding/nextra/tree/main/docs"
           editLink="Edit this page on GitHub"
           sidebar={{ defaultMenuCollapseLevel: 1 }}
+          footer={footer}
+          nextThemes={{
+            defaultTheme: 'dark',
+            disableTransitionOnChange: true
+          }}
```
dimaMachina, line 120:
```suggestion
```

Issue #3567, opened 2024-10-27
_Meta file not finding files when nesting catch-all route in v4


Firgrep:
**Describe the bug** A clear and concise description of what the bug is.

Nesting the catch-all route one level down. /nested-path/[...mdxPath]] seems it's not obeying the _meta.ts file properly anymore, type: "page", is not being respected. 

Additionally, any nested paths inside of `content` are "not found". Without the nested path everythig seems to work fine.

![Screenshot from 2024-10-27 21-31-12](https://github.com/user-attachments/assets/7180f786-59bf-45f7-80f9-05712ecf23aa)

![image](https://github.com/user-attachments/assets/d6afa161-5a9a-4c27-b4ad-03daa476f1c8)

**To Reproduce** Steps to reproduce the behavior:

1. Go to 'a nextra project that uses v4 catch-all route'
2. Nest the catch-all route one level down
3. Run the app

**Expected behavior** A clear and concise description of what you expected to
happen.

Expected that it would be able to maintain the content folder integrity and its meta config.

**Screenshots** If applicable, add screenshots to help explain your problem.

**Desktop (please complete the following information):**

Ubuntu 24.04.1 LTS
Chrome
Chrome 130 on Linux

sgmustafi:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

dimaMachina:
Made some progress:

<img width="253" alt="image" src="https://github.com/user-attachments/assets/f32636b4-32e6-4bd0-a289-2509ed0d35f4">

<img width="696" alt="image" src="https://github.com/user-attachments/assets/57177aca-0d16-44df-976b-5a65fe845763">

Firgrep:
Wuuhuu! Nice! 🥳👌

PR #3593, opened 2024-10-29
[v4] add `contentDirBasePath` option to configure `content` directory under a catch-all subdirectory

dimaMachina:
fixes https://github.com/shuding/nextra/issues/3567 https://github.com/shuding/nextra/discussions/3582


Firgrep:
Question: will the content folder need to mimic the nested path from /app? E.g.: 
/App/mine/docs/[[...modxPaths]]/page.tsx
/Content/mine/docs/my-docs.mdx

Or can the content be anything?
/Content/my-docs.mdx

How will all of this relate to the items in the navigation if we use the basepath?

Thanks!

dimaMachina:
> Question: will the content folder need to mimic the nested path from /app? E.g.:

nope, you don't need to mimic, you can see this docs example https://github.com/shuding/nextra/tree/v4-50/examples/docs

Also, I am thinking of removing `useContentDir` option since we can know whether `content` directory is used based on `catchAllBasePath` value, wdyt?

Firgrep:
Alright, cool!

Yes, that makes sense, the user will have to set some value in the config anyhow to activate catch-all, so the ’useContentDir’ seems redundant.

.changeset/olive-seas-clap.md
```diff
@@ -0,0 +1,7 @@
+---
+'nextra-theme-blog': patch
+'nextra-theme-docs': patch
+'nextra': patch
```
dimaMachina, line 4:
```suggestion
'nextra-theme-blog': feat
'nextra-theme-docs': feat
'nextra': feat
```

packages/nextra/src/server/loader.ts
```diff
@@ -112,16 +112,19 @@ export async function loader(
...
-      dir: useContentDir ? path.join('content', locale) : APP_DIR,
-      isAppDir: !useContentDir
+      dir: contentDirBasePath ? path.join('content', locale) : APP_DIR,
+      isAppDir: !contentDirBasePath
     })
-
-    const { pageMap, mdxPages } = generatePageMapFromFilepaths(relativePaths)
+    const { pageMap, mdxPages } = generatePageMapFromFilepaths(
+      relativePaths,
+      // replace forward slash
```
dimaMachina, line 120:
```suggestion
      // remove forward slash
```

dimaMachina:
@Firgrep I found one issue for `docs` folder:

So now, this `_meta` is ignored from adding to `pageMap`

image

image

- we cannot rename `docs` folder here
- we cannot also use `display: 'children'` (to hide the folder and show only children).

This will be not fixed in this PR, and more likely be fixed after 4 RC.0.

When you set `contentDirBasePath: '/docs'` - pages `page.{tsx,mdx,jsx}` and `_meta` files  in `app` dir  are not added to `pageMap`, to fix it I have an idea: when `contentDirBasePath` is set - still add all `page.{tsx,mdx,jsx}` and `_meta` files from `app` dir to `pageMap`

dimaMachina:
@Firgrep ok this is fixed 😁 now we have we have 2 modes working together `content` and `page.{jsx,tsx,mdx}` routes added both to `pageMap`  as well `_meta` from `content` and `_meta` from `app`

Firgrep:
@dimaMachina Fixed as in you fixed what you said couldn't be fixed in this PR in the previous comment? I'm just reading it now. Ah, so there can co-exist a `_meta` file inside `/app` alongside `/content` ?

dimaMachina:
fix coming in another PR today

Firgrep:
This man is on 🔥 Can't wait for the next pre-release that includes these changes to implement it!

dimaMachina:
See the video demonstrating the working behaviour of both modes together and they play well, it's for a whiiiile I had in mind at some point to make them together 😜


https://github.com/user-attachments/assets/038118dc-22ac-42d3-badc-65c87a0cb40e

Firgrep:
@dimaMachina That's really awesome! Nextra is going to be more powerful than ever! You could even put that video up on the docs, or at least in your blog post.

dimaMachina:
looks nice idea!

PR #3616, opened 2024-10-31
[v4] do not treat `content/page.{mdx,md}` as `index` page

dimaMachina:
- do not treat `content/page.{mdx,md}` as index page
- skip visiting directories which starts with underscore for `app` directory

### DIFF
--- a/examples/docs/content/page.mdx
+++ b/examples/docs/content/page.mdx
@@ -1,0 +1,3 @@
+# Just Page
+
+{/* this file should not be treated as App Router page */}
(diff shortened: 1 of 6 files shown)


examples/docs/content/page.mdx
```diff
@@ -0,0 +1,3 @@
+# Just Page
+
+{/* this file should not be treated as App Router page */}
```
dimaMachina, line 3:
```suggestion
{/* this file will be NOT treated as App Router page */}
```

PR #3624, opened 2024-11-01
Autofix: [v4] navbar order is different from _meta

lattaai1:
I investigated the issue with the sidebar order in the Nextra docs theme. The problem appears to be in the `Menu` component within the `sidebar.tsx` file. To fix this, I've modified the `Menu` component to respect the order of items as they are passed in the `directories` prop.

Here's what I changed:

1. Added a `sort` function to sort the directories based on their `sort` property if available.
2. Modified the `Menu` component to use this sorting function before rendering the items.

This change ensures that the sidebar items are displayed in the correct order as specified in the configuration. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission


dimaMachina:
no thank you, bb

PR #3663, opened 2024-11-06
Accept `tabClassName`

hasparus:
I'm adding `tabClassName` prop to make styling HeadlessTab easier (without `> button` selector).
cc @dimaMachina

### DIFF
--- a/packages/nextra/src/client/components/tabs/tabs.tsx
+++ b/packages/nextra/src/client/components/tabs/tabs.tsx
@@ -2,11 +2,11 @@
 
 import {
   Tab as HeadlessTab,
-  TabProps as HeadlessTabProps,
   TabGroup,
   TabList,
   TabPanel,
-  TabPanels
+  TabPanels,
+  type TabProps as HeadlessTabProps
 } from '@headlessui/react'
 import type {
   TabGroupProps,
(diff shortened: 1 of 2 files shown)


packages/nextra/src/client/components/tabs/tabs.tsx
```diff
@@ -32,16 +33,18 @@ export const Tabs: FC<
     items: (TabItem | TabObjectItem)[]
     children: ReactNode
     storageKey?: string
-  } & Pick<TabGroupProps, 'defaultIndex' | 'selectedIndex' | 'onChange'> &
-    Pick<TabListProps, 'className'>
+    className?: TabListProps['className']
+    tabClassName?: HeadlessTabProps['className']
```
dimaMachina, line 37:
```suggestion
    className: TabListProps['className']
    tabClassName: HeadlessTabProps['className']
```
they should already have `undefined` as possible value, also can you add a changeset? `patch`

hasparus: Won't this create issues in projects with `exactOptionalPropertyTypes: true`?

i've just made a CodeSandbox to show this: https://codesandbox.io/p/sandbox/f9y9cl

image

dimaMachina (approved):

Issue #3672, opened 2024-11-07
[v4] Nextra adds all static sites to navigation while using useContentDirPath


ArikSquad:
**Describe the bug**
Nextra adds all static sites to navigation while using useContentDirPath. Also the /docs could be cut from there but it's not the main point of this issue.

**To Reproduce** Steps to reproduce the behavior:

1. Have static sites on your website
2. use v4
3. useContentDirPath

**Expected behavior**
I expected src/app and all the static pages not to show up. _Also I expected the /docs in the top to not show up._

**Screenshots**
![image](https://github.com/user-attachments/assets/98cdc1c5-5034-4ae3-a743-3802f40bf059)

Talking about the src/app showing there

**Desktop:**

- OS: Windows 11
- Browser: Chrome
- Version 130?

**Additional context** 
Using [nextra@4.0.0-app-router.22](https://github.com/shuding/nextra/releases/tag/nextra%404.0.0-app-router.22)

dimaMachina:
Files which you don't want to show you can always hide with `display: hidden`

> I expected src/app and all the static pages not to show up. Also I expected the /docs in the top to not show up.

the whole point of `content` dir is:

- use `content` dir as an alternative to Next.js `pages` dir to avoid `page.{md,mdx}` convention
- **`to include`** in `pageMap` mdx pages from `content` dir and `page.{mdx,jsx,tsx}` from `app` dir 
- remove `layout: raw` and replace it with `page.{jsx,tsx}` files

> Also the /docs could be cut from there but it's not the main point of this issue.

you can use `type: page` or `display: children` in `_meta` file

ArikSquad:
If I'm gonna use the content dir, I expect it not tho use the jsx routes from app. Also I don't know where to put the `type: page` or `display: children`, I tried the children display but I wasn't sure how. Also I can't display hide the src things cuz they aren't in the content dir

dimaMachina:
> Also I don't know where to put the type: page or display: children,

in `_meta` file in `app` dir, look at docs example https://github.com/shuding/nextra/blob/v4-v2/examples/docs/app/_meta.js

> Also I can't display hide the src things cuz they aren't in the content dir

this is strange, can you share your project?

ArikSquad:
When I'm home I'll invite you to my repo

ArikSquad:
I invited you to my repo, it's in the `feature/nextra4` branch

ArikSquad:
[...]

PR #3677, opened 2024-11-09
make Nextra works with `src/app` and `src/content` directories

dimaMachina:
fixes https://github.com/shuding/nextra/issues/3672


packages/nextra/src/server/index.ts
```diff
@@ -31,6 +32,13 @@ const GET_PAGE_MAP_RE = new RegExp(
   GET_PAGE_MAP_PATH.replaceAll('/', SEP).replaceAll('.', '\\.')
 )
 
+export function getContentDirectory() {
+  // Next.js take priority to `app` rather than `src/app`, we do the same for
+  // `content` directory
```
dimaMachina, line 37:
```suggestion
  // Next.js gives priority to `app` over `src/app`, we do the same for `content` directory
```

PR #3678, opened 2024-11-09
[v4] migrate to zustand v5

87xie:
## Why:

Closes: N/A

Since v5 is the current major version of Zustand, I think it would be appropriate to upgrade to v5 in `nextra-theme-docs` as well.

I reviewed the current usage of `useConfig` and found that using a `selector` and `shallow` is unnecessary. I simplified the implementation so that each call to `useConfig` returns the current state directly and ensures reference consistency, so `useStoreWithEqualityFn` and `shallow` are no longer needed. With this change, migration to v5 is successful.

## Check off the following:

      opening the PR).

### DIFF
--- a/.changeset/ninety-birds-teach.md
+++ b/.changeset/ninety-birds-teach.md
@@ -1,0 +1,5 @@
+---
+'nextra-theme-docs': patch
+---
+
+migrate to zustand v5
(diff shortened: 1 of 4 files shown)


dimaMachina (approved):
thank you!

Issue #3675, opened 2024-11-08
Extremely slow compilation and high resource usage, even on v4 with turbopack


oprypkhantc:
**Describe the bug**

Hey. 

On the latest v4 release (version .22), the nextra-docs template takes >60sec to render the "Documentation" page, while using 300-400% CPU and 1.8GB of RAM on a modern MacBook Pro.

In of itself, a cold start taking some time isn't a huge problem; but even next renders (after changing a single page of content) still take >10sec to complete, all while using 2GB of RAM. Turbopack doesn't help at all here; it just uses 8x-10x times RAM, without any visible improvement to render time.

There is an old closed issue, mentioning that NextJS development is slow, but this really isn't the case with an example app from NextJS - where builds take <100ms. The difference is more than 100x in this case between a clean example NextJS app and a Nextra Docs template, which seems suspicious :)

**To Reproduce**

1. Clone the `examples/docs` template from `v4-v2` branch
2. Install deps, run the NextJS server: `yarn run next dev`
3. Navigate to `http://localhost:3000/docs`, wait for it to render
4. Change anything in `content/index.mdx`
5. See the change after 10-20sec

**Expected behavior**

Rerenders should probably be faster.

**Screenshots**

An example of how much resources initial render with Turbopack uses (some of it is shared memory, but still):

<img width="1129" alt="Screenshot 2024-11-08 at 23 59 44" src="https://github.com/user-attachments/assets/2936a0f3-1ec0-416b-83e7-4861577402a4">

dimaMachina:
Changes from `content` folder may be slow compared to `page.{jsx,tsx,mdx}` convention because it uses dynamic import

https://github.com/shuding/nextra/blob/bf029ce8c2208a2c37c55dbbd67f35b56bc3ce98/packages/nextra/src/client/pages.ts#L11-L13

And it seems like Webpack/Turbopack precompiles all files from it.

Did you notice the same problems on migrated nextra.site website ([`docs` folder](https://github.com/shuding/nextra/tree/v4-v2/docs))?

> See the change after 10-20sec

On my mac I don't need to wait this time, everything happens less for a second

P.S. Also, solving [this issue](https://github.com/vercel/next.js/issues/71959) should improve performance for Turbopack projects

oprypkhantc:
> Did you notice the same problems on migrated nextra.site website ([docs folder](https://github.com/shuding/nextra/tree/v4-v2/docs))?

[...]

PR #3697, opened 2024-11-12
[v4] initialize `transformerTwoslash` only 1 time outside of loader

dimaMachina:
fixes https://github.com/shuding/nextra/issues/3675#issuecomment-2470895556


akirchhofer306-ctrl (commented):
So guess a comma and a ! Do make a difference after all

PR #3742, opened 2024-11-26
[v4] compile `nextra/components`, `nextra/hooks` and `nextra-theme-blog` source code with react-compiler

dimaMachina:


packages/nextra/src/client/mdx-components/pre/copy-to-clipboard.tsx
```diff
@@ -19,21 +19,23 @@ export const CopyToClipboard: FC<ComponentProps<'button'>> = props => {
...
+  > = async event => {
+    setCopied(true)
+    if (!navigator.clipboard) {
+      console.error('Access to clipboard rejected!')
+      return
+    }
+    const container = event.currentTarget.parentNode!.parentNode!
+    const content = container.querySelector('pre code')?.textContent || ''
+    try {
+      // container should be not inside to support value blocks (conditional, logical, optional chaining, etc.) within a try/catch statement
```
dimaMachina, line 33:
```suggestion
      // container should be not inside a try/catch statement, otherwise react-compiler give an error
```

PR #3815, opened 2024-12-16
Export Meta type

hasparus:
## Why:

I'm exposing it so the user doesn't have to depend import `zod` to do `z.infer`.


packages/nextra/src/server/schemas.ts
```diff
@@ -165,8 +165,12 @@ const itemSchema = linkItemSchema
...
 
+export type Item = z.infer<typeof itemSchema>
+
 export const metaSchema = z
   .string()
   .or(menuItemSchema)
   .or(separatorItemSchema)
   .or(itemSchema)
+
+export type Meta = z.infer<typeof metaSchema>
```
dimaMachina, line 176: maybe 
```suggestion
export type MetaRecord = Record<string, z.infer<typeof metaSchema>> 
```

PR #3830, opened 2024-12-19
Fix the signature of transformTitle

hasparus:
## Why:

`transformTitle` is changing the `MetaRecord` type to `unknown`. By adding that generic parameter `T`, we allow it to preserve the type.

I took liberty of moving `transformTitle` up, because there's only one union constituent accepting a string. This also makes the `MetaRecord` type more readable in error messages. @dimaMachina Let me know if that makes sense.

## Check off the following:

      opening the PR).

### DIFF
--- a/.changeset/rotten-llamas-drum.md
+++ b/.changeset/rotten-llamas-drum.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
+Fix `MetaRecord` type so it's no longer `unknown`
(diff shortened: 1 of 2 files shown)


dimaMachina (approved):
great fix!

.changeset/rotten-llamas-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
dimaMachina, line 2:
```suggestion
"nextra": patch
```

PR #3852, opened 2024-12-20
[v4] search tweaks

dimaMachina:


jellli:
Is there any necessary to migrate flexsearch to pagefind in V3?

dimaMachina:
> Is there any necessary to migrate flexsearch to pagefind in V3?

nope, this will be a big breaking change

PR #3999, opened 2025-01-23
handle non-"\w" characters

loveloki:
## Why:

Currently, only mdx files named with `\w` can be processed correctly.

image

Because `cleanFilePath` use `.replaceAll(/[\W_]+/g, '_')` handle filename, so other character will be drop. and finnal can't generate/get correctly route.

## What's being changed

handle route use `encodeURI` and `decodeURI`, This will handle more filenames.

1. `fillPageMap` use `encodeURI`, this can handle (allmost) file paths as `\w` supported characters.
2. `cleanFilePath` use `encodeURI`, this can matches the results generated by `fillPageMap`
3. clint use `decodeURI`, this can get correctly file path.

## preview link

[error case](https://stackblitz.com/edit/vitejs-vite-2sjt2wk6)

## Check off the following:

      opening the PR).

### DIFF
--- a/packages/nextra/src/client/pages.ts
+++ b/packages/nextra/src/client/pages.ts
@@ -7,7 +7,10 @@ import { logger } from '../server/utils.js'
 export async function importPage(pathSegments: string[] = [], lang = '') {
   const RouteToFilepath = await getRouteToFilepath(lang)
 
-  const pagePath = RouteToFilepath[pathSegments.join('/')]
+  const path = pathSegments.join('/')
+  const decodePath = decodeURI(path) // handle non-"\w" characters
+
+  const pagePath = RouteToFilepath[decodePath]
   try {
     // eslint-disable-next-line @typescript-eslint/no-require-imports, unicorn/prefer-module -- Require statement enables Fast Refresh
     return require(`private-next-content-dir/${lang && `${lang}/`}${pagePath}`)
--- a/packages/nextra/src/server/page-map/to-page-map.ts
+++ b/packages/nextra/src/server/page-map/to-page-map.ts
@@ -80,7 +80,7 @@ export function convertToPageMap({
       }
       const item = {
         name: key || 'index',
-        route: `/${path}`
+        route: `/${encodeURI(path)}` // handle non-"\w" characters
       }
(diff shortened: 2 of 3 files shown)


packages/nextra/src/server/page-map/to-page-map.ts
```diff
@@ -80,7 +80,7 @@ export function convertToPageMap({
       }
       const item = {
         name: key || 'index',
-        route: `/${path}`
+        route: `/${encodeURI(path)}` // handle non-"\w" characters
```
dimaMachina, line 83: I think this is not needed? without this change routes are still clickable
image

packages/nextra/src/client/pages.ts
```diff
@@ -7,7 +7,10 @@ import { logger } from '../server/utils.js'
 export async function importPage(pathSegments: string[] = [], lang = '') {
   const RouteToFilepath = await getRouteToFilepath(lang)
 
-  const pagePath = RouteToFilepath[pathSegments.join('/')]
+  const path = pathSegments.join('/')
+  const decodePath = decodeURI(path) // handle non-"\w" characters
```
dimaMachina, line 11:
```suggestion
  const pathname = pathSegments.join('/')
  const decodePath = decodeURI(pathname) // handle non-"\w" characters
```
to avoid confusing with Node.js' `path` module

packages/nextra/src/client/pages.ts
```diff
@@ -7,7 +7,10 @@ import { logger } from '../server/utils.js'
...
+
+  const pagePath = RouteToFilepath[decodePath]
```
dimaMachina, line 13:
```suggestion
  const pagePath = RouteToFilepath[decodedPath]
```

packages/nextra/src/server/page-map/to-page-map.ts
dimaMachina, line 83:
```suggestion
        route: `/${path}`
```

loveloki: Yes, you are right. This is not needed. I was tracking here at first, and mistakenly thought that I needed to deal with it here

Issue #2927, opened 2024-06-12
Search display result inside tab but when clicking on the search result, the page doesn't select the correct tab


gh-axel-czarniak:
Hello,

When searching for information using the nextra search bar, if one of the result is inside a tab component, clicking on the result will redirect to the page but it doesn't open the correct tab containing the text that was searched. 

For example:
On the official website, https://nextra.site/docs, search for `yarn` and then click the first link inside the Tab component result. 
```
Example
Yarn is a software ....
```

This will redirect to the tab page but doesn't open the yarn tab. Thus in the first search, we can't find what we were looking for. 

Is it possible to update this behavior ? 

Thanks

PR #4028, opened 2025-01-24
select right tab and scroll into view when html element with `location.hash` id is inside `<Tabs.Tab>`

dimaMachina:
fixes https://github.com/shuding/nextra/issues/2927

### DIFF
--- a/.changeset/cuddly-cups-promise.md
+++ b/.changeset/cuddly-cups-promise.md
@@ -4,4 +4,4 @@
 
 fix console error from `<Search>` results error `TypeError: Cannot destructure property 'results' of '(intermediate value)' as it is null.`
 
-select right tab and scroll into view when search result content is inside `<Tabs.Tab>`
+select right tab and scroll into view when html element with `location.hash` id is inside `<Tabs.Tab>`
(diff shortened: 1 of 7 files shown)


.changeset/cuddly-cups-promise.md
```diff
@@ -0,0 +1,7 @@
+---
+"nextra": patch
+---
+
+fix console error from `<Search>` results error `TypeError: Cannot destructure property 'results' of '(intermediate value)' as it is null.`
+
+select right tab and scroll into view when search result content is inside `<Tabs.Tab>`
```
dimaMachina, line 7:
```suggestion
select right tab and scroll into view when html element with `location.hash` id is inside `<Tabs.Tab>`
```

packages/nextra/src/client/components/search.tsx
```diff
@@ -108,11 +108,14 @@ export const Search: FC<SearchProps> = ({
...
-        )
+      const response = await window.pagefind!.debouncedSearch<PagefindResult>(
+        value,
+        searchOptions
+      )
+      if (!response) return
+
+      const { results } = response
+
       const data = await Promise.all(results.map(o => o.data()))
```
dimaMachina, line 119:
```suggestion
      const data = await Promise.all(response.results.map(o => o.data()))
```

PR #4039, opened 2025-01-25
v3: Add defaultOpen option to sidebar configuration

### LINKED ISSUE #4032: how to collapse the sidebar by default?
I'm trying to set the sidebar to be collapsed by default and only be opened when the user wants it open but i can't find an option to do this

Image

hichemfantar:
Introduce a `defaultOpen` option to the sidebar configuration, allowing users to specify whether the sidebar should be open by default. This change enhances customization of the sidebar's initial state.

closes #4032

## Check off the following:

      opening the PR).

### DIFF
--- a/docs/pages/docs/docs-theme/theme-configuration.mdx
+++ b/docs/pages/docs/docs-theme/theme-configuration.mdx
@@ -386,6 +386,7 @@ Customize the entire navbar component.
 | sidebar.defaultMenuCollapseLevel | `number`  | Specifies the folder level at which the menu on the left is collapsed by default. Defaults to 2 |
 | sidebar.autoCollapse             | `boolean` | If true, automatically collapse inactive folders above `defaultMenuCollapseLevel`               |
 | sidebar.toggleButton             | `boolean` | Hide/show sidebar toggle button. Defaults to `false`                                            |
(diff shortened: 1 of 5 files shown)


hichemfantar:
There's weird behavior where if the default is false, then you open the sidebar, the elements don't take the full width

could be existing animation code race condition

image

dimaMachina:
Thanks for contributing, since Nextra 4 is released, I do longer accept a new features for v3, sorry for this

hichemfantar:
can you reconsider backporting some things? because the app router is a huge change that many people are avoiding because it's can be buggy at times

hichemfantar:
fixed here as well

dimaMachina:
> can you reconsider backporting some things? because the app router is a huge change that many people are avoiding because it can be buggy at times

I am sorry but I don't have enough free time to maintain both Nextra versions, I can accept only for bugs/security fixes for Nextra 3

hichemfantar:
I also rely on v3 for several projects and migrating isn't feasible atm especially considering the app router doesn't have full feature parity and can be unstable at times compared to the pages router.

I'm willing to work with you as i don't want to have to maintain a separate fork. If the changes aren't breaking and tested, would you consider them?

hichemfantar:
wrapping issue fixed here as well

hichemfantar:
please merge, this pr fixes layout bugs and adds a small non breaking feature

PR #4053, opened 2025-01-25
generate unique anchor id for `<summary>` elements based on its content at build time and add anchor link icon for `<summary>`

dimaMachina:
## desktop
image

## mobile
image


packages/nextra-theme-docs/src/components/sidebar.tsx
```diff
@@ -292,10 +292,11 @@ export const MobileNav: FC = () => {
 
   const menu = useMenu()
   const pathname = usePathname()
+  const hash = useHash()
```
dimaMachina, line 295:
```suggestion
  const hash = useHash()
```

PR #4064, opened 2025-01-26
deprecate `Tabs#items` prop, suggest to use `Tabs.Tab#name` and `Tabs.Tab#disabled` props instead

dimaMachina:


packages/nextra/src/client/components/tabs/index.client.tsx
```diff
@@ -114,6 +132,24 @@ export const Tabs: FC<
...
+        isValidElement<TabPanelProps>(child) &&
+        ({
+          label: child.props.name!,
+          disabled: child.props.disabled!
+        } satisfies TabObjectItem)
+    )!
+
+  if (process.env.NODE_ENV !== 'production') {
+    warnOnce(
+      'You are using deprecated `Tabs#items` prop, and which will be removed in Nextra 5. Use `Tabs.Tab#name` and `Tabs.Tab#disabled` props instead.'
```
dimaMachina, line 149:
```suggestion
      'You are using the deprecated `Tabs#items` prop, which will be removed in Nextra 5. Instead, use the `Tabs.Tab#name` and `Tabs.Tab#disabled` props.'
```

Issue #4081, opened 2025-01-27
ability to enable toc and breadcrumb in type="page"


hichemfantar:
**Is your feature request related to a problem? Please describe.** A clear and
concise description of what the problem is. Ex. I'm always frustrated when [...]

I tried enabling toc in `docs/app/_meta.global.tsx` on https://nextra.site/about but it's not doing anything

```
  about: {
    type: 'page',
    theme: {
      toc:true
    }
  },
```

**Describe the solution you'd like** A clear and concise description of what you
want to happen.

setting toc to true should show the toc.
implicit and undocumented behavior like this should be avoided

**Describe alternatives you've considered** A clear and concise description of
any alternative solutions or features you've considered.

**Additional context** Add any other context or screenshots about the feature
request here.

it already shows in the mobile navbar so it makes sense for desktop to also have the same func

![Image](https://github.com/user-attachments/assets/34901f9f-f8f0-4357-a295-e4b008148f5e)

Issue #4030, opened 2025-01-24
How do i add table of contents in a regular page?


hichemfantar:
I tried enabling toc in the meta file on pages like https://nextra.site/about but it's not doing anything

hichemfantar:
got it to work by nesting inside a folder and renaming it to index

hichemfantar:
why is

hichemfantar:
now i have to deal with this problem 

![Image](https://github.com/user-attachments/assets/3f79c32f-52d4-499b-b673-bff16772b140)

hichemfantar:
@dimaMachina any plans to tackle this because it's very annoying

is it atleast possible to render a group directly as a link if there's only one link inside?

PR #4082, opened 2025-01-27
Ability to render TOC in type page

hichemfantar:
## Why:

this pr gives the user more control and adheres to user expectation.
implicit, undocumented, and no workaround behavior like this should be avoided.

Closes: #4081
Closes: https://github.com/shuding/nextra/issues/4030

http://localhost:3000/about

before
image

after
image

docusaurus allows rendering toc on regular pages which is useful and i need this
image

I realize this changes previous behavior but this is needed and it's best to do it now while v4 is new and very few people are using it.

## Check off the following:

      opening the PR).

### DIFF
--- a/.changeset/twelve-hotels-nail.md
+++ b/.changeset/twelve-hotels-nail.md
@@ -1,0 +1,5 @@
+---
+"nextra-theme-docs": patch
+---
+
+allow change `_meta` file `theme.toc` property with `type: 'page'`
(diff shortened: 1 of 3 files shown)


dimaMachina (approved):
I am ok with this change, but I don't see the reason to make configurable breadcrumb, since it's navigation, and for single page it doesn't make sense

PR #4083, opened 2025-01-27
Fix capitalization of alert name in GitHub alert HOC

hichemfantar:
## Why:

css has capitalize support

image

no visual changes

## Check off the following:

      opening the PR).

### DIFF
--- a/.changeset/nice-weeks-dream.md
+++ b/.changeset/nice-weeks-dream.md
@@ -1,0 +1,5 @@
+---
+"nextra": patch
+---
+
+improve github alert syntax name in DOM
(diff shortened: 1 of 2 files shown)


dimaMachina (changes requested):
This will not work in custom themes since users may not import nextra css

hichemfantar:
then should we use inline css `text-transform: capitalize;` ?

dimaMachina:
> then should we use inline css `text-transform: capitalize;` ?

we can, but we can use simpler approach which I just pushed in your PR

hichemfantar:
Actually the css is the simpler approach because we can avoid js and the non-null assertion, also using css, the user can change the rendering behavior if they want by overriding css but using js removes the control from the user.

PR #4090, opened 2025-01-28
Refactor Anchor component to correctly detect all URLs

hichemfantar:
## Why:

correctly handles all types of links, local, anchor, external, http, https, and other kinds

## Check off the following:

      opening the PR).

### DIFF
--- a/packages/nextra/src/client/mdx-components/anchor.tsx
+++ b/packages/nextra/src/client/mdx-components/anchor.tsx
@@ -1,21 +1,43 @@
+'use client'
+
 import cn from 'clsx'
 import Link from 'next/link'
-import type { ComponentPropsWithoutRef, FC } from 'react'
-import { EXTERNAL_URL_RE } from '../../server/constants.js'
+import {
+  useEffect,
+  useState,
+  type ComponentPropsWithoutRef,
+  type FC
+} from 'react'
 import { LinkArrowIcon } from '../icons/index.js'
+
+const isExternalUrl = (href: string) => {
+  try {
+    const url = new URL(href, window.location.origin)
+    return url.hostname !== window.location.hostname
+  } catch {
+    return false
+  }
+}
 
 export const Anchor: FC<ComponentPropsWithoutRef<'a'>> = ({
   href = '',
   ...props
 }) => {
-  props = {
+  const { children } = props
+  const [isClient, setIsClient] = useState(false)
+
+  useEffect(() => {
+    setIsClient(true)
+  }, [])
+
+  const combinedProps = {
     ...props,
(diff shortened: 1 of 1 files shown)


dimaMachina (commented):
I don't think we need to convert this component to client.

What is the difference of the current approach by yours for detecting external URL?

dimaMachina:
> you can make all these checks but there's always something you could miss and all of this is handled by the js API.

we don't want to check everything just `https?` is fine

I don't want to overcomplicate this component, I don't see convincing arguments

hichemfantar:
There are many variations of urls that aren’t being checked.

What if the component resolves incorrectly and the user can’t override the behavior?

the best solution is to get rid of the fragile auto detection and allow a prop “external” that the user can pass. This way the user isn’t stuck when the inference fails

hichemfantar:
Also what if it’s an http url that points to the same website?

hichemfantar:
> > Also what if it’s an http url that points to the same website?
> 
> We can't handle every case, but if you want in your website - you can by passing custom `a` components in `mdx-components` and fix this case

That’s exactly the point of this pr, it allows you to detect every case

you could however keep the existing implementation and add an extra prop “external” to override when needed

hichemfantar:
> > you can make all these checks but there's always something you could miss and all of this is handled by the js API.
> 
> we don't want to check everything just `https?` is fine
> 
> I don't want to overcomplicate this component, I don't see convincing arguments

The component is already complex because it has implicit undocumented behavior

dimaMachina:
it was documented in blogpost https://the-guild.dev/blog/nextra-4#markdown-links-changes

hichemfantar:
blogs aren't documentation however, blogs are generally for old users that are looking to migrate.

this should be documented in the documentation itself under the link section. also migration guides should be included directly in the documentation. navigating to a blog on another website is bad ux. even i didn't know about this

example https://docusaurus.io/docs/migration/v3

dimaMachina:
> blogs aren't documentation however, blogs are generally for old users that are looking to migrate and news. new users won't know about this, especially since it doesn't show up in search

I can't do everything sorry, it's an Open Source project where I am a one maintainer which I maintain based on my free time

hichemfantar:
That's what my PRs are for, I'm willing to work with you but I also need to know my efforts won't be blocked when the solution is proper and doesn't cause any regressions.
If you're willing to be flexible, we can make nextra pretty good and properly compete with other established solutions.

there's nothing more frustrating than finding a limitation or hard coded behavior that can easily be made flexible.
Do reconsider

hichemfantar:
i have a lot of experience working with documentation generators, i'm willing to share my experience in form of PRs and help make nextra be on par with other established solutions

Issue #4116, opened 2025-01-30
v4 remote docs example bug 


hichemfantar:
I went to https://the-guild.dev/blog/nextra-4#remote-docs

I ran the swr site, click on introduction which went to http://localhost:3000/en/remote/graphql-yoga

![Image](https://github.com/user-attachments/assets/fd440335-1a56-4aa6-b041-b8d0f020b72e)

![Image](https://github.com/user-attachments/assets/8dbb93ba-6f20-4693-984a-a0bb97cae447)

the same route works in the v3 example

0xARK:
Hello, can you telle me how did you manage to start the swr-site example please ? I tried but I'm not familiar with pnpm scripts.

I'm trying to setup the swr-site to test the remote docs feature because I encountered the same issue as you. By integrating the example in my app, I figured out that I needed to change 
```js
const route = params.slug.join('/')
```
to
```js
const route = params.slug?.join('/') || ''
```

hichemfantar:
@0xARK use corepack

PR #4117, opened 2025-01-30
fix remote docs example

hichemfantar:
## Why:

Closes: #4116 

Update page routing to accommodate an optional slug parameter, ensuring proper handling when the slug is not provided.

## Check off the following:

      opening the PR).


hichemfantar:
@dimaMachina the fix works and is the same as v3 except it uses empty string instead of 'index'

letting you know so you're aware in case it's not intended

dimaMachina (approved):
Thanks for fixing it, it's strange that the build command doesn't fail

Issue #4192, opened 2025-02-11
Max Search Result Length


pwils33:
My team and I would like to be able to specify a max character limit for search results. We have problems with some of the search results being too long, as shown in the screenshot. 

![Image](https://github.com/user-attachments/assets/15006938-f12c-42db-b070-3ff67bbd3d4f)

These usually result in trying to search for something that's in a table. We'd like the ability to specify a max character limit for search results to address this issue. I have a working version locally that would address this issue, as shown here.

![Image](https://github.com/user-attachments/assets/dd993edc-5ca0-44e3-8cf8-f02ddbb08a5f)

Can I make a pull request to add this ability? It would need to be for Nextra v3 as it's not feasible for us right now to adopt Nextra v4

dimaMachina:
closing due no activity

PR #4197, opened 2025-02-11
Add parameter to docs theme search component to specify a max character limit for search results

pwils33:
## Why:
Some search results can be too long, we've seen this most for results that links to a table, it results in a bad experience for users when one of those search results takes up the whole space for the search results. This adds an optional property to the docs theme search component that will limit the length of the search results to the specified character limit. 

This is achieved by evaluating each search result, and only includes certain characters around matches to the search query. The amount of characters included before and after the matched substring is determined by the number of matches in the full result, and the total length specified

Closes: https://github.com/shuding/nextra/issues/4192

Example of a search result that is too long
Screenshot 2025-02-10 at 4 54 39 PM

Same search result with changes applied, character limit set to 300
Screenshot 2025-02-10 at 4 55 39 PM

## Check off the following:

      opening the PR).

### DIFF
--- a/.changeset/red-deers-occur.md
+++ b/.changeset/red-deers-occur.md
@@ -1,0 +1,6 @@
+---
+'nextra-theme-docs': patch
+---
+
+Add parameter to docs theme search component to specify a max character limit
+for search results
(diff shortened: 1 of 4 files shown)


dimaMachina:
no new features for the legacy Nextras version, sorry. More info about it I wrote here https://github.com/shuding/nextra/pull/4039#issuecomment-2614081167

PR #4254, opened 2025-02-21
Remove livepeer from showcase

87xie:
## Why:

Closes: N/A

It looks like the [Livepeer Docs](https://docs.livepeer.org/developers/introduction) has been migrated to [Mintlify](https://mintlify.com/docs/quickstart). However, I noticed that [Mastra]() is using Nextra. Perhaps we can add it to the showcase?

## Check off the following:

      opening the PR).


dimaMachina (approved):
Thanks, I also noticed Mastra using Nextra :)

PR #4067, opened 2025-01-26
Conditionally render footer based on children presence to align with user expectation and remove empty space

hichemfantar:
## Why:

Render the footer content only if it is defined in the theme configuration, improving the component's flexibility and responsiveness to theme changes.

Closes https://github.com/shuding/nextra/issues/4031

## What's being changed (if available, include any code snippets, screenshots, or gifs):

before
image

after
image

## Check off the following:

- [x] I have reviewed my changes in staging, available via the **View
      deployment** link in this PR's timeline (this link will be available after
      opening the PR).

PR #4366, opened 2025-03-19
Refactor footer prop types to be optional and to use ReactNode and enhance reactNode validation

hichemfantar:
## Why:

this is a continuation of https://github.com/shuding/nextra/pull/4067 that allows to skip setting a footer entirely instead of having to provide `<></>` as a value to stop the footer from rendering and allows you to use react nodes which are more flexible

## Check off the following:

      opening the PR).


dimaMachina:
superseded by https://github.com/shuding/nextra/pull/4372

PR #4372, opened 2025-03-20
fix: use correct `ReactNode` zod validation for `Layout.footer/banner/editLink/feedback.content/toc.backToTop/toc.extraContent/toc.title/search` and `Navbar.children/projectIcon/chatIcon`

dimaMachina:


packages/nextra/src/server/schemas.ts
```diff
@@ -70,16 +70,35 @@ export const nextraConfigSchema = z.strictObject({
...
-  data =>
-    isValidElement(data) ||
-    (Array.isArray(data) && data.every(value => isValidElement(value))),
-  { message: 'Must be React.ReactNode' }
+  function checkReactNode(data): data is ReactNode {
+    if (
+      // Check if it's a valid React element
+      isValidElement(data) ||
+      // Check if it's null or undefined
+      data == null ||
```
hichemfantar, line 84: undefined is a valid react node

dimaMachina: `== null` matches both `null` and `undefined`

packages/nextra-theme-docs/src/stores/config.tsx
```diff
@@ -25,8 +25,8 @@ export function useConfig() {
 export const ConfigProvider: FC<{
   children: ReactNode
   pageMap: PageMapItem[]
-  navbar: ReactElement
-  footer: ReactElement
+  navbar: ReactNode
+  footer: ReactNode
```
hichemfantar, line 29: footer should be optional imo to avoid being forced to pass null or undefined

dimaMachina: `undefined` is one of union types of `ReactNode`

hichemfantar:
take a look at my comments

packages/nextra/src/server/schemas.ts
hichemfantar: we should make it an explicit check with ===, i'm not a fan of == since it's implicit behavior

packages/nextra-theme-docs/src/stores/config.tsx
hichemfantar: i think we should set it optional explicitly because we're indicating that the prop itself is optional, it works in this case because it coincidentally agrees with the react node type. for example, if in the future react stopped accepting undefined as a valid node then this code would break and the footer prop would no longer be optional for people that don't pass the footer prop.
correct me if im wrong

PR #4375, opened 2025-03-21
fix: suppress key warning

hichemfantar:
## Why:

suppress key warning

also Ignored all MDX files because prettier doesn't support mdx v3

had this bug when i ran prettier

image

image

## Check off the following:

      opening the PR).


dimaMachina:
Thanks for the catch. fixed in https://github.com/shuding/nextra/pull/4394

I know about MDX2/3 problems with prettier, but I still want to keep them formatted

PR #4440, opened 2025-04-07
feat(TSDoc): support flattening return object fields

dimaMachina:


.changeset/soft-rocks-dance.md
```diff
@@ -0,0 +1,7 @@
+---
+'nextra': minor
+'nextra-theme-blog': minor
+'nextra-theme-docs': minor
+---
+
+feat(TSDoc): support flattening return fields
```
dimaMachina, line 7:
```suggestion
feat(TSDoc): support flattening return object fields
```

PR #4451, opened 2025-04-08
tooltip improvement proposal

eduardmartos:
## Why:

The **Tooltips** in the **Breadcrumbs** and **Pagination** _components_ of **Nextra Theme Docs** are not working properly when you define **"title"** in **MDX files** as **ReactElement** instead of **string**, for example, if you want to add an icon before the text in that way:

**_meta.js**
```js
  ...
  'about': {
    title: (
      <DynamicIcon icon="CiSquareQuestion">About</DynamicIcon>
    ),
  }
  ...
```
then, when you move the cursor over the element, it shows the incorrect text "[object Object]" as you can see in the next screenshots:

**Expected tooltip with icon inside title**:
image

**Current situation**:
image

Same happens on bottom pagination. Is the same issue, same solution.

Closes:

I have not found any open issues regarding this behavior at this moment.

My proposal to solve this wrong behavior is to add a little helper that searches for the last string child in the title (as ReactElement), and then use this string as title attribute in the final **span** or **a** element, to replace the current "[object Object]", that is happening because title expects a string and is receiving an Object (ReactElement).

Helper it can be **get-last-string-child.ts** on **utils** folder with this code:
```typescript
import React from 'react'

interface MyComponentProps {
  children?: React.ReactNode | string
}

export function getLastStringChild
(
  element: React.ReactNode | string
): string {
  if (typeof element === 'string') {
    return element
  }

  if (React.isValidElement(element)) {
    const props = element.props as MyComponentProps
    const children = React.Children.toArray(props.children)
    for (let i = children.length - 1; i >= 0; i--) {
      const child = children[i]
      const result = getLastStringChild(child)
      if (result) {
        return result
      }
    }
  }
  
  return ''
}
```
- Add to **utils/index.ts**:
```typescript
export { getLastStringChild } from './get-last-string-child'
```

Then we need to adapt the **components\breadcrumb.tsx** file:
```typescript
...
import { getLastStringChild } from '../utils'
...
<ComponentToUse
  ...
  title={getLastStringChild(item.title)}
  ...
  >
...
```

And the **components\pagination.tsx** file:
```typescript
...
import { getLastStringChild } from '../utils'
...
<NextLink
  ...
  title={getLastStringChild(prev.title)}
  ...
  >
...
<NextLink
  ...
  title={getLastStringChild(next.title)}
  ...
  >
...
```

## Check off the following:

      opening the PR).


dimaMachina (changes requested):
your screenshot is too small to understand your issue

packages/nextra-theme-docs/src/components/pagination.tsx
```diff
@@ -36,7 +37,7 @@ export const Pagination: FC = () => {
     >
       {prev && (
         <NextLink
-          href={prev.route}
+          href={getLastStringChild(prev.route)}
```
dimaMachina, line 40: Why change `route`, route is always `string`

packages/nextra-theme-docs/src/components/pagination.tsx
```diff
@@ -49,7 +50,7 @@ export const Pagination: FC = () => {
       )}
       {next && (
         <NextLink
-          href={next.route}
+          href={getLastStringChild(next.route)}
```
dimaMachina, line 53: same here

packages/nextra-theme-docs/src/components/breadcrumb.tsx
```diff
@@ -42,7 +43,7 @@ export const Breadcrumb: FC<{
                 href &&
                   'x:focus-visible:nextra-focus x:ring-inset x:hover:text-gray-900 x:dark:hover:text-gray-100'
               )}
-              title={item.title}
+              title={getLastStringChild(item.title)}
```
dimaMachina, line 46:
```suggestion
              title={typeof item.title === 'string' ? item.title : undefined}
```
what do you think about this approach?

packages/nextra-theme-docs/src/components/pagination.tsx
eduardmartos: yes! sorry, was a mistake I just changed a few seconds ago... let me correct this. Thanks to noticed it

packages/nextra-theme-docs/src/components/breadcrumb.tsx
eduardmartos: yes, this will be another solution, but it will produce different results depending if you are using icons or not... I assumed that undefined makes that tooltip never appears, makes that final a or span has no title attribute, I'm right?

eduardmartos: You prefer this simple solution to less changes, right? In that way we can remove the added recursive function... as you want... I'd like the two solutions if I can solve not seeing "[object Object]" that is very annoying right now :)

dimaMachina: I would prefer a simpler solution for this edge case

eduardmartos: Nice. let me change in that way...

eduardmartos: Works as expected... with icon no title attribute otherwise as current live version. Good night!

packages/nextra-theme-docs/src/components/breadcrumb.tsx
```diff
@@ -42,7 +42,7 @@ export const Breadcrumb: FC<{
                 href &&
                   'x:focus-visible:nextra-focus x:ring-inset x:hover:text-gray-900 x:dark:hover:text-gray-100'
               )}
-              title={item.title}
+              title={typeof item.title === 'string' ? item.title : undefined}
```
dimaMachina, line 45:
```suggestion
              // Fix `[object Object]` tooltip when title is ReactNode
              title={typeof item.title === 'string' ? item.title : undefined}
```

packages/nextra-theme-docs/src/components/pagination.tsx
```diff
@@ -37,7 +37,7 @@ export const Pagination: FC = () => {
       {prev && (
         <NextLink
           href={prev.route}
-          title={prev.title}
+          title={typeof prev.title === 'string' ? prev.title : undefined}
```
dimaMachina, line 40:
```suggestion
          // Fix `[object Object]` tooltip when title is ReactNode
          title={typeof prev.title === 'string' ? prev.title : undefined}
```

dimaMachina (approved):

PR #4458, opened 2025-04-10
split `TSDoc` component logic to `TSDoc` component and `generateDocumentation` function

dimaMachina:


packages/nextra/src/server/tsdoc/tsdoc.tsx
```diff
@@ -42,57 +51,49 @@ async function renderMarkdownDefault(description?: string): Promise<ReactNode> {
   return <MDXRemote compiledSource={rawJs} />
 }
 
+/**
+ * Component which render the props table for a TypeScript type or interface.
```
dimaMachina, line 55:
```suggestion
 * Component which renders the props table for a TypeScript `type`, `interface` or `function`.
```

PR #4461, opened 2025-04-11
fix: remove KitQL from showcase

yeskunall:
## Why:

Their docs are no longer built using Nextra, but [Starlight instead](https://github.com/jycouet/kitql/tree/main/docs).

## Check off the following:

      opening the PR).


dimaMachina (approved):
Thank you!

Issue #3103, opened 2024-08-06
Support a callback function to capture search results.


Mike-Logit:
I can't find a way to add a callback function that gets called with the search results (from flexsearch) so that I can add some analytics/logs etc... about what the user searched for.

I checked the code but the theme config options are very limited, and the search function isn't accessible or open to extension.

I don't think it's supported but it would be great to have if possible. I don't think it would be too difficult to add.

dimaMachina:
Hi, do you still need it? Is there an alternative in other search engine like algolia, or am a with these callbacks functions?

dimaMachina:
closing due no reply

PR #4621, opened 2025-06-18
Add callback in search component

tuliren:
## Why:

The `Search` component currently does not have custom callbacks. In practice, it is often important to collect data about what users are searching for.

Closes: #3103.

This PR adds an `onSearch` callback to the `SearchProps` in `client/components/search.tsx`. This callback, when specified, is called upon every search input change.

## Check off the following:

      opening the PR).


dimaMachina (approved):
Great addition, yep you will need your client search wrapper

tuliren:
Thank you for the review.

I just confirmed that the change works on the example doc.

Issue #4779, opened 2025-09-02
Copy button pushes centered page titles to the left


clicktodev:
https://github.com/user-attachments/assets/022afb0d-6c54-46c0-a62b-f7a73d417c92

<img width="2824" height="1708" alt="Image" src="https://github.com/user-attachments/assets/330386b4-9da3-44fc-abe8-8620de554157" />

PR #4780, opened 2025-09-02
Enhance CopyPage layout and functionality + Copilot link

clicktodev:
## Why:

Closes: #4779

Adjust the layout of the CopyPage component within ClientWrapper for better alignment. 
Add copilot link.
Introduce a new CopyCheckIcon to provide visual feedback when content is copied.
Use link component directly for better a11y.
Add open and close animations
Change copy icon to markdown icon in list item

image

https://github.com/user-attachments/assets/176b14a6-9175-42e4-883f-e6a042ba84ca

## Check off the following:

      opening the PR).


clicktodev:
@dimaMachina

clicktodev:
it's usually best to keep the button label the same and just change the icon to indicate the action state (loading, success).
also changing the button label causes content shift as you can see in the linked issue.

the open ai docs don't change the text as well which helps avoid button resize and potential content shift.
image


same with shadcn docs
image

the openai docs implement the button on the same line as the page title with flex wrap which works pretty nicely.

clicktodev:
I personally don't mind the extra space, we can use it as a command bar for more tools in the future.
I also think it makes sense to separate it since it's not part of the md content.

or we can put the copy button on the right above the toc component when it's rendered.
when the toc component isn't rendered, then we fallback to the current render position on top of the page title.

image

clicktodev:
bump

paleloser:
Hi! We're having issues as well with the placement of the `CopyPage` button. For us, the solution so far has been copying and pasting the component into our source code and apply the desired modifications. Then set the `copyPageButton: false` layout prop  and injecting the component into the pages `Wrapper`.

All of that could've been solved if the component had an ID and we were able to customize it with plain CSS.

In short, can we reuse this MR to give the component an ID? Thanks!

Issue #4825, opened 2025-10-21
Empty lines are invisible when their code block language is unsupported


MrWillCom:
**Describe the bug** Empty lines in code blocks are invisible if syntax highlighting doesn't support them.

**To Reproduce** Steps to reproduce the behavior:

```
\`\`\`js
import { useState } from 'react'

function Counter() {
  const [count, setCount] = useState(0)
  return <button onClick={() => setCount(count + 1)}>{count}</button>
}
\`\`\`

\`\`\`unsupportedLanguage
import { useState } from 'react'

function Counter() {
  const [count, setCount] = useState(0)
  return <button onClick={() => setCount(count + 1)}>{count}</button>
}
\`\`\`
```

(I'm sorry, "\\\`" above is actually "\`")

**Screenshots**:

<img width="1456" height="610" alt="Image" src="https://github.com/user-attachments/assets/636f6320-0d97-404f-b8cc-fa0f8a9176ea" />

**Expected behavior** Empty lines should be preserved.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context** I don't think this problem has anything to do with the platforms. Because I found that those supported languages will have their empty lines being built into a space, which helps preserve the line, but unsupported ones won't.

PR #4826, opened 2025-10-21
Fix missing empty lines when encountering unsupported languages

MrWillCom:
## Why:

Closes: #4825

A space is added to `::after` if the code `<span />` is `:empty`.

## Check off the following:

      opening the PR).


MrWillCom:
Just found that when `showLineNumbers` is applied, this problem no longer occurs.

Issue #4782, opened 2025-09-03
Ability to specify content directory name


jajaperson:
It would be nice to be able to specify an arbitrary path to act as the content directory, e.g. have some `.md` files stored in `/Prose/` and treat this directory the same way `content/` is currently.

dimaMachina:
if you want to submit PR you need to introduce new nextra option e.g. `contentDirName` and replace this line

https://github.com/shuding/nextra/blob/c6912e2d2bbbed0684e7cc14517524112fafbfca/packages/nextra/src/server/index.ts#L40

with

```diff
-function getContentDirectory() {
+function getContentDirectory(contentDirName = 'content') {
  // Next.js gives priority to `app` over `src/app`, we do the same for `content` directory
-  const [contentDir = 'content'] = fg.sync(['{src/,}content'], {
+  const [contentDir = contentDirName] = fg.sync([`{src/,}${contentDirName}`], { 
```

jajaperson:
@dimaMachina happy to work on this, although I can't work out what your using to make types.generated.ts from the zod schemata?

dimaMachina:
You need to run tests and it will be generated https://github.com/shuding/nextra/blob/c6912e2d2bbbed0684e7cc14517524112fafbfca/packages/nextra/package.json#L69

PR #4835, opened 2025-11-08
Add `contentDirName` option

jajaperson:
## Why:

Closes: #4782

This adds a `contentDirName` option to the Nextra config, and the `CONTENT_DIR` is now set via

```ts
function getContentDirectory(dirName: string) {
  // Next.js gives priority to `app` over `src/app`, we do the same for the `content` directory
  const [contentDir = dirName] = fg.sync([`{src/,}${dirName}`], {
    onlyDirectories: true,
  })
  return contentDir
}
```

## Check off the following:

      opening the PR).


jajaperson:
Hang on, the doc's aren't quite right let me fix that.

donovanglover:
Thanks for implementing this feature!

PR #4972, opened 2026-04-03
fix: include lastCommitTime in MDX compiler cache key

Gdhanush-13:
## What
Fixes #4963 - `remarkAssignFrontMatter` captures `lastCommitTime` in its closure at compiler-creation time. Since the compiler is cached per `{format, isPageImport}`, the first file compiled locks its git timestamp into the shared cached compiler, causing all subsequent files to inherit the wrong timestamp.

## Fix
Included `lastCommitTime` in the compiler cache key:

```diff
-cachedCompilerForFormat[`:`]
+cachedCompilerForFormat[`::`]
```

Also relaxed the `Record` type from a narrow template literal to `string` to accommodate the new key format.

Files with the same last-commit date still share a cached compiler, so there is no performance regression for most files.

## Files Changed
- `packages/nextra/src/server/compile.ts`

### DIFF
--- a/packages/nextra-theme-docs/src/layout.tsx
+++ b/packages/nextra-theme-docs/src/layout.tsx
@@ -10,20 +10,13 @@ import type { LayoutProps } from './types.generated'
 
 export type ThemeConfigProps = z.infer<typeof LayoutPropsSchema>
 
-export const Layout: FC<LayoutProps> = (props) => {
+export const Layout: FC<LayoutProps> = props => {
   const { data, error } = LayoutPropsSchema.safeParse(props)
   if (error) {
     throw z.prettifyError(error)
   }
-  const {
-    footer,
-    navbar,
-    pageMap,
-    nextThemes,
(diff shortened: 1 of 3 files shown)


dimaMachina (commented):
> Alternatively, the cached compiler could be made timestamp-aware by passing lastCommitTime through file.data on the vFile (as compileMetadata already does) rather than through the plugin options closure. This would allow compiler caching while still supporting per-file timestamps.

I think this will be a better fix? can you add this change? 🙏

Gdhanush-13:
Done! Updated the implementation as suggested — lastCommitTime is now passed through ile.data on the vFile rather than through the plugin options closure. This means the compiler cache key is now just ${format}: (without lastCommitTime), so the cached compiler can be shared across files with different timestamps. 
emarkAssignFrontMatter reads lastCommitTime directly from ile.data the same way it already reads 
eadingTime and 	itle.

Issue #4989, opened 2026-05-01
Incompatible with Zod v4.4.x


Wesley-Young:
## Description

After Zod is resolved to `4.4.1`, a Nextra docs site that works with Zod `4.3.6` starts failing during both `next build` and `next dev`.

This happens even when the docs app itself does not directly depend on `zod`. In [my workspace](https://github.com/SaltifyDev/milky), `nextra@4.6.1` and `nextra-theme-docs@4.6.1` both depend on `zod` via `^4.1.12`, so pnpm can resolve Nextra's Zod dependency to `4.4.1` when the lockfile is updated.

With Zod `4.3.6`, the docs build succeeds. With Zod `4.4.1`, `next build` fails during prerendering, and `next dev` fails to render the whole site with:

```txt
Invalid input: expected nonoptional, received undefined
```

The browser-side stack in dev is only the React Server Components client stack:

```txt
resolveErrorDev
../node_modules/.pnpm/next@16.2.4_react-dom@19.2.5_react@19.2.5__react@19.2.5/node_modules/next/dist/compiled/react-server-dom-turbopack/cjs/react-server-dom-turbopack-client.browser.development.js (3260:51)

processFullStringRow
../node_modules/.pnpm/next@16.2.4_react-dom@19.2.5_react@19.2.5__react@19.2.5/node_modules/next/dist/compiled/react-server-dom-turbopack/cjs/react-server-dom-turbopack-client.browser.development.js (4427:23)

processFullBinaryRow
../node_modules/.pnpm/next@16.2.4_react-dom@19.2.5_react@19.2.5__react@19.2.5/node_modules/next/dist/compiled/react-server-dom-turbopack/cjs/react-server-dom-turbopack-client.browser.development.js (4370:7)

processBinaryChunk
../node_modules/.pnpm/next@16.2.4_react-dom@19.2.5_react@19.2.5__react@19.2.5/node_modules/next/dist/compiled/react-server-dom-turbopack/cjs/react-server-dom-turbopack-client.browser.development.js (4593:19)

progress
../node_modules/.pnpm/next@16.2.4_react-dom@19.2.5_react@19.2.5__react@19.2.5/node_modules/next/dist/compiled/react-server-dom-turbopack/cjs/react-server-dom-turbopack-client.browser.development.js (4924:9)
```

In production build, Next redacts the server component error:

```txt
Error occurred prerendering page "/guide/communication".
[...]

PR #4990, opened 2026-05-02
fix(nextra-theme-docs): include children in LayoutPropsSchema.safeParse to fix Zod 4.4.x compatibility

tzellman:
## Problem

Upgrading Zod to 4.4.0 causes a prerender failure in `next build` for any page that doesn't go through the nextra MDX wrapper (e.g. plain Next.js Server Component pages colocated in a nextra app). The error is:

```
✖ Invalid input: expected nonoptional, received undefined
  → at children
```

This is caused by a breaking change in Zod 4.4.0's `$ZodObject` property iteration. Previously, when a required key was **absent** from the parsed object, `z.custom(fn)` would be called with `undefined` and validation would pass if `fn(undefined)` returned `true`. As of 4.4.0, Zod emits an `"invalid_type: expected nonoptional"` error for absent required keys _before_ calling the custom validator, regardless of what the validator would return.

The root cause is in `layout.tsx`. `children` is destructured out of props before `safeParse` is called, so it is never a key in the input object:

```tsx
// children is stripped — `'children' in themeConfig` is false
export const Layout: FC<LayoutProps> = ({ children, ...themeConfig }) => {
  const { data, error } = LayoutPropsSchema.safeParse(themeConfig)
```

`LayoutPropsSchema` declares `children: reactNode` as required. With Zod ≤4.3.6 the `reactNode` custom validator accepted `undefined` and the parse succeeded. With Zod 4.4.x the parse fails.

## Fix

Include `children` in the `safeParse` input so the key is present, and explicitly destructure it out of `data` so it does not flow into the `rest` object passed to `ThemeConfigProvider` (which already excludes `children` via `Omit<ThemeConfigProps, ... | 'children'>` in `theme-config.ts`):

```tsx
export const Layout: FC<LayoutProps> = ({ children, ...themeConfig }) => {
  const { data, error } = LayoutPropsSchema.safeParse({ children, ...themeConfig })
  if (error) {
    throw z.prettifyError(error)
  }
  const { footer, navbar, pageMap, nextThemes, banner, children: _, ...rest } = data
```

The schema is unchanged — `children` remains required in `LayoutPropsSchema`, which keeps the existing type test in `layout-props.test-d.ts` passing.

May be related to #4989.

## Test plan

- [ ] `pnpm test` passes in `packages/nextra-theme-docs`
- [ ] Upgrade `zod` to `^4.4.0` in `packages/nextra-theme-docs/package.json` and confirm `next build` succeeds on a nextra app that includes non-MDX pages

### DIFF
--- a/.changeset/fix-zod-44-children-nonoptional.md
+++ b/.changeset/fix-zod-44-children-nonoptional.md
@@ -1,0 +1,5 @@
+---
+"nextra-theme-docs": patch
+---
+
+Fix Zod 4.4.x compatibility: include `children` in `LayoutPropsSchema.safeParse` so it is not treated as non-optional
--- a/packages/nextra-theme-docs/src/__tests__/layout-props.test-d.ts
+++ b/packages/nextra-theme-docs/src/__tests__/layout-props.test-d.ts
@@ -5,14 +5,14 @@ import type { LayoutProps } from '../types.generated'
 
 describe('Assert types', () => {
   test('LayoutProps should be identical', () => {
-    const WithoutLoose = z.strictObject({
+    const _WithoutLoose = z.strictObject({
       ...LayoutPropsSchema.shape,
       nextThemes: z
         .strictObject(LayoutPropsSchema.shape.nextThemes.unwrap().shape)
         .optional()
     })
 
-    type Expected = z.input<typeof WithoutLoose>
+    type Expected = z.input<typeof _WithoutLoose>
     type Actual = LayoutProps
     assertType<IsEqual<Expected, Actual>>(true)
     return expectTypeOf<Actual>().toEqualTypeOf<Expected>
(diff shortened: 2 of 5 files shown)


terrymun:
The maintainer of zod has merged a fix that will restore the behavior seen in 4.3.x: https://github.com/colinhacks/zod/pull/5939

Hopefully with the next zod release this fix will not be necessary.

tzellman:
Unfortunately the zod 4.4.3 fix doesn't help here. That release added optin = "optional" to $ZodCatch, which only covers schemas created with .catch(). Nextra's reactNode is z.custom() ($ZodCustom), which doesn't get that property, so the object parser still throws the "nonoptional" error for the missing children key before the custom validator even runs.

Quick test to confirm it's still broken with 4.4.3:

```
  const reactNode = z.custom((data) => data == null || isValidElement(data));
  const schema = z.strictObject({ children: reactNode });
  const { children, ...rest } = props;
  schema.safeParse(rest); // still → { expected: "nonoptional", path: ["children"] }
  ```

dimaMachina:
@tzellman please add changeset and I will merge your fix

tzellman:
> /changeset

@dimaMachina should be all set now. Let me know if you need anything else! Thanks!

dimaMachina:
@tzellman @terrymun please ping @shuding on X to update the npm token in the Nextra GitHub repo so we can land this fix in the next patch release!

kyu-sketchsoft:
Thank you for fixing this bug. I’ll wait for this PR to be included in the latest release.

PR #4995, opened 2026-05-11
fix(nextra): use repository.workdir() for GIT_ROOT to support git worktrees

### LINKED ISSUE #4996: next build hangs on docs sites built from a git worktree (GIT_ROOT resolves to .git/worktrees/)
**Describe the bug**

`next build` on a Nextra docs site is significantly slower (often timing out) when the build is run from inside a git repository with multiple [git worktrees](https://git-scm.com/docs/git-worktree). The cause is in `packages/nextra/src/server/loader.ts`:

```ts
// repository.path() returns the `/path/to/repo/.git`, we need the parent directory of it
const GIT_ROOT = repository ? path.join(repository.path(), '..') : ''
```

In a normal checkout, `repository.path()` returns `<repo>/.git/`, so `GIT_ROOT` resolves correctly to the repo root. In a git worktree, `repository.path()` returns `<repo>/.git/worktrees/<name>/`, so `GIT_ROOT` resolves to `<repo>/.git/worktrees/` — a directory that contains no source files.

Every per-file `getFileLatestModifiedDateAsync(path.relative(GIT_ROOT, mdxPath))` call then asks libgit2 to walk the full history looking for a path like `../../docs/content/foo.mdx`, which is enormously slow.

**To Reproduce**

1. Clone any Nextra docs site (or use one of the `examples/` apps).
2. Create a worktree: `git worktree add ../site-build`.
3. From the worktree directory, run `next build`.
4. Compare wall-clock time and per-file `getFileLatestModifiedDateAsync` cost to a build from the original checkout.

**Expected behavior**

`GIT_ROOT` should resolve to the actual working directory of the repository regardless of whether the build is run from the main checkout or a worktree, so `getFileLatestModifiedDateAsync` calls finish in the usual ~16 ms each.

**Additional context**

- Nextra version: 4.6.1 (also present on `main` as of today)
- `@napi-rs/simple-git` exposes `Repository.workdir()`, which returns the correct working directory for both regular checkouts and worktrees (and `undefined` for bare repos). Using it as the primary source, with `path.join(repository.path(), '..')` as a fallback for bare repos, fixes the issue without changing behaviour for regular checkouts.
- Fix PR: #4995

Craga89:
## Why: Running `next build` on a Nextra docs site from inside a [git worktree](https://git-scm.com/docs/git-worktree) hangs (or takes ~100× longer than expected) because every MDX page's git-timestamp lookup walks the entire commit history from the wrong root.

Closes: #4996

`loader.ts` calculates `GIT_ROOT` as:

```ts
// repository.path() returns the `/path/to/repo/.git`, we need the parent directory of it
const GIT_ROOT = repository ? path.join(repository.path(), '..') : ''
```

In a normal checkout `repository.path()` returns `<repo>/.git/`, so this works. In a **git worktree** it returns `<repo>/.git/worktrees/<name>/`, so `GIT_ROOT` becomes `<repo>/.git/worktrees/` — a directory that contains no source files. Every subsequent `getFileLatestModifiedDateAsync(path.relative(GIT_ROOT, mdxPath))` call then asks libgit2 to walk the full history looking for a path like `../../docs/content/foo.mdx`, which is enormously slow.

## What's being changed:

Use `repository.workdir()` instead of `path.join(repository.path(), '..')`. `workdir()` is provided by `@napi-rs/simple-git` (already a dependency) and returns the actual working directory for both regular checkouts **and** worktrees. For bare repositories — where `workdir()` returns `undefined` — fall back to the previous behaviour, preserving the existing contract.

The trailing-slash strip keeps the resulting path shape consistent with the previous `path.join(..., '..')` output, which downstream `path.relative(GIT_ROOT, ...)` calls rely on.

Already validated downstream via `pnpm patch` on Nextra 4.6.1 in our monorepo — `next build` from a git worktree went from timing out at 90 s+ to completing in ~25 s, and a regular checkout build is unchanged.

## Check off the following:

- [ ] I have reviewed my changes in staging, available via the **View deployment** link in this PR's timeline (this link will be available after opening the PR).
- [x] Linked an issue (#4996)
- [x] Added a changeset (`patch` for `nextra`)

### DIFF
--- a/.changeset/fix-git-root-worktree.md
+++ b/.changeset/fix-git-root-worktree.md
@@ -1,0 +1,9 @@
+---
+"nextra": patch
+---
+
+Fix `GIT_ROOT` resolution for git worktrees so MDX `getFileLatestModifiedDateAsync` lookups don't hang `next build`.
+
+`repository.path()` returns `<repo>/.git/worktrees/<name>/` inside a worktree, so `path.join(repository.path(), '..')` resolved to `<repo>/.git/worktrees/` — the wrong directory. Every per-file timestamp lookup then walked the full history from a path that didn't exist relative to the calculated root, taking ~1.6 s per file (a 110-file site went from ~1.8 s total to ~176 s).
+
+`repository.workdir()` returns the correct working directory for both regular checkouts and worktrees, with a fallback to the old behaviour for bare repositories where `workdir()` is undefined.
--- a/packages/nextra/src/server/loader.ts
+++ b/packages/nextra/src/server/loader.ts
@@ -45,8 +45,17 @@ const repository = await (async () => {
   }
 })()
 
-// repository.path() returns the `/path/to/repo/.git`, we need the parent directory of it
-const GIT_ROOT = repository ? path.join(repository.path(), '..') : ''
+// `repository.workdir()` returns the working directory for both regular checkouts
+// and git worktrees. Fall back to `path.join(repository.path(), '..')` for bare
+// repositories where `workdir()` is undefined. In a worktree, `repository.path()`
+// points at `<repo>/.git/worktrees/<name>/`, so joining with `..` resolves to the
+// wrong directory and breaks `getFileLatestModifiedDateAsync`.
+const GIT_ROOT = repository
+  ? (repository.workdir() || path.join(repository.path(), '..')).replace(
+      /\/$/,
+      ''
+    )
+  : ''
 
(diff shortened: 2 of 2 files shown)


dimaMachina (approved):

dimaMachina:
https://github.com/shuding/nextra/pull/4990#issuecomment-4400148644

Issue #5010, opened 2026-06-02
Release workflow has been failing for ~6 months — npm publish returns E404 Not Found


amotarao:
The `Release` workflow (`.github/workflows/release.yml`) has been failing on every "Version Packages" merge since **2025-12-31 (#4886)**. The last successful publish run was **2025-12-04 (#4855)** — roughly **6 months ago**.

Failing Version Packages runs:

| Date | PR | Run |
|---|---|---|
| 2025-12-31 | #4886 | failed |
| 2026-04-07 | #4975 | failed |
| 2026-05-08 | #4992 | failed |
| 2026-05-28 | #4997 | failed (latest, run `26580877266`) |

### Error

All failing runs share the same root cause in the `changesets/action` publish step:

```
🦋  error an error occurred while publishing nextra-theme-blog: E404 Not Found - PUT https://registry.npmjs.org/nextra-theme-blog - Not found
🦋  error 'nextra-theme-blog@4.6.5' is not in this registry.
🦋  error an error occurred while publishing nextra-theme-docs: E404 Not Found - PUT https://registry.npmjs.org/nextra-theme-docs - Not found
🦋  error an error occurred while publishing nextra: E404 Not Found - PUT https://registry.npmjs.org/nextra - Not found
🦋  error packages failed to publish:
 ELIFECYCLE  Command failed with exit code 1.
```

An `npm publish` returning `E404 Not Found - PUT` on packages that *do* exist on the registry is almost always an **authentication problem**, not a missing-package problem. The most common causes:

- The `NPM_TOKEN` secret has expired or been revoked.
- The token type is wrong (e.g., replaced with one that doesn't have publish scope).
- The token belongs to an account that no longer has publish permission on these packages.

### Impact

No new versions of `nextra`, `nextra-theme-docs`, or `nextra-theme-blog` have been released to npm since the **4.6.4** line, even though many fixes have been merged to `main` over the past 6 months (Next.js 16 compatibility, security bumps, Pagefind `basePath` fix, `LayoutPropsSchema` fix, git worktree fix for `GIT_ROOT`, etc.). Users cannot consume these fixes without installing from git.

### Suggested fix

Rather than just rotating `NPM_TOKEN` again, I'd recommend migrating this workflow to **npm Trusted Publishing (OIDC)**, which removes the long-lived token entirely and prevents this whole class of "token silently expired" failures from recurring.

Concretely:

[...]

PR #5011, opened 2026-06-02
ci(release): switch npm publish to Trusted Publishing (OIDC)

amotarao:
## Why:

Refs: #5010

The `Release` workflow has been failing on every "Version Packages" merge for ~6 months because `npm publish` returns `E404 Not Found - PUT` — the `NPM_TOKEN` secret has almost certainly expired or lost publish scope (see #5010 for the full log breakdown and history).

Rather than just rotating the token and waiting for it to expire again, this PR migrates the release workflow to **npm Trusted Publishing (OIDC)** so the workflow no longer depends on a long-lived secret at all. The npm CLI mints a short-lived OIDC token from GitHub Actions on each publish, and provenance attestations are attached automatically.

`.github/workflows/release.yml`:

- Grant `id-token: write` (alongside `contents: write` / `pull-requests: write`) on the release job so the workflow can request an OIDC token from GitHub.
- Set `registry-url: https://registry.npmjs.org` on `actions/setup-node` so the npm CLI knows which registry to exchange the OIDC token against.
- Drop `NPM_TOKEN` from the `changesets/action` env. With Trusted Publishing configured on npmjs.com, `npm publish` will authenticate via OIDC instead.

> **Note on the three permissions.** The current workflow has no `permissions:` block and relies on the repository's default workflow token scope. The moment any `permissions:` block is declared it becomes the complete allow-list and all unlisted scopes drop to `read`. So `contents: write` and `pull-requests: write` are listed explicitly alongside `id-token: write` to preserve the existing behavior — `contents` is needed for `changesets/action` to push the Version Packages branch and for `changeset publish` to push git tags, and `pull-requests` is needed to open/update the Version Packages PR.

```diff
     runs-on: ubuntu-latest

+    permissions:
+      contents: write
+      pull-requests: write
+      id-token: write
+
     steps:
       ...
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
           node-version-file: .node-version
           cache: pnpm
+          registry-url: https://registry.npmjs.org
       ...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

### ⚠️ Required configuration on npmjs.com (maintainer action needed)

This PR is **not self-sufficient** — for the workflow to authenticate, a Trusted Publisher needs to be configured on npmjs.com for each published package. @shuding (or another package owner) please configure the following on https://www.npmjs.com/package/<pkg>/access for each of:

- `nextra`
- `nextra-theme-docs`
- `nextra-theme-blog`

With these values:

| Field | Value |
|---|---|
| Publisher | GitHub Actions |
| Organization or user | `shuding` |
| Repository | `nextra` |
| Workflow filename | `release.yml` |
| Environment name | _(leave blank)_ |

Docs: https://docs.npmjs.com/trusted-publishers

Once the Trusted Publisher is configured on all three packages and this PR is merged, the next Version Packages PR merge should publish cleanly without any token. The existing `NPM_TOKEN` secret can be deleted after a successful publish.

If you'd rather defer the OIDC migration and just rotate the token first, that's fine too — feel free to close this PR. I opened it because it removes the recurring-expiry footgun entirely.

## Check off the following:

      opening the PR).


dimaMachina (approved):
Thank you for your contribution, please ping Shu Ding too https://github.com/shuding/nextra/pull/4990#issuecomment-4400148644

